PR #35, opened 2012-02-26
Added support for plugins

jcayzac:
This adds support for plugins using the `[plugin_name:arg]` syntax.

`test/index.js` shows a basic "youtube" plugin.


jcayzac:
Hi there,

There were lots of good comments here, highlighting use cases that weren't taken into account by my code, and suggesting alternative mechanisms for extending Markdown.

I myself believe this might be out of scope for a Markdown processor after all, and more in line with what's expected of a templating engine —I'm now using Pug as my document format, which is extensible and allows inline Markdown too.

I'm closing this pull request, as it's old and has failed to get consensus.

PR #74, opened 2012-09-03
Adding Support for GFM Tables

kitsonk:
Here is a modified pull request that only adds support for tables that are supported (but undocumented) as part of GFM.  It is implemented in the same fashion as other GFM features.  It appears to be fully compatible with the [Sundown](https://github.com/vmg/sundown) library (which is the markdown parser GFM is based on), though it is strictly based on the [PHP Markdown Extra](http://michelf.ca/projects/php-markdown/extra/) who appears to have introduced the "standard".

The following:

```
Header 1 | Header 2
-------- | --------
Cell 1   | Cell 2
Cell 3   | Cell 4
```

Will render as:

| Header 1 | Header 2 |
| --- | --- |
| Cell 1 | Cell 2 |
| Cell 3 | Cell 4 |


brianllamar:
Anyone else having trouble getting tables to render using marked?
```
var singleColTest = "| one ||-----|  | <code>a&#124;a</code> |"

app.get('/', function (req, res) {
    // res.send('Hello World!');
    res.send(marked(singleColTest));
});
```
returns this
```
| one 
|-----|
| <code>a&#124;a</code> |
```

PR #83, opened 2012-10-02
Added check for token.escaped in heading tok()

darcyclarke:
I ran into this problem while trying to add anchor tags to `token.text` for headings. I noticed that any HTML added to `token.text` was being escaped by default. It seems like this could be easily fixed by adding in a check for `token.escaped` before calling `inline.lexer`. 

I'll make a quick pull request for headings, but my guess is that this kind of check should be done anywhere you're using `inline.lexer` as someone may want to add HTML to `token.text` before parsing. 

_Quick side note:_ the `code` `token.type` does seem to do this check, not sure if it was a conscious decision to leave it out everywhere else.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -521,7 +521,7 @@ function tok() {
       return '<h'
         + token.depth
         + '>'
-        + inline.lexer(token.text)
+        + (token.escaped ? token.text : inline.lexer(token.text))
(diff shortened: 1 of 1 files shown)


joshbruce:
Closing due to staleness and merge conflicts. See #956

PR #96, opened 2012-11-20
Added check for token.escaped in paragraph tok()

tlvince:
Following #83, this does the same for paragraphs.


joshbruce:
Closing due to staleness and merge conflicts. See #956

PR #176, opened 2013-06-18
Add info on tests to readme, move benchmarks lower

tasuki:
Adds information on how to run tests and which ones are failing. Please
note we can't use test numbers, as the tests are run asynchronously and
may finish in any order.


joshbruce:
Closing due to merge conflicts, stale, and not current focus.

PR #180, opened 2013-06-22
Support for Inline and Display Math

summivox:
I'm working to support [MathJax](http://www.mathjax.org/).

Output tag format:
http://docs.mathjax.org/en/latest/model.html#how-mathematics-is-stored-in-the-page

The idea for now:
- `$ ... $` => inline math
- `$$ ... $$` => display math

The code I wrote is mostly monkey-patching. Could anyone help get it settled?

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -204,7 +204,6 @@ Lexer.prototype.token = function(src, top, bq) {
 
     // math
     if (cap = this.rules.math.exec(src)) {
-      console.log(cap); //DEBUG
       src = src.substring(cap[0].length);
       this.tokens.push({
         type: 'math',
@@ -715,7 +714,6 @@ InlineLexer.prototype.output = function(src) {
 
     // math
     if (cap = this.rules.math.exec(src)) {
-      console.log(cap); //DEBUG
       src = src.substring(cap[0].length);
       out += '<script type="math/tex">'
         + cap[1] //FIXME: filter <script> & </script>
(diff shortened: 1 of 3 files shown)


rymohr:
Why introduce a new syntax when you can just piggyback on code blocks and use a custom renderer?

    ```latex
    f(x) = \int_{-\infty}^\infty
        \hat f(\xi)\,e^{2 \pi i \xi x}
        \,d\xi
    ```

https://github.com/chjj/marked#overriding-renderer-methods

summivox:
Because inline math.

cben:
Scholdoc found a reasonable way to use backticks for inline math - require
exactly 2 backticks:
http://scholarlymarkdown.com/Scholarly-Markdown-Guide.html#inline-math
It's almost compatible with existing docs (using more than one backtick is
rare), and still allows anything to be escaped in new docs as you can
always use 3 or more backticks instead of 2.
I'm not saying that's a great syntax, just mentioning it.

If you want to compare to syntaxes others implemented, I'm cataloguing them
at https://github.com/cben/mathdown/wiki/math-in-markdown (feel free to
edit, it's a wiki).
The fragmentation is somewhat sad.
My _personal_ opinion is that most scientists writing in markdown will
flock to pandoc so it's best to stick to one of the syntaxes that pandoc
supports.


2015-03-10 0:44 GMT+02:00 summivox <notifications@github.com>:

summivox:
@cben : I was aware of these work. This fragmentation is unfortunate yet inevitable -- considering we actually have CommonMark now.

Very ironic -- this patch was here over 2 years ago.

joshbruce:
Clsoing due to staleness, possibly going beyond the specifications we are focusing on, and merge conflicts. See #956

PR #226, opened 2013-08-10
lex should only remove a tab in line start

amitport:
lexer should only replace tabs to spaces if they are on the beginning of a line, otherwise those tabs may be meaningful (for example they can be part of a pre-formatted code string)


joshbruce:
Closing as stale and missing tests.

PR #238, opened 2013-08-28
add emoji parsing support to gfm markdown

### LINKED ISSUE #233: Support emoji's like in Github Markdown
:octocat: The markdown parser used by Github supports [emoji](http://www.emoji-cheat-sheet.com/)s.

:interrobang: It'd be nice if they would be enabled in `marked` so they show up correctly in dependant projects like npm's readme pages, as they currently display as unparsed codes.

:point_right: https://github.com/isaacs/npm-www/issues/392 :point_left:

The :warning: and :x: are great for emphasis.

Bartvds:
A simple implementation to work with emoji's like proposed in original #233 issue.

Unlike the complex idea mentioned in #233 this is a more elegant solution that only generates the replacement html code (or whatever you override as template). So no complicated data modules or funky stuff. 

Users will need to source and host the images for themselves, but can customise the template function to be as as simple or fancy as they require (examples provided).

Closes #233, work to solve https://github.com/isaacs/npm-www/issues/392

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,34 @@ Default: `false`
 
 Use "smart" typograhic punctuation for things like quotes and dashes.
 
+### emoji
+
+Type: `Boolean` / `String` / `Function`
+Default: `false`
+
+Enable replacement of colon-delimited Emoji emoticon codes like `:warning:` or `:+1:`. This option requires the `gfm` option to be true. :octocat: 
+
+* `String`: parsed as template for the replacement value, accepts the `{emoji}` variable.
+* `true`: use the default &lt;img&gt; with a local URL prefix.
+* `Function`: called with each emoji code, must return a replacement string like an &lt;img&gt; or &lt;span&gt; tag. Don't forget to use `escape(emoji)` or `encodeURIComponent(emoji)` where needed. 
+
+```js
+marked.setOptions({
+  emoji: function (emoji) {
+    return '<span data-emoji="' + emoji + '"></span>';
+  }
+});
+```
+
+The default template:
+
+````
+<img src="/graphics/emojis/warning.png" alt=":warning:" title=":warning:" 
+  class="emoji" align="absmiddle" height="20" width="20">`
+````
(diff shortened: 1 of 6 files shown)


johanbrook:
Any updates on this? I'd love for this to be merged.

wayou:
is this ready to merge now :question:

gtramontina:
… is there anything preventing this merge?

gkoberger:
I've been using it without issue for over a year. :+1:

Snede:
It's 2017 now, can we please have emojis?

joshbruce:
Believe this is resolved by using UTF emojis. Closing.

joshbruce:
@niknokseyer: Don't believe so. For macOS you bring up the special characters palette `control + command + space` - select the emoji 😀- it gets placed where the cursor is in the text file where the Markdown is...behind the scenes, the UTF character encoding should be there and you should be good to go.

Let me know if that's not the case.

(Note: I don't think this counts for things that are placed in attributes within an element.)

christianbundy:
@joshbruce 

Hello!

That's an option, but that's not GFM. Many platforms, from Slack to Signal, use :emoji: because the ergonomics are better for *most* users. Yes, power users in *some* desktop environments can use chords like the one you suggested, but for most people `:+1:` is much more ergonomic.

I hope you'll reconsider. Thanks for your work on this project.

joshbruce:
@christianbundy: Hey! Thanks for the comment. 

One of the other things about GFM is that there are fundamentals differences between the spec and the implementation on GitHub. GitHub does not use the specification only; therefore, some of the things we are able to do with it are not, strictly speaking, GFM. For example, as of today, emojis are not part of the spec itself.

We are trying to get to the point where we can match the spec to almost 100%, which is not the same as matching the capabilities of GitHub, if that makes sense.

https://github.github.com/gfm/

This, of course, is subject to change depending on circumstance but that's kinda where we are.

PR #244, opened 2013-09-06
Return the bullet string

janpaul123:
Since markdown supports different kinds of bullets, such as "-" and "+", it can be useful to see which one was used. We have used this to create a different markdown dialect in which different bullet types have different meanings.


joshbruce:
Closing due to staleness, missing tests, and does not seem to follow the two specifications we are concentrating on. See #956

PR #259, opened 2013-10-09
Create bower.json

KyleAMathews:


OmgImAlexis:
@KyleAMathews this can probably be closed since bower.json was added at the [end of April 2014](https://github.com/chjj/marked/commits/master/bower.json).

PR #268, opened 2013-10-18
added wrap option

mvhenderson:
New option to control if code block is wrapped with pre/code elements. Default is true for backward compat.

Can be turned off to let the highlight function handle wrapping for custom layouts.


joshbruce:

PR #271, opened 2013-10-26
add support for `lookupCodeLink` option.

kriskowal:
This would allow `marked` to create cross references for technical terms enquoted in back ticks if the user provides a lookup function that returns a link. The function may fall through if there is no corresponding link, and may be implemented using any data structure.


joshbruce:
Closing as stale and merge conflicts. Please resubmit, if desirable after reading #956

PR #284, opened 2013-11-12
support tables in lists like github

q3aiml:
Not sure how often anyone tries to do this other than me, and maybe it is a sign I need to simplify my writing, but github supports tables in lists so here is a simple patch plus tests so marked does too.

For example:
- a list
- table in list
  
  | Weee |
  | --- |
  | I'm in a list |

| Awwww |
| --- |
| Not in a list |

Thanks for looking and thanks for marked.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -200,7 +200,7 @@ Lexer.prototype.token = function(src, top) {
     }
 
     // table no leading pipe (gfm)
-    if (top && (cap = this.rules.nptable.exec(src))) {
+    if (cap = this.rules.nptable.exec(src)) {
       src = src.substring(cap[0].length);
 
       item = {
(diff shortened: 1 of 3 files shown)


Sarev0k:
@chjj What would it take to get this merged?  I'd also like to see this fixed in Marked.

michaelmagistro:
+1 - This would be an excellent feature. Any feature that adds functionality to the lists is great. Personally I always find myself using nested lists, so having features like this available within the list itself would be a great help!

Treri:
Merge please!

joshbruce:
Closing due to merge conflicts and unable to verify whether this capability is actually part of the GFM spec - or a side-effect of the implementation.

PR #373, opened 2014-03-24
Call to AMD#define now specifies a module name to fix an issue with some...

mtritschler:
... AMD implementations (i.e. almond.js)

Fixes https://github.com/chjj/marked/issues/372

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1256,7 +1256,7 @@ marked.parse = marked;
 if (typeof exports === 'object') {
   module.exports = marked;
 } else if (typeof define === 'function' && define.amd) {
-  define(function() { return marked; });
+  define('marked', [], function() { return marked; });
 } else {
   this.marked = marked;
 }


joshbruce:
@UziTech - Thoughts?

UziTech:
I've never used AMD but I don't think it will hurt anything to include the optional id and dependency arguments for define()

Feder1co5oave:
I'm not sure about this: http://requirejs.org/docs/whyamd.html#namedmodules

> You should avoid naming modules yourself, and only place one module in a file while developing. However, for tooling and performance, a module solution needs a way to identify modules in built resources.

Idk what that means.

joshbruce:
Not for nothing, and pardon the ignorance, what problem does this solve exactly that isn't already solved by ES5 (or ES6)?? 

Adding this, if I'm understanding any of it, would seem to increase friction for developers understanding how Marked works, yeah? Possibly having to learn something new. I don't see this pattern much in code I play with - but I see the ES5 and ES6 solutions quite a bit.

UziTech:
AMD is [already implemented](https://github.com/markedjs/marked/blob/cb8277b00ada825676a42f6aea5d93c3fdc66884/lib/marked.js#L1374) in marked. this would just give the AMD module a name so it would work with [almond.js](https://github.com/requirejs/almond).

Since [ES6 script modules](https://caniuse.com/#feat=es6-module) are pretty much available in all major browsers I don't think people should be encouraged to use AMD anymore. (I'm not saying that people should move away from AMD if there site has been using it for a while. But since marked doesn't support almond.js now I don't think we should encourage people using marked with almond.js going forward.)

Feder1co5oave:
I say if there is no disadvantage to merging this, and it would actually
make us compatible with almond.js (I don't even know what that is), why not?

UziTech:
Would be one more thing we have to worry about breaking and seams to give us (and our users) no benefit.

Feder1co5oave:
Sounds fair. As I said, I don't even know what almond.js is...

joshbruce:
Given:

1. The seeming lack of desire from others - no real traction on this proposal for support.
2. Seems to commit us to supporting a specific library or methodology for doing something supported by modern browsers.
3. And so on...

Gonna close. As always, can reconsider later but, for now, think maintaining focus and minimizing complexity is a good strategy. Thank you for the submission!

PitPik:
...what's wrong with fixing a wrong implementation?
```define(function() { return marked; });``` is just wrong. That's not how define works.
The correct way is ```define(name, deps, factory)``` so in case of marked it is: ```define('marked', [], function() { return marked; });```

AMD is part of UMD and is used quite a lot, still. Even modern libraries like webpack uses it and most ES7 or ES6 pre-compile even use it. So, writing ```import foo from '../whatever'``` translates to UMD so ES5 can run it.
Google Chrome started supporting ```import``` with M61, so not too long ago, and there is a lot of software out there that has to support older browsers like IE??

BTW: Not supporting UMD is also ok. ;o)

UziTech:
@PitPik according to https://github.com/markedjs/marked/pull/373#issuecomment-368139981 the AMD build tool is supposed to give the module a name not the module itself.

Since marked doesn't have any dependencies `define(function() { return marked; });` is the same as `define([], function() { return marked; });` as far as I know. So we are just following the requirejs best practices.

PR #380, opened 2014-03-31
Handle renderer arguments in Parser constructor

AaronO:
`Parser.parse` was calling `Parser` with `Parser(options, renderer)` but the constructor didn't handle it. So I just tweaked the constructor so that it now accepts the `renderer` argument.


joshbruce:
Closing as stale and unresolved. See #956

PR #392, opened 2014-04-23
use SVG to display npm module's version

Mithgol:
Rationale:
- SVG looks better on mobile devices with high pixel density
- SVG file size is smaller


joshbruce:
Merged elsewhere.

PR #411, opened 2014-05-05
Move CLI usage up on README.

cirosantilli:
- this is one of the first thing new users will want to know
- makes it clear that there is a CLI interface
- CLI section is much shorter than the full function documentation, and gets hidden by it if it comes afterwards


joshbruce:
Closing due to merge conflicts and out of scope for focus - see #956

PR #443, opened 2014-07-03
previously, when a fenced code block was in the second paragraph of a li...

sfrank2147:
previously, when a fenced code block was in the second paragraph of a list, marked would not respect the spacing of the code block in gfm mode.  This fixes that bug.


joshbruce:

Issue #234, opened 2013-08-23
Add an option to produce valid XHTML


lpinca:
As discussed on IRC i would like to have an option to produce XHTML output for tags with empty elements (e.g. `<br/>`).
This is the default behaviour of `Markdown.pl`.
By default marked will render to modern HTML and will use the empty element syntax only when the option is `true` (`xhtml: true`).
If i'm not mistaken there are only 3 tags involved (`<br>`, `<hr>` and `<img>`).
This feature will allow me to add marked's output in a xml document which will be invalid otherwise.
Unfortunately CDATA does not work in my case, but luckily HTML entities are not a problem.

attilavago:
@chjj , unfortunately you're wrong. XHTML is so not dead. The .epub ebook format uses it extensively.

jimmywarting:
I'm digging into the code an see this xhtml support... is it really needed still doe? feels so old

UziTech:
If someone wants to create an extension that renders xhtml I think we could remove the option.

PR #454, opened 2014-07-25
README: Document xhtml option; fix two spelling errors.

DavidAnson:
Wanting XHTML-compliance, I started to add my own "xhtml" option before I realized one was already present. Issue #234 makes it sound like this feature was a little controversial - but the support is in the code, so I thought it would be good to mention it in the README.

While making that change, I noticed two small misspellings and corrected them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ marked.setOptions({
   pedantic: false,
   sanitize: true,
   smartLists: true,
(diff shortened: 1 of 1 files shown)


joshbruce:
Closing as stale, merge conflicts, and replaced by consolidation of HTML5

PR #463, opened 2014-08-01
Add explicit ids for headers

killall5:


KrishnaPG:
Wondering if this PR takes care of this situation https://github.com/chjj/marked/issues/664 or falls prey to the same problem.

KrishnaPG:
+1 for this. Guess looks like it resolves the case-inconsistency problem too.

Feder1co5oave:
Please describe what this PR is for.

lib/marked.js
```diff
@@ -146,6 +146,12 @@ Lexer.prototype.lex = function(src) {
  * Lexing
  */
 
+Lexer.prototype.id_generator = function(title, explicit_id) {
+	var draft = explicit_id ? explicit_id : title;
+	draft = this.options.headerPrefix + draft;
+	return draft.toLowerCase().replace(/[^\w]+/g, '-');
```
KrishnaPG: Explicit ID should be returned as is, without being case-converted or transformed in any way. If author supplied explicit IDs, there must be a good reason why he did so. Hence it should be used as is. No point in transforming it any further.

joshbruce:
Believe this is already taken care of.

PR #468, opened 2014-08-09
Ability to provide a custom sanitize routine

bzwheeler:


chjj:
I like this idea, but I'd rather have a separate option called `sanitizer`.

PR #493, opened 2014-09-25
Support HTML start tags with new-lines in them.

tommie:
HTML5 Start tags can contain "one or more space characters" (8.1.2.1 #3).

A "space character" is space, tab, LF, FF or CR (2.4.1).

This change makes marked support any Unicode white space in tags by restricting
the autolink to not contain any /\s/ character. It is slightly more permissive
than the HTML5 standard, but that is unlikely to be a problem in practice.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -447,7 +447,7 @@ Lexer.prototype.token = function(src, top, bq) {
 
 var inline = {
   escape: /^\\([\\`*{}\[\]()#+\-.!_>])/,
-  autolink: /^<([^ >]+(@|:\/)[^ >]+)>/,
+  autolink: /^<([^\s>]+(@|:\/)[^\s>]+)>/,
   url: noop,
   tag: /^<!--[\s\S]*?-->|^<\/?\w+(?:"[^"]*"|'[^']*'|[^'">])*?>/,
   link: /^!?\[(inside)\]\(href\)/,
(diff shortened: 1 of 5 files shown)


joshbruce:

tommie:
As you wish. Though marked still doesn't support this, so wouldn't it be better if we got it merged?

(There was no point in me keeping the branch mergeable when I don't see a hint that it will be merged before the next conflict pops up. GitHub didn't send me a notification about your label changes in December.)

joshbruce:
Hey @tommie. Yeah. Understand. See #956. I'm also not sure if it has been fixed since reviving the project...I seem to recall a fix for something like this by either @Feder1co5oave or @KostyaTretyak...what version are you using that still doesn't support this, just want to make sure.

Feder1co5oave:
Looks good to me... Autolinks should not contain whitespace: http://spec.commonmark.org/0.28/#absolute-uri

I'd like to see more test cases to show this though

tommie:
#956 is great news!

> what version are you using that still doesn't support this?

Three years later, I'm not really sure. I was using it through Polymer's <marked-element>, and that was probably before Polymer v1. Polymer v2 exists now. I just looked at the code, and it seemed the regexp still wasn't looking at all whitespace characters.

> I'd like to see more test cases to show this though

@Feder1co5oave, I'm showing the new functionality works, and old tests show there is no change in existing behavior. What type of tests would you like?

PR #500, opened 2014-10-15
create tokens for encountered link definitions

grant-g:
The Eclipse project's Orion [1] markdown editor uses marked both to generate HTML content and to identify the regions in the .md source for syntax styling.  Marked's current lack of generated tokens for encountered link definitions makes it impossible for Orion's editor to reliably identify the regions following a link definition.

This pull request creates tokens for encountered link definitions.  These tokens do not affect the generated output in any way, but enable a client of marked's lexer to know the link definition's character range, which is the only missing piece to complete mapping from the .md source to the generated .html.

[1] https://orionhub.org/

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -372,6 +372,11 @@ Lexer.prototype.token = function(src, top, bq) {
         href: cap[2],
         title: cap[3]
       };
+      this.tokens.push({
+        type: 'def',
+        href: cap[2],
+        title: cap[3]
+      });
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


KrishnaPG:
Wondering if this PR resolves this problem: https://github.com/chjj/marked/issues/664

grant-g:
No, #664 is different.

Feder1co5oave:
+1, I like the idea. Although I would have used `link_def` instead of `def` as the type of the items, since it sounds more clear.

joshbruce:
Closing as stale, missing tests (??), and merge conflicts. Created issue and flagged as milestone 0.5.0: #1025

Feder1co5oave:
I could take a look at this, but I'm wondering if it makes sense to have and call a `def` renderer, since link definitions don't produce any output per se, by spec

PR #502, opened 2014-10-16
Un-ignoring `bower.json` so that the `bower.json` file is available when installing from bower.

kaijarayne:
Many people rely on the `bower.json` being included when installing from bower. The `bower.json` holds a lot of information people rely on when using bower (the version number of the repo, and the `main` property in particular). 

Currently, this file is not available when installing from bower. 

Our use case is that we use this [gulp build tool](https://github.com/ck86/main-bower-files) to compile all vendor dependencies.  It's nice because you allow the repo owner to control the repo's point of entry.

It is a small change and would help the bower community if you were to accept this pull request. Thanks!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -14,6 +14,8 @@
   "main": "lib/marked.js",
   "license": "MIT",
   "ignore": [
+    "**/.*",
+    "!bower.json",
     "node_modules",
     "bower_components",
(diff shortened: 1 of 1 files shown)


joshbruce:
Believe bower issues have been addressed in 2017 - closing as stale.

PR #513, opened 2014-11-19
Adding support for registering new block or inline rules

barretts:
I've gone through a process to allow for the addition of new unique rules to **marked**. A new rule can be added easily enough by passing a properly formatted POJO to registerRule. The downside of this addition is a increased parse time, by 32.6%, when running the included bench test. The original project was focused on speed, and I wonder if this is enough deviation to be a full fork and not be pulled in.

---
This change is [Reviewable](https://reviewable.io/reviews/chjj/marked/513)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # marked
 
 > A full-featured markdown parser and compiler, written in JavaScript. Built
-> for speed.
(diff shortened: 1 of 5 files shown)


joshbruce:
See #1026 

Closing due to merge conflicts and stale.

PR #517, opened 2014-12-01
Raw html

greggman:
I don't know if you think is a good change or if I'm just not understanding something but I thought I'd pass it on

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,33 @@ Default: `false`
 Conform to obscure parts of `markdown.pl` as much as possible. Don't fix any of
 the original markdown bugs or poor behavior.
 
+### rawHtml
+
+Type: `boolean`
+Default: `false`
+
+Don't look inside HTML content, just output the raw HTML as is. If you have markup
+with embedded HTML like this
+
+    some markup
+    <div>
+    a = b * c * d;
+    </div>
+
+By default marked will turn that into
+
+    <p>some markup</p>
+    <div>
+    a = b <em> c </em> d;
+    </div>
+
+`rawHtml: true` will change that output to
+
+    <p>some markup</p>
+    <div>
+    a = b * c * d;
+    </div>
+
 ### sanitize
 
 Type: `boolean`
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1055,7 +1055,7 @@ Parser.prototype.tok = function() {
       return this.renderer.listitem(body);
     }
     case 'html': {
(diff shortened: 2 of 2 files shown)


joshbruce:
Not sure I"m following what this PR actually does. Closing to reduce clutter of PRs to focus on fixing issues rather than enhancements at this time.

greggman:
It says what it does right in the PR.  Copied from the PR

Don't look inside HTML content, just output the raw HTML as is. If you have markup with embedded HTML like this

some markup

```
<div>
a = b * c * d;
</div>
```

By default marked will turn that into

```
<p>some markup</p>
<div>
a = b <em> c </em> d;
</div>
```

rawHtml: true will change that output to

```
<p>some markup</p>
<div>
a = b * c * d;
</div>
```

If it's not clear the second snippet has no `<em>` around the `* c *`

joshbruce:
There is not a test, unless I'm missing it to demonstrate how it will work in practice. Also, I'm not fully understanding why you would want to pass raw html into Marked - to receive exactly what you passed in. Marked exists explicitly to convert Markdown into the HTML...not take HTML and leave it alone.

Tests and use cases would be helpful to this end.

greggman:
There's all kinds of reasons you might want to embed a chunk of HTML in your markdown and not have it get munged by the parser.

In any case I don't know if you care but here's a commonmark example

http://spec.commonmark.org/dingus/?text=some%20markup%0A%0A%20%3Cdiv%3E%0A%20%20%20%20a%20%3D%20b%20*%20c%20*%20d%3B%0A%20%3C%2Fdiv%3E

You can see it doesn't insert the `<em>` where as here's the same thing with marked

https://runkit.com/greggman/5a41dc712cef4c0013920b8f

Maybe that's a feature you like. For me I had complex tables/equations I needed in my markdown and marked was not able to handle them, hence this PR.

joshbruce:
@greggman: Thanks for the detail and example. It appears that what commonmark.js is doing is to ignore markdown within an element passed to the parser.

```
some markup
    
a = b *c* d;

a = b * c * d;

 <div>
    a = b * c * d;
 </div>
```

This will italicize the first `c`. It does appear there might be something with the regex for emphasis, as we should only be italicizing those things that are not spaced. In other words `*c*` should be italicized while `* c *` would not be.

I will put this with the mixed content ticket (#985). Again, the biggest issue with the PR is that there is no test demonstrating that the solution works (see #956).

PR #528, opened 2014-12-12
Generate unique ids for headers with duplicate text

patrickarlt:
I am working on a table of contents generator using marked for another project and noticed that marked does not generate unique ids when headers in the same markdown contain the same text.

For example a link to `#examples` would always go to the first header.

```markdown
# Thing 1

## Examples

## Options

# Thing 2

## Examples

## Options
```

This PR changes the default behavior making to that duplicate headers are given incrementing integers so they are unique, `example-1`, `options-1` in my example.

This is probably a breaking change though since there is 1 failed test but I think its worthwhile since it brings the markdown behavior closer to Github which uses the same strategy to keep header ids unique.

#### Benchmarks

this does come at a small performance cost of ~250-300 milliseconds

##### unique-header-ids

```
marked completed in 3270ms.
marked (gfm) completed in 3504ms.
marked (pedantic) completed in 2968ms.
robotskirt completed in 620ms.
showdown (reuse converter) completed in 9196ms.
showdown (new converter) completed in 13454ms.
markdown.js completed in 14361ms.
```

##### master

```
marked completed in 3009ms.
marked (gfm) completed in 3251ms.
marked (pedantic) completed in 2760ms.
robotskirt completed in 592ms.
showdown (reuse converter) completed in 9018ms.
showdown (new converter) completed in 13270ms.
markdown.js completed in 13973ms.
```

##### Tests

I've added a test for this behavior but this change also causes the the top level paragraphs test to fail

https://github.com/chjj/marked/blob/master/test/tests/toplevel_paragraphs.gfm.text

This is because the header "how are you" header is repeated multiple times in the document so the ID no longer matches. If you want to merge this will change that test to pass.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -148,6 +148,7 @@ Lexer.prototype.lex = function(src) {
 
 Lexer.prototype.token = function(src, top, bq) {
   var src = src.replace(/^ +$/gm, '')
+    , links = {}
     , next
     , loose
     , cap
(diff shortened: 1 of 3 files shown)


chjj:
That 200ms is devestating. There may be a better way to optimize this. I think this uniqueness issue was discussed a while back, but we decided to leave things as they are for the time being. Can we work on optimizing this more?

chjj:
Closing for now.

josephernest:
@chjj @patrickarlt Is there some news on this?

I also have sometimes identical `<h2>` names and then the anchor links / id are not unique, that's a problem.
```

# Thing 1

## Examples

## Options

# Thing 2

## Examples
```

PR #535, opened 2015-01-15
Support Markdown's image size specification

tatsy:
I modified the **'Renderer.prototype.image'** function so that it supports Markdown's image size specification.

For example, a Markdown code

```
![](your_image.jpg =100x200)
```

can be interpreted as

``` html
<img src="your_image.jpg" width="100" height="200" />
```

Thanks!

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -881,9 +881,20 @@ Renderer.prototype.link = function(href, title, text) {
 };
 
 Renderer.prototype.image = function(href, title, text) {
+  var imsizeReg = /([\s\S]+) =([0-9]*)x([0-9]*)/;
+  var match = imsizeReg.exec(href);
+  if(match) href = match[1];
   var out = '<img src="' + href + '" alt="' + text + '"';
   if (title) {
     out += ' title="' + title + '"';
+  }
+  if(match) {
+    if(match[2] != '') {
+      out += ' width="' + match[2] + '"';
+    }
+    if(match[3] != '') {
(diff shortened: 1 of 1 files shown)


capaj:
:+1: would be nicer with a test though

doortts:
+1
May I know when this pull-request will be merge?

doortts:
@tatsy by the way, where can I find `Markdown's image size specification`? Does spec exist?

tatsy:
Hello @doortts.

As you know, official Markdown does not support image size specification. So this is only an option.

If you'd like to specify image size in Markdown, I recommend you to use [markdown-it](https://github.com/markdown-it/markdown-it).

I already provides the plugin for this to specify image sizes ([markdown-it-imsize](https://github.com/tatsy/markdown-it-imsize)).

chjj:
Which implementations is this currently a part of?

isiahmeadows:
@chjj 

Apparently [Mou](http://25.io/mou/) does, but most that actually support this consider themselves derivatives AFAICT [1], like Kramdown and MultiMarkdown.

[1] This was only from a simple Google search, BTW. I'm not *that* aware of the likely hundreds of implementations out there.

webdevbrian:
👍 for this support in the future. Was wondering why this wasn't working. Thanks @tatsy, though it would be better with this merged

joshbruce:
Closing due to merge conflicts, missing tests, and no reference to which supported specification we are targeting. See #956 

Sorry @doortts

PR #536, opened 2015-01-21
Tags whitelist support

mloginov:
options: {
    sanitize: true,
    allowedTags: 'iframe|img|video'
}

For https://github.com/chjj/marked/issues/133

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -605,9 +605,14 @@ InlineLexer.prototype.output = function(src) {
         this.inLink = false;
       }
       src = src.substring(cap[0].length);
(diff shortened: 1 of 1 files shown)


chjj:
Please refactor this so a RegExp is not dynamically created each call. A hash table check would be faster than a regex: reduce an array of tags to an object and check the object's property.

paibamboo:
please continue :)

joshbruce:
I know this is old, see #956 

Closing:

1. Never refactored.
2. Conflicts.
3. Missing tests.
4. Stale.

PR #537, opened 2015-01-23
blacklist vbscript: fixes #492

### LINKED ISSUE #492: XSS in IE
Though set `sanitize` option to `true` could filter some xss code, but it still miss a `link` case in IE.

use:

``` [xss link](vbscript:alert(1&#41;)```

will get a link

```<a href="vbscript:alert(1)">xss link</a>```

this script does not work in IE 11 edge mode, but works in IE 10 compatibility view.

evilpacket:


alduethadyn:
Will this fix be merged?

yoshuawuyts:
Do note that this warrants a new release, just merging it in won't be sufficient.

alduethadyn:
true enough - should have been more clear

okuryu:
@chjj Could you publish to npm?

chjj:
Done.

chjj:
I also want to say to everyone, and @naholyr: PRs for the sanitizer are welcome. The sanitizer is something I added on a whim just because I could. I try to make it as secure as possible, but I would never make a guarantee that it is absolutely secure. If you care about security, you should use a more heavy duty html sanitizer. Hopefully one day we can get to that point though.

naholyr:
I'm pretty clear with that, to be fully honest the aim of my comment was more like well, I'm not sur it worth the effort knowing there will always be an unplanned edge case ;)

I'd go for a warning in the README (+ link to http://htmlpurifier.org/ for example?).

Issue #137, opened 2013-03-14
Beauty spots: Dashes and Correct Quotes.


jashkenas:
I've absolutely been loving Marked lately. Its API is a pleasure to work with, and the ability to "lex" Markdown streams is incredibly handy.

But I am missing a thing or two from other Markdown parsers -- It would be great if Marked could be extended to convert `--` into an em-dash, and turn straight quotes surrounding words into correct curly quotes. So that you could write like this:

```
"Bollocks" -- he interrupted,
```

... and have it come out typographically correct.

joshbruce:
Closing as merge was performed.

PR #540, opened 2015-01-28
Smartypants &ndash and &mdash

karland:
Hi,

according to [Gruber's
Smartypants](http://daringfireball.net/projects/smartypants/)
`--` result in an `&ndash;`(\u2013) and `---` in an `&mdash;`(\u2n14).

As far as I can tell, `marked` converts `--` into `&mdash;`

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -711,8 +711,10 @@ InlineLexer.prototype.outputLink = function(cap, link) {
 InlineLexer.prototype.smartypants = function(text) {
   if (!this.options.smartypants) return text;
   return text
-    // em-dashes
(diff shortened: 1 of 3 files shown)


guypursey:
+1. I agree with this. It's something I've been meaning to put in as a pull request myself for a while. It's also mentioned in #137 but the discussion never went any further.

For me having an em dash surrounded by spaces (as Smartypants gives currently if you use a double hyphen) is ugly; I'd only use an em-dash in certain cases or without spacing. It's good to have a choice, as Gruber's spec originally allowed for.

chjj:
Well done. Merging.

PR #542, opened 2015-02-02
Add -mangle/-no-mangle to enable/disable mangling of email addresses.

dchest:
Hello,

I write my email newsletters in Markdown and I don't want email addresses to be mangled (I think spam filters will have trouble with the mangling), so I introduced -mangle/-no-mangle option (-mangle is default).

Would you consider this for inclusion?

(Warning: no tests, could someone give me a hand on writing a test?)


chjj:
Merging for now. Will probably rename the option.

ccorcos:
Whats the point of mangling in the first place?

PR #547, opened 2015-02-20
Add flags to tablerow

sventschui:


chjj:
Why would this affect tablerow?

sventschui:
We're setting a css class in a custom renderer on the tablerow

sventschui:
@chjj any comments on this?

PR #551, opened 2015-02-24
Add alternative text for non-existing image file

dokenzy:
I added some code for non-exsiting image file. In my project, image files are base64-encoded strings. So, all images in markdown make 404 error.
I guess this is also useful for other peoples.


joshbruce:
Closing due to merge conflicts and missing tests.

PR #548, opened 2015-02-22
Added minimized version for browsers

wastevensv:
Ran marked.js through [javascript-minifier.com](http://javascript-minifier.com/) to get marked.min.js a minimized version for the browser.

PR #553, opened 2015-02-26
Add a Gulpfile

minecrafter:
This currently just handles updating marked.min.js (so we don't have to do things like #548), but Gulp allows for more functionality to be added over time.


adam-lynch:
My two cents: use [gulp-load-plugins](https://github.com/jackfranklin/gulp-load-plugins), especially if more functionality will be added over time. It's really simple but I think it's the best plugin out there. It reduces code, adds a little security by making sure any plugin used has to be saved in your `package.json` and therefore checked-in, etc.

PR #557, opened 2015-03-08
Fixes #538; Parse non-strings

### LINKED ISSUE #538: Parsing only integer bug
Running the `marked.js` on client side with a value which is only integer marked throws an error:

```javascript
marked( 123 );
```
bildschirmfoto 2015-01-26 um 10 14 06

rsandor:
Current default behavior expects the `src` parameter of `marked` to be a string. While this is a pretty legitimate expectation, in JS there are many possible, perfectly valid, scenarios where this might not be the case (e.g. numbers).

This addresses the issue by adding a quick check at the top of `marked`. Basically if the `src` parameter is defined, not a string, and does have a `toString` method then we simply convert it to a string before parsing.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1133,6 +1133,10 @@ function merge(obj) {
  */
 
 function marked(src, opt, callback) {
+  if (typeof src != 'string' && typeof src.toString == 'function') {
+    src = src.toString()
+  }
+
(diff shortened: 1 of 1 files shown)


chjj:
I almost think a better solution would be: `src += ''`. We can do that. Otherwise, if you put in bad inputs, expect bad outputs.

PR #562, opened 2015-03-21
Added headings option

adam-lynch:
This allows for headings to be disabled (and therefore treated like paragraphs). I know you can do it without a config option but it's awkward, I just did it. You need to:

- Manually use the lexer and parser.
- Create your own lexer.
- Override the lexer rule with a noop after it's instantiated.
- Make sure you do the above correctly or else any future instantiations will also have this noop instead of the original rule.
- Make sure you pass your options to the lexer and the parser.
- Make sure your options were merged with `marked.defaults` because the lexer and parser don't do it.
- Do this for each custom renderer / use case you have. I have three.

With this change, you can just set `headings: false` and it's done. The option defaults to `true`. I've added a test for ignoring headings, one for testing headings as is (i.e. `headings: true`) and I've updated `marked.min.js`.

Best "head"-related GIF I found:

head


chjj:
There might be a better more dynamic way to do this for any rule: see #579

PR #567, opened 2015-04-05
Correct version in bower.json

aspyatkin:
Hey, while installing the package with Bower I noticed that package version was not updated in Bower file  (according to GitHub releases page and npmjs registry, now it is 0.3.3, while bower addresses 0.3.2). So this PR fixes this.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "marked",
-  "version": "0.3.2",
+  "version": "0.3.3",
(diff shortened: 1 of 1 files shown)


andrewchilds:
Bower uses git tags for versioning and ignores the `version` property in bower.json (so you might as well just remove `version` from bower.json): https://github.com/bower/spec/blob/master/json.md#version

PR #570, opened 2015-04-06
Correct def_blocks test expected outcome and make test pass

TassosD:


TassosD:
Please take a look to make sure this IS the correct expected HTML outcome

chjj:
Those tests are failing for a reason. Those `[x]:` definitions shouldn't end up as definitions. It's just there to document the failing behavior for now.

PR #573, opened 2015-04-09
typographic switch

iamstarkov:
Hey, I'm using your markdown compiler for a while and want to say thank you for it!

Btw, I’m little obsessed with typography and wrote a bunch of small, full tested and safe to use typographic modules. That’s why I’m wondering will you prefer to delegate typography to 3rd party packages and concentrate marked’s development only on parsing and compiling?

if no, it's fine, I understand. If yes, check out this pull-request and feel free to comment and ask me anything. Will look forward for your response.

PS. My origin idea of this pull-request was that compiler should not be responsible for typography, it should do well parsing and compiling and what do you think?

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -710,19 +710,8 @@ InlineLexer.prototype.outputLink = function(cap, link) {
 
 InlineLexer.prototype.smartypants = function(text) {
   if (!this.options.smartypants) return text;
-  return text
-    // em-dashes
-    .replace(/--/g, '\u2014')
-    // opening singles
-    .replace(/(^|[-\u2014/(\[{"\s])'/g, '$1\u2018')
-    // closing singles & apostrophes
-    .replace(/'/g, '\u2019')
-    // opening doubles
-    .replace(/(^|[-\u2014/(\[{\u2018\s])"/g, '$1\u201c')
-    // closing doubles
-    .replace(/"/g, '\u201d')
-    // ellipses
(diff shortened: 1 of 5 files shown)


matmuchrapna:
hey, hey =) is anybody here? :ghost:

matmuchrapna:
polite ping =)

matmuchrapna:
@chjj is project dead and deprecated?

JCMais:
@matmuchrapna Project is probably dead, given the number of open issues and pull requests.

chjj:
If you want to pull those modules' code into marked, I'll reconsider it, but I really don't want marked to have that many dependencies since it should be used easily in the browser as well.

iamstarkov:
All the deps are one-liners, it’s not big deal, it’s even not _small_ deal

iamstarkov:
on the other side, you will have benefit in form of delegating typography to 3rd party plugins

chjj:
@iamstarkov, marked has no dependencies for a reason. It shouldn't require a build for the browser. A lot of people just `<script src="raw.github.com/.../marked.js>` in their html. There shouldn't be any requires. If they're just one-liners that's great. Let's include them inline.

iamstarkov:
didn’t about how consumers are using marked, seems reasonable.

> If they're just one-liners that's great. Let's include them inline.

okay, it will work. how to deal with responsibility delegation here?

PR #579, opened 2015-04-16
Added ignore() method to disable specific markup.

gbradley:
Works by re-assigning noop; gfm must be false.


adam-lynch:
+1. 

What if something is in `block` and in `inline`?

---

@chjj are you planning to merge this then since you closed #562? :smile:

PR #582, opened 2015-04-22
Do not render strong or em when text contains tag

formulahendry:
Marked will render strong or em with a wrong output in below situation. We should not render strong or em when text contains tag.

```
<table>
  <tr>
    <td>Te__st</td>
    <td>Strin__g</td>
  </tr>
  <tr>
    <td>Te _st</td>
    <td>Strin_ g</td>
  </tr>
</table>
```


chjj:
This is trickier than it appears since the original markdown doesn't process any html as markdown. It's another question of whether to remain true to markdown or add more of our own dynamic behavior. The negative lookahead for a tag might work, but I want to keep marked from being an html parser. Open to argument here.

joshbruce:
Closing as stale, missing tests, and merge conflicts.

Issue #588, opened 2015-05-06
Add option to allow json output instead of a string


4ok:
It is necessary to obtain [BEMJSON](https://en.bem.info/technology/bemjson/v2/bemjson/) from markdown markup.

PR #589, opened 2015-05-06
#588 Add option to allow json output instead of a string

4ok:


chjj:
This changes a lot and adds a lot of method calls. Is the overhead worth the benefit?

4ok:
I have no idea how to do differently.

The rendering result was originally conceived in order to return a string. To return result in any formats, were invented such a decision. In addition this solution is extensible.

Moreover, the method call not greatly affect the speed of execution.

4ok:
Look module [markdown-bemjson](https://github.com/bem-incubator/markdown-bemjson), it based on forked marked.

joshbruce:
Closing due to merge conflicts, stale, missing tests, and beyond scope for the focus per #956.

PR #591, opened 2015-05-18
Added a math parsing feature

shovon:
The parser expects a block surrounded in `$$ ... $$` or an inline component surrounded in `$ ... $`. You supply the renderer, via a `math` option, which is an object with a `render` method that returns a string, representing the outputed math.

This is different from #180, in that the renderer is not limited to just MathJax, but KaTeX will work as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ marked.setOptions({
   pedantic: false,
   sanitize: true,
   smartLists: true,
-  smartypants: false
+  smartypants: false,
+  math: null
 });
 
 console.log(marked('I am using __markdown__.'));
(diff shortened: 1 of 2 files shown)


joshbruce:
Closing as missing tests and not current focus per #956

qubyte:
A quick and dirty solution I took to this (which I use for server rendering) is to add a custom highlighter for code blocks tagged as `mathematics`. It's not perfect (results in SVG embedded in pre tags), but it does work for non-inline maths. Note: this requires marked to be called with a callback since mathjax is asynchronous.

```javascript
const mathjax = require('mathjax-node');

marked.setOptions({
  async highlight(code, language, callback) {
    if (language !== 'mathematics') {
      // Highlighting for other languages
    }

    const result = await mathjax.typeset({ math: code, format: 'TeX', svg: true });

    if (result.errors) {
      return callback(result.errors);
    }

    callback(null, result.svg);
  }
});
```

PR #600, opened 2015-05-29
changed regex to allow parenthesis in link names

RemiArnaud:
Before the patch, this:
[blabla](link%28text%29) would generate incorrect href="link(text" 

now it generate href="link(text)"

The fix is to make the regex greedy, so it takes all the characteres in the href until the last ')'

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -464,7 +464,7 @@ var inline = {
 };
 
 inline._inside = /(?:\[[^\]]*\]|[^\[\]]|\](?=[^\[]*\]))*/;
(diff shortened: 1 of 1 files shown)


AlastairTaft:
+1 Just hit this with inline images

adam-lynch:
If I understand correctly and this would fix the fact that `https://en.wikipedia.org/wiki/Football_(disambiguation)` autolinks to `<a href="https://en.wikipedia.org/wiki/Football_(disambiguation">https://en.wikipedia.org/wiki/Football_(disambiguation</a>)`, then a big :+1:

RemiArnaud:
Yes it does

joshbruce:
#984

joshbruce:
Closing due to merge conflicts and missing tests.

Issue #606, opened 2015-06-20
Support RMarkdown


rgbkrk:
RMarkdown ends up putting an extra spin on the code fences:

<pre>
```{r, echo = FALSE}
sliderInput("bins", "Number of bins:", min = 1, max = 50, value = 30)

renderPlot({
  x <- faithful[, 2]  # Old Faithful Geyser data
  bins <- seq(min(x), max(x), length.out = input$bins + 1)

  # draw the histogram with the specified number of bins
  hist(x, breaks = bins, col = 'darkgray', border = 'white')
})
```
</pre>

Within marked, this ends up getting picked up as paragraph sections, where each empty line delineates the next paragraph.

![screenshot 2015-06-20 14 05 12](https://cloud.githubusercontent.com/assets/836375/8268808/1b02d348-1755-11e5-9bb3-78f2cd194f23.png)

Ended up noticing that by overriding paragraph (when I was really trying to build on the `code` renderer):

```
  var _paragraph = renderer.paragraph;

  renderer.paragraph = function(text) {
    console.log(text);
    _paragraph.call(this, text);
  };
```

rgbkrk:
Note: RMarkdown also has YAML front matter which I can already extract out using solutions from #485.

rgbkrk:
Here's the [Markdown Document Format](http://rmarkdown.rstudio.com/markdown_document_format.html)

joshbruce:
See #956 

RMarkdown is not a targeted specification at this time.

PR #608, opened 2015-06-20
More liberal language regex in fences

rgbkrk:
Addresses #606 by allowing for spaces in the "language" field. Users that want to read RMarkdown can follow this up with their own parsing logic.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -75,7 +75,7 @@ block.normal = merge({}, block);
  */
 
 block.gfm = merge({}, block.normal, {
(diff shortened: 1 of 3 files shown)


rgbkrk:
Now with tests!

Honestly, the way I'd use this is in overriding `renderer.code`:

```

renderer.code = function(code, language) {
  // Parse the RMarkdown details out of language
  // These include the actual language, which is typically R
  // As well as a series of comma delimited options
};
```

rgbkrk:
In fact, this is akin to the style of markdown that pandoc will produce and consume as well: http://pandoc.org/demo/example9/pandocs-markdown.html

rgbkrk:
Closing, since I'm using commonmark now.

PR #613, opened 2015-07-01
Allow options JavaScript file to be provided for cli

matthewp:
This adds an `--options` option. The use case is if you want to provide
more advanced options that are possible with the cli options. For
example you might want to provide a syntax highlighting function. You
could define a module like:

``` js
var highlight = require("highlight.js");

module.exports = {
  langPrefix: 'hljs ',
  highlight: function(code){
    return highlight.highlightAuto(code).value;
  }
};
```

and then you can use it from the cli:

``` shell
cat post.md | marked --options marked_options.js
```

### DIFF
--- a/bin/marked
+++ b/bin/marked
@@ -86,6 +86,10 @@ function main(argv, callback) {
       case '--tokens':
         tokens = true;
         break;
+      case '--options':
+        var p = require("path").join(process.cwd(), argv.shift());
+        options = require(p);
+        break;
       case '-h':
       case '--help':
         return help();


matthewp:
@chjj any opinion on this?

UziTech:
This seems very useful, however I think we are going to deprecate most options at some point to try to slim down marked, and I feel like this would be a short lived feature.

Thoughts @joshbruce @styfle?

styfle:
I would hold off on this one too (it seems its already been on hold for 3 years).

Thinking out loud here, we might want to create a separate package `marked-cli` that depends on the `marked` package. That way the users who have no intention of every using the cli don't get anything extra.

UziTech:
I like the `marked-cli` idea 👍

joshbruce:
Been thinking about separating packages a bit as well. Marked (client-side), Marked CLI, then the docs site. But this would be something more in the 1.x or 2.x imho.

matthewp:
This can be closed then if you all are thinking of a different approach.

joshbruce:
@matthewp: Thanks for the okay. We'll probably leave it open for a little bit. We're mainly just looking to triage things a bit better to turn the focus on passing the test cases and examples for the two supported specifications.

PR #616, opened 2015-07-07
Fixed parsing of empty code blocks.

### LINKED ISSUE #615: Weird parsing of empty blocks of code (with triple backticks)
Originally reported in https://github.com/sourcejs/Source/issues/155#issuecomment-117956180:

Marked's parsing of named code blocks is weird.

    > var m = require('marked')
    > m('```js\n```')
    '<p><code>js</code></p>\n'
    > m('```js\n\n```')
    '<pre><code class="lang-js">\n\n</code></pre>\n'

Why isn't the first example recognized as an empty block of `lang-js`?

papandreou:
Fixes #615 

There's another test failing (def_blocks), but it was also failing before this fix.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -75,7 +75,7 @@ block.normal = merge({}, block);
  */
 
 block.gfm = merge({}, block.normal, {
-  fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]*?)\s*\1 *(?:\n+|$)/,
+  fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]*?)\n\1 *(?:\n+|$)/,
   paragraph: /^/,
   heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/
 });
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -75,7 +75,7 @@ block.normal = merge({}, block);
  */
 
 block.gfm = merge({}, block.normal, {
-  fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]+?)\s*\1 *(?:\n+|$)/,
+  fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]*?)\s*\1 *(?:\n+|$)/,
   paragraph: /^/,
   heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/
 });
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -75,7 +75,7 @@ block.normal = merge({}, block);
  */
 
 block.gfm = merge({}, block.normal, {
-  fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]*?)\n\1 *(?:\n+|$)/,
+  fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n(?:([\s\S]*?)\n|)\1 *(?:\n+|$)/,
   paragraph: /^/,
   heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/
 });
@@ -189,7 +189,7 @@ Lexer.prototype.token = function(src, top, bq) {
       this.tokens.push({
         type: 'code',
         lang: cap[2],
-        text: cap[3]
+        text: cap[3] || '' // will come out as undefined when the block is empty
       });
       continue;
     }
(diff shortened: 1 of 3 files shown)


papandreou:
@chjj I don't understand why the test suite [removes all whitespace before comparing the actual and expected output](https://github.com/chjj/marked/blob/master/test/index.js#L115-L116). As you can see from my fumbling around with these highly whitespace-sensitive fixes here, it would have been a great help if the fixtures were matched exactly, and it would also make the error messages from the test suite much more readable.

I'll happily contribute a rewrite of the test suite in that direction if you're interested.

chjj:
Interesting. What's the use-case for empty code blocks? Does GFM allow them? Weird. Anyway, yeah, this looks alright.

As for the test suite, I didn't want to make the html whitespace output 100% accurate to the original markdown. It's a mess, and I like the way marked does it better. Every engine does it a bit differently and a lot of other markdown engines borrowed what I added to the original test suite, so I make the whitespace insignificant: also because it _is_ technically insignificant in the html world.

I agree the test failure output is annoying to read. I wouldn't be opposed to keeping the whitespace insignificant for testing, but giving better error output based on the character index at which it failed. It shouldn't be hard to find the corresponding character index for the whitespace version.

papandreou:
> Interesting. What's the use-case for empty code blocks?

Funny story, actually. We're building an assertion library called [Unexpected](http://unexpected.js.org/), which is very focused on high quality error messages, which we're trying to improve all the time. That makes it a challenge to keep the documentation up-to-date, so we decided to build a test runner of sorts that actually runs the code snippets in our documentation and compares the error message to the explicitly labelled output, for example:

```
    In case of a failing expectation you get the following output:

    ```javascript
    expect({ a: 'f00', b: 'bar' }, 'to have own properties', {
      a: 'foo',
      c: 'baz'
    });
    ```

    ```output
    expected { a: 'f00', b: 'bar' } to have own properties { a: 'foo', c: 'baz' }
    
    {
       a: 'f00', // should equal 'foo'
                // -f00
                // +foo
      b: 'bar',
      c: undefined // should equal 'baz'
    }
    ```
```

This is already pretty good for regression tests. But sometimes when we're implementing a new feature we just want to base the documentation on the actual error message, so we have another script that runs all the code snippets and updates the `output` blocks with the resulting error messages. Those changes can then be comfortably reviewed using `git diff`. In cases where we just want to bootstrap the documentation from the actually output error message, we just use that script, and that's where the empty `output` blocks come in. They serve as a marker for where we want the actual error message added:

```
    In case of a failing expectation you get the following output:

    ```javascript
    expect({ a: 'f00', b: 'bar' }, 'to have own properties', {
      a: 'foo',
      c: 'baz'
    });
    ```

    ```output
    ```
```

This sort of documentation-driven development is working remarkably well for us, so we'll probably try experimenting with it in other projects unrelated to testing soon :)

papandreou:
> As for the test suite, I didn't want to make the html whitespace output 100% accurate to the original
> markdown. It's a mess, and I like the way marked does it better. Every engine does it a bit differently
> and a lot of other markdown engines borrowed what I added to the original test suite, so I make the
> whitespace insignificant: also because it is technically insignificant in the html world.

> I agree the test failure output is annoying to read. I wouldn't be opposed to keeping the whitespace
> insignificant for testing, but giving better error output based on the character index at which it failed.
> It shouldn't be hard to find the corresponding character index for the whitespace version.

Since the test cases with both the input and the expected output are checked into the test folder here, I maintain that it would be better to match them against each other whitespace sensitively, and then just adjust the expected output of the 10-ish cases where they currently don't match up. At least then we'd know exactly what comes out.

Oh, it just occurred to me that you might have inherited the fixtures from some other project that you don't want to be completely whitespace-compatible with? If you're not keeping the fixtures in sync with an upstream project, I still think we should just make the adjustments or come up with a matching strategy that only ignores the kinds of differences that you explictly don't care about.

chjj:
@papandreou, your PR caused 2 tests to fail. I've reverted it to the old rule changing `[\s\S]+' to `[\s\S]*` which passes your added tests.

chjj:
> Since the test cases with both the input and the expected output are checked into the test folder here, I maintain that it would be better to match them against each other whitespace sensitively, and then just adjust the expected output of the 10-ish cases where they currently don't match up. At least then we'd know exactly what comes out.

Like I said, I think it would be better to keep the tests whitespace-insignificant, but make the error output display the whitespace properly. HTML is whitespace insignificant so this presents no problem.

> Oh, it just occurred to me that you might have inherited the fixtures from some other project that you don't want to be completely whitespace-compatible with?

The original tests were taken from the official markdown test suite (released on a mailing list many years back). I don't plan on touching them. In `test/new` I've added my own tests.

papandreou:
Damn, sorry about that. Reverting to the old rule will mean that certain whitespace won't be emitted as part of the output, though, and in my use case all that whitespace is significant.

I guess I'll have to come back with another attempt later.

chjj:
@papandreou, I modified the old rule to allow no text within the code fences, so now the tests you added pass, along with all the others.

PR #624, opened 2015-07-16
Fix the default options example

dtinth:
We have recently been attacked with XSS because we did not set marked's `sanitize` option. We did not set the `sanitize` option because the docs says that it is `true` by default in the example:

> Example setting options with default values:
> 
> ``` js
> var marked = require('marked');
> marked.setOptions({
>   renderer: new marked.Renderer(),
>   gfm: true,
>   tables: true,
>   breaks: false,
>   pedantic: false,
>   sanitize: true,
> ```

It is later stated in the docs that `sanitize` is `false` by default:

> ### sanitize
> 
> Type: `boolean`
> Default: `false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ marked.setOptions({
   tables: true,
   breaks: false,
   pedantic: false,
-  sanitize: true,
-  smartLists: true,
+  sanitize: false,
+  smartLists: false,
   smartypants: false
 });
 


rgbkrk:
:+1:

rgbkrk:
Same with smartLists too I presume?

dtinth:
Yeah, and that’s included in this PR.

dtinth:
This is already fixed, closing.

PR #629, opened 2015-07-23
Fix small typo occur[r]ed

vincentwoo:


joshbruce:
Thanks @vincentwoo!

vincentwoo:
Blast from a 2015 past :) Merry christmas

PR #630, opened 2015-07-23
Refined the readme.

Pomax:
This change makes it explicit that this is a RegEx based parser, with all the limitations that come with that (such as not being able to deal with parentheses in URL for inline links)


Pomax:
this is clearly never going to land.

joshbruce:
@Pomax: It might have. :)

Having said that, I think we're trying to figure out exactly what Marked wants to be in the future.

See #956

PR #634, opened 2015-08-03
Fix sanitize default in README

dbkaplun:
`sanitize: false` by default, according to https://github.com/chjj/marked/blob/18fb6a639a4a77dd59650879bcad10d833c40067/lib/marked.js#L1245


joshbruce:
Merged elsewhere.

Issue #625, opened 2015-07-16
Do not convert > into &gt; in code blocks


mercmobily:
Hi,

At the moment, in a code block, if you have:

```
def foo
  { :foo => "bar" }
end
```

It gets turned into:

```
def foo
  { :foo =&gt; "bar" }
end
```

Regardless of what `sanitize` is.

In https://github.com/chjj/marked/issues/160#issuecomment-18611040 @scottgonzalez suggested:

```
// Let marked do its normal token generation.
tokens = marked.lexer( content );

// Mark all code blocks as already being escaped.
// This prevents the parser from encoding anything inside code blocks
tokens.forEach(function( token ) {
    if ( token.type === "code" ) {
        token.escaped = true;
    }
});

// Let marked do its normal parsing, but without encoding the code blocks
parsed = marked.parser( tokens );
```

Is there a case were `<code>` blocks should actually be escaped like that, really?

Shouldn't this be the default behaviour, without hacking?
Sorry I would have reopened #160  but couldn't.

Merc.

mercmobily:
Please let us know. There is an open PR for this too... #636

joshbruce:
#983

Issue #529, opened 2014-12-17
Option to turn HTML escaping off


adam-lynch:
I'd love this please. 
### My use case

I want to compile markdown for places where HTML isn't supported, just plain strings (like in Mac & Windows 8+ notifications for example). 

I've already created a custom marked renderer which "cleans" markdown. So I try convert it to the nicest possible plaintext representation I can; i.e. it leaves some things pretty much as they are (like `**strong**`), it converts inline codespans to `{code}` (or `{html}` if code is passed), etc.

It works well but I just noticed that single quotes are being converted to `&#39;`.

joshbruce:
#985

Issue #269, opened 2013-10-21
Add an option to not espace quotes.


philipwalton:
I stumbled upon an issue that's giving me problems because marked is escaping double and single quotes in all situations and there's no option to change it.

https://github.com/chjj/marked/blob/master/lib/marked.js#L999-L1000

I'm no markdown expert, but I'm converting a project to node that previously used Ruby's [rdiscount](http://dafoster.net/projects/rdiscount/), which does not escape quotes. I do not see any reason (from an HTML perspective) to force escaping of quotes, so I think it makes sense to add an option to prevent this behavior.

If this is something you'd consider, I'd be happy to submit a pull request. Just let me know.

ethanmick:
Just ran into this. :+1:

adam-lynch:
Related: #529

chris5marsh:
Plus a really silly point: `&apos;` isn't understood by IE&, which unfortunately is still on my list of supported browsers. An option to turn off some or all entity conversions?

cri5ti:
The code in question is here: https://github.com/chjj/marked/blob/master/lib/marked.js#L685

Shouldn't we have something like this?
```js
if (this.options.sanitize)
    out += this.renderer.text(escape(this.smartypants(cap[0])));
else
    out += this.renderer.text(this.smartypants(cap[0]));        
```

rosshinkley:
I know I'm late to the party, but I just hit this as well doing something similar with Handlebars + marked and thought I'd throw in my +1.

r0hitsharma:
+1 Faced the same issue (Handlebars + marked)

@cri5ti, your suggestion fixes my issue. Thanks!

snollygolly:
+1 Same issue for me as well.

patrickheeney:
+1 Also ran into this. How is everyone working around this issue? Just applying @cri5ti code change after npm install?

caizone:
+1 it's works for me.

matt-:
If someone would like to submit a PR with passing tests that will help move the issue along.

KostyaTretyak:
[...]

PR #636, opened 2015-08-03
Renderer HTML escaping

FredrikAppelros:
Move escaping of HTML from `InlineLexer` to `Renderer`. This will
modularize marked further and allow using it with custom renderers that
do not intend to output HTML.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -583,8 +583,7 @@ InlineLexer.prototype.output = function(src) {
           : this.mangle(cap[1]);
         href = this.mangle('mailto:') + text;
       } else {
-        text = escape(cap[1]);
-        href = text;
+        href = text = cap[1];
       }
       out += this.renderer.link(href, null, text);
       continue;
@@ -593,8 +592,7 @@ InlineLexer.prototype.output = function(src) {
     // url (gfm)
     if (!this.inLink && (cap = this.rules.url.exec(src))) {
       src = src.substring(cap[0].length);
-      text = escape(cap[1]);
-      href = text;
+      href = text = cap[1];
       out += this.renderer.link(href, null, text);
       continue;
     }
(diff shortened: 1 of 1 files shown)


FredrikAppelros:
This indirectly addresses the issues #269, #529 and #625. More importantly, it makes marked a little bit more modular. You can now use the parser and lexer components with a different renderer that doesn't output HTML.

For the fun of it I also made a small branch which directly fixes the issues mentioned above [here](https://github.com/FredrikAppelros/marked/commit/f85dd76c86fa66e0e85bf3b3f5103ccb677154f4), but I personally don't understand why you would want to disable the escaping if it is to be rendered in a browser anyway. :confused:

adam-lynch:
:+1: 

This should be probably also contain changes to the readme, right?

A reason why someone mightn't like it: users overriding renderer methods will inadvertently lose escaping. Why this doesn't matter: this is already the case when overriding link rendering; https://github.com/chjj/marked/blob/18fb6a639a4a77dd59650879bcad10d833c40067/lib/marked.js#L872.

>I personally don't understand why you would want to disable the escaping if it is to be rendered in a browser anyway.

It mightn't be rendered in a browser. We need it for desktop notifications, etc. as I said in #529.

FredrikAppelros:
Yes, this change will affect existing custom renderers that relies on the escaping being done within the lexer. Not sure if it really requires any changes to the README as there was no mention of it being escaped in the first place.

diega:
Hi!
Is there any chance for this to be reviewed and/or merged?

Regards

mercmobily:
+1 Please. It has been more than a year...

joshbruce:
See #746 

We are planning some changes eventually to the Marked library architecture. Adding reference to that issue. Closing due to merge conflicts, stale, and missing tests.

PR #638, opened 2015-08-07
Do not autolink a trailing "!" in a URL

bantic:
I noticed in another project that uses marked that the text "http://localtunnel.me!" was being turned into a link that incorrectly included the trailing "!".

This PR changes the gfm url regex to explicitly not include a trailing "!".

It also expands the tests for gfm links to test for all the possible ending punctuation that the regex doesn't capture as part of the link as well as a test for no trailing punctuation.

I wasn't sure what the correct behavior is for determining whether a character is part of a link or punctuation. The exclamation point is technically a valid character to use in a url, but so are the closing parens (`")"`) and single and double quote marks, which are also part of the [existing url regex](https://github.com/chjj/marked/blob/master/lib/marked.js#L499). It seems far more likely that a sentence written in markdown would end with a url followed by a "!" than that a url would end with a "!", though.

For reference, Rinku, the link-detecting library used by github's markdown gem, [allows "?", "!", ".", "," and ":" as trailing punctuation](https://github.com/vmg/rinku/blob/master/test/autolink_test.rb#L44-L46).

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -496,7 +496,7 @@ inline.pedantic = merge({}, inline.normal, {
 
 inline.gfm = merge({}, inline.normal, {
   escape: replace(inline.escape)('])', '~|])')(),
-  url: /^(https?:\/\/[^\s<]+[^<.,:;"')\]\s])/,
+  url: /^(https?:\/\/[^\s<]+[^<.,:;"'!)\]\s])/,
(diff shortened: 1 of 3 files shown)


joshbruce:
Closing as stale and merge conflicts. Created issue to keep it on our radar. Please resubmit based on latest, if still a problem and defined in one of the specifications being targeted - see #956

PR #648, opened 2015-08-21
Drop unused renderer argument.

mina86:
Neither Parser nor InlineLexer accept renderer aragument so don’t
pass it to them.  Caller can set options.renderer property instead.


joshbruce:
Closing as stale, most likely fixed elsewhere, and merge conflicts.

PR #649, opened 2015-08-21
Avoid declaring the same variable twice.

mina86:
src is an argument to Lexer.prototype.token so there is no need to
declare it as a variable.  It doesn’t seem to do any harm, but some
linters/compilers (read: Closure Compiler) complain.  Since there’s
almost zero cost in getting rid of the duplicate declaration, do so.


joshbruce:
Closing as stale, merge conflicts, and most likely fixed elsewhere.

PR #651, opened 2015-08-24
Update default value of sanitize

thejameskyle:
See:
- https://github.com/chjj/marked#sanitize
- https://github.com/chjj/marked/blob/master/lib/marked.js#L1245


thejameskyle:
Clearing out old issues, closing for inactivity

PR #660, opened 2015-09-09
Add an option to disable markdown elements

Guichaguri:
A simple change that let you disable any markdown element.
It's better creating an option than using [unsafe tricks and hacks](https://github.com/chjj/marked/issues/420#issuecomment-44746538)

## Examples

```javascript
marked.setOptions({
    disabledElements: ['hr', 'heading', 'list']
});

marked('---'); // Returns "---"
marked('# I am not a heading'); // Returns "# I am not a heading"
marked('* Item'); // Returns "* Item"
```

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -114,7 +114,7 @@ function Lexer(options) {
   }
   
   for(var i = 0; i < this.options.disabledElements.length; i++) {
-    this.rules[this.options.disabledElements[i]].exec = function() {};
+    this.rules[this.options.disabledElements[i]] = noop;
   }
 }
 
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -112,6 +112,10 @@ function Lexer(options) {
       this.rules = block.gfm;
     }
   }
+  
+  for(var i = 0; i < this.options.disabledElements.length; i++) {
+    this.rules[this.options.disabledElements[i]].exec = function() {};
+  }
 }
 
 /**
@@ -1252,7 +1256,8 @@ marked.defaults = {
   smartypants: false,
   headerPrefix: '',
   renderer: new Renderer,
-  xhtml: false
+  xhtml: false,
+  disabledElements: []
 };
 
 /**


Guichaguri:
I just found #579. I don't know which one you want to merge, both of them have pros and cons.

Guichaguri:
Fixed merge conflicts

joshbruce:
Thanks for the PR @Guichaguri! Much appreciated. Can you write the tests demonstrating the examples? (Note: I am wondering about the flexibility of our testing harness to actually test all the things we need it to.)

Guichaguri:
How should I write it? After a quick look through the test code, it seems like it can only test with predefined options.

joshbruce:
@Guichaguri: Let me tag in @UziTech, @matt-, and @Feder1co5oave to help on possibly getting a test written on this.

UziTech:
Ya, the tests aren't very option friendly at the moment. I am working on redoing the tests to allow options to be set with [`front-matter`](https://www.npmjs.com/package/front-matter)

stay tuned...

Feder1co5oave:
I think this feature is destructive: once you've disabled an element, you
cannot re-enable it. Correct me if I'm wrong

UziTech:
I submitted pr #1002 to fix the tests and allow options to be specified with [`front-matter`](https://www.npmjs.com/package/front-matter)

@Feder1co5oave it looks like you can re-enable all elements by setting the `disabledElements` option to an empty array (or removing the elements that you want to re-enable)

Guichaguri:
Once the test PR gets merged, I'll write tests for all cases, including re-enabling disabled elements.

I should also write documentation for it.

joshbruce:
@Guichaguri: Can you also handle @Feder1co5oave's question. I'm not sure I would choose the word "destructive" just because of connotation...maybe "global" - the entire markdown string being parsed??

Which I think is true, yeah?

It would be interesting if it could be open to extension a little bit...like "disable the frist 5 instances of H1" or "disable the first, second, and fifth H1". Having said that, given the developer requests to date, being able to stop Marked from rendering an element in a markdown is the request...not granular control.

Am I missing something in @Feder1co5oave's question.

Feder1co5oave:
That's what I'm talking about:

```node
> marked('# heading?')
'<h1 id="heading-">heading?</h1>\n'
> marked('# heading?', {disabledElements: ['heading']})
'<p># heading?</p>\n'
> marked('# heading?')
'<p># heading?</p>\n'
> marked('# heading?', {disabledElements: []})
'<p># heading?</p>\n'
```

UziTech:
I was wrong, I didn't realize the rules were overwritten in the lexer. @Guichaguri maybe a better approach would be to copy the rules to `this.rules` only if they are not in `this.options.disabledElements` otherwise set the rule as `noop`. That way the static rule sets are left alone.

Also should it be called `disabledRules` instead of `disabledElements`. I feel like elements refer to tags like `h1`, `h2`, etc.

joshbruce:
My two cents would be that, yes, "elements" are individual elements (skip h1) while "rules" are all the *types* (skip headings). Can we have the ability to specify both?

sungwoncho:
What can do to help this feature get merged to master?

joshbruce:
@sungwoncho: Hello! Right now our focus is on complying with our chosen specifications and that's where we need the most help. This feature, if approved, would not be put in until after 1.0 or later. Therefore, from an over-arching project perspective, if you are still interested in moving toward at least the possibility of having this feature include sooner rather than later, then I recommend helping us with the specification compliance mission of 0.x: https://github.com/markedjs/marked/issues/1216

Up to you! Thanks again.

PR #666, opened 2015-09-30
Fix default value for "sanitize"

bobby-brennan:


joshbruce:
Should have been corrected elsewhere.

Issue #201, opened 2013-07-19
Hashtags in the beginning are rendered as headings


jorilallo:
I ran into a problem while testing hashtags:

    marked("#hashtag test\n test")

renders as

```
<h1>hashtag test</h1>
<p> test</p>
```

while the expected output is

```
<p>#hashtag test</p>
<p> test</p>
```

silkimen:
+1 GFM

chjj:
Fixed.

xaka:
I know it's been a while, but for the future reference I want to point out that this change should have never landed as a patch (0.3.2 -> 0.3.3), but at least as a minor (0.3.2 -> 0.4.0) because it's backward incompatible from customer's perspective, who don't care about whitespace because it's been allowed for a long time and now it breaks existing badly formatted markdowns. No complains whatsoever. Just suggestions :wink:

Stephen-Mc:
Very unhappy at this change, it broke all our documentation by no longer rendering the headings correctly. Would love to see it reverted, as breaking existing docs just to be able to render hashtags without an escape character does not strike me as the right set of priorities.

Thankfully most of our docs are parsed for conversion to HTML, and are still rendered correctly on our websites, but I have 100 or so pages that are not parsed this way and now need to be updated with the new heading style.

zcyzcy88:
How about `<h1>` require space, and `<h2>`~`<h6>` don't require space?
Because people are used in Twitter style text, when they type `#`, it means hashtag, not heading.

---

**Example:**
```md
Text Text Text Text Text Text Text Text.
#tag1 #tag2
```

---

**Now render as:**

Text Text Text Text Text Text Text Text.
# tag1 #tag2

---

**Expect:**

Text Text Text Text Text Text Text Text.
#tag1 #tag2

Stephen-Mc:
That would be even more confusing than the current state of affairs.

Grey2k:
#hashtag# test

PR #671, opened 2015-11-02
Fixed header formatting

onlineth:
Fixed header formatting for both "#[text]" and "# [text]"


Feder1co5oave:
You should provide some simple examples to show this behaviour.

Anyway, this problem is only relevant for GFM, since if you disable it you can omit the space between the hash and the header's text. In fact, Github uses hashes to link issues, that's the reason to require a space in between.
~~~markdown
###Header

### Header
~~~

###Header

### Header

Feder1co5oave:
Read #201

onlineth:
Thank you @Feder1co5oave for pointing that out

PR #683, opened 2015-11-16
Fix smartypants test inputs.

kohler:
The `text.smartypants` tests differ between the `test/tests` and `test/new` directories (so after `node test --fix`, `node test` fails the smartypants tests). The `test/tests` versions appear to be the right ones.


joshbruce:

PR #702, opened 2015-12-29
add chinese character range to heading escape regexp

gengen1988:
It allow marked render Chinese by default like that:

``` markdown
## 灵活性
```

to this:

``` html
<h2 id="灵活性">灵活性</h2>
```

instead of:

``` html
<h2 id="-">灵活性</h2>
```

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -797,7 +797,7 @@ Renderer.prototype.heading = function(text, level, raw) {
     + level
     + ' id="'
     + this.options.headerPrefix
-    + raw.toLowerCase().replace(/[^\w]+/g, '-')
+    + raw.toLowerCase().replace(/[^\w\u4e00-\u9eff]+/g, '-')  // \u4e00-\u9eff is common chinese character range
     + '">'
     + text
     + '</h'


Feder1co5oave:
The default heading rendered generates IDs just for convenience (many complained about this, too) and it must not be relied upon. Write a custom rendered with a generating function that suits your needs. I might suggest using something like https://github.com/jeremys/uslug

gengen1988:
Thanks for the suggestion. I can use uslug in my own project.

However some basic tool like "Atom Editor" use marked to preview markdown, also some tools such as "markdown to pdf" or "markdown toc" depend on it. I cannot force them all use uslug to render a usable heading link in Chinese or Russian.

Some python implements will render heading link to a random ID. Maybe we can use that as a default implements instead?

zetsin:
```
const renderer = new Marked.Renderer()
renderer.heading = function (text, level, raw) {
  return '<h'
    + level
    + ' id="'
    + this.options.headerPrefix
    + raw.toLowerCase().replace(/[\s]+/g, '-')
    + '">'
    + text
    + '</h'
    + level
    + '>\n';
}
Marked.setOptions({
  renderer: renderer,
})
```

PR #703, opened 2015-12-30
Improve parsing of non-pedantic emphasis.

kohler:
The goal of non-pedantic emphasis is to avoid emphasizing
the_middles_of_identifiers. The previous code failed, however, to
do this; for example, in this line

```
one_word_with_underscores another_.
```

emphasis was applied from "underscores" through "another".

Implement a variant of the rules in redcarpet. Redcarpet's rules
say that an underscore matching `/\S_[A-Za-z0-9]/` does not count.


lib/marked.js
```diff
@@ -460,7 +460,8 @@ var inline = {
   code: /^(`+)\s*([\s\S]*?[^`])\s*\1(?!`)/,
   br: /^ {2,}\n(?!\s*$)/,
   del: noop,
-  text: /^[\s\S]+?(?=[\\<!\[_*`]| {2,}\n|$)/
+  text: /^[\s\S]+?(?=[\\<!\[_*`]| {2,}\n|$)/,
+  moretext: /^(?:[^\s([{]_[A-Za-z0-9]|[A-Za-z0-9]_\S)/
```
Feder1co5oave: This regex looks suspiciously specific. Why not `\S` instead of `[A-Za-z0-9]`, for instance?

kohler: The point of `moretext` is to support “no intra-emphasis”, the GFM feature for words_containing_underscores. And that feature's implementation (in redcarpet) uses [A-Za-z0-9]:

https://github.com/vmg/redcarpet/blob/master/ext/redcarpet/markdown.c#L613

joshbruce:
Closing as stale, believe fixed elsewhere, and merge conflicts.

PR #705, opened 2016-01-04
Update default value for sanitize option

RickWieman:
The default value of `sanitize` is `false` according to the description of `sanitize` (line 251 of README.md), but the example still listed `true` as the default value. A small test pointed out that the default value is indeed `false`.


RickWieman:
Hmm, I see this is a duplicate of #481, #624, #634, #651 and #666. Therefore, I'll close this one again.

Issue #722, opened 2016-02-20
feature request : optional MathJax support


Jasper0819X:
Here‘s an example, which works:

https://gist.github.com/3846masa/cde99a18a8d7ad225cf7

rhauteki:
[KaTeX](https://github.com/Khan/KaTeX) is the future of math typesetting for the web.

KaTex vs. MathJax:
![katex-comparison](http://khan.github.io/KaTeX/katex-comparison.gif)

rhauteki:
I've submitted a pull request #723 to ignore everything between $ or $$ to allow the use of KaTeX or MathJaX so the user doesn't have to worry about using _ or * in TeX formulas

MrJasper:
Thanks a lot for your work, and sorry for my late...
I've tested your code in vscode and found a bug:

![bug](https://cloud.githubusercontent.com/assets/15362491/13379482/abf381c6-de61-11e5-83dc-cef6f1034f01.png)

how to fix it? I know little about js.

rhauteki:
Thanks for pointing out, I guess I should've made it a block-level grammar rule, I'll change the code and do so more extensive testing ...

Feder1co5oave:
I implemented this a while ago to use it with MathJax (I'll definitely have a glance at KaTeX since it looks promising). I use it regularly and I find it is quite stable.

https://github.com/chjj/marked/compare/master...Feder1co5oave:0341a2f

Basically you parse your markdown which includes `$ ... $` (inline math) and `$$ ... $$` (block level displaymath) tags, and then run `MathJax.Process` on the container you put your generated html code into.

anaderi:
would be nice to have it merged if no issues are found. (works fine for me so far)

wookayin:
Any updates?

joshbruce:
Closing for now. Please see #956 

See two targeted Markdown specifications.

PR #723, opened 2016-02-20
Safely use TeX code, define keywords and get the attention ...

rhauteki:
So I'm creating this thing to make a simple scientific documentation page. For this I am using markdown but needed to get around the use of markdown characters inside TeX equations, so I decided to fork this parser and extend it, maybe It should be made optional? 

## TeX formulas
Use TeX `$\partial_{\mu} A^\mu = 0$` inside markdown without worrying about the special characters.
```
$$\mathcal{L}=\bar\psi(i\gamma^\mu D_\mu-m)\psi -\frac{1}{4}F_{\mu\nu}F^{\mu\nu}$$
```
Everything in between `$` or `$$` will be ignored and can be rendered by [KaTeX](https://github.com/Khan/KaTeX/tree/master/contrib/auto-render)

***
_These two extensions seemed useful? I don't know ..._
### Define keyword

```
:keyword: a word or concept of great significance.
```

will be rendered to

```html
<dl>
  <dt>keyword:</dt>
  <dd>
      <p>a word or concept of great significance.</p>
  </dd>
</dl>
```

### Alert

```
>! The auto-render extension isn't part of KaTeX proper, it should be separately included! 
```

will be rendered to

```html
<div role="alert">
<p>The auto-render extension isn't part of KaTeX proper, it should be separately included!</p>
</div>
```


rhauteki:
A bug #722 was found, so more extensive testing is required ...

ViktorQvarfordt:
What's the status on this? I'd like to be able to use MathJax or KaTeX with marked, but this issue is a showstopper.

lejafar:
You can check out my changes changes f0fd2644ae96179b3252431e8aaa53d9d5b0ee94, this is an earlier commit as I might have been to fast accepting the pull request. Looking at it right now, it wasn't inline with what I actually wanted to do ...

I basically told marked to ignore everything between two '$' so that I could use '^' and '_' without worries, then I used the [auto-renderer ](https://github.com/Khan/KaTeX/tree/master/contrib/auto-render) to render everything between '$' inline and '$$' in display mode. 

It's probably better to render inside marked by changing renderer methods as there might be a reason why the auto-render extension isn't part of KaTeX proper, something like

``` JavaScript
Renderer.prototype.formula = function(tex) {
    return katex.renderToString(tex);
}
```

ViktorQvarfordt:
I see some issues with this. For instance, it does not handle `\begin{}..\end{}`.

Have a look at my new PR #799 and see what you think.

ViktorQvarfordt:
I dislike the idea of interweaving marked and KaTeX. What about support for other math renderers (like MathJax)? What about if markdown is rendered on the server, and math in the browser? I think the markdown rendering and math rendering should be separate, making marked agnostic to how one renders the math.

Also, note that does have auto-render like MathJax: https://github.com/Khan/KaTeX/tree/master/contrib/auto-render

PR #728, opened 2016-03-14
Update readme example to reflect defaults

nehero:
Changed sanitize default from true to false in the example.


JKillian:
👍  👍  👍  👍 

This really should be merged! People assuming that sanitizing is being done when it's not could lead to big security problems. (I personally think sanitize should be true by default!)

PR #732, opened 2016-03-23
Adding inline only parsing as a configuration option

webfella:
I was looking at using the alternative `markdown-it` as they have a `renderInline` method to only parse inline elements.

I didn't want to pull in their lib due to the extra filesize, so I've made some tweaks to enable this as a configuration option which is disabled by default.


joshbruce:
Closing due to stale, merge conflicts, and not focus at this time. See #956

Issue #10, opened 2011-12-18
Add newline between tokens


markc:
I see on line 506, in parse(), you separate tokens with a space which makes viewing the rendered HTML source really hard to read. If you could change that single space to a newline then it makes all the difference and is no more or less efficient in bandwidth as the resulting filesize is the same.

```
- return out.join(' ');
+ return out.join('\n');
```

PR #740, opened 2016-04-17
Allow to add custom lexer tokens

jskorepa:
Split Lexer.token to Lexer.token and Lexer.tokenLoop

This allows to change tokenLoop and create custom markdown tag.

Usage example ($$equation$$):
````
lexer.rules.equation = /^\$\$(((?!\$\$).)*)\$\$/;

lexer._tokenLoop = lexer.tokenLoop;
lexer.tokenLoop = function(src, top, bq) {
  var cap;
  if ((cap = this.rules.equation.exec(src))) {
    if (cap[0].length > 1) {
      this.tokens.push({
        type: 'equation',
        text: cap[0]
      });
    }
  }

  return lexer._tokenLoop(src, top, bq);
}
````

Then you can change Parser.tok:
````
var parser = new marked.Parser();

parser._tok = parser.tok;
parser.tok = function() {
  if(this.token.type == 'equation') {
    return this.token.text;
  } else {
    return parser._tok();
  }
}
````

This is simple way to allow chjj/marked#379 as well as chjj/marked#564 and chjj/marked#425

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -160,14 +160,14 @@ Lexer.prototype.token = function(src, top, bq) {
 
 Lexer.prototype.tokenLoop = function(src, top, bq) {
   var next
-   , loose
-   , cap
-   , bull
-   , b
-   , item
-   , space
-   , i
-   , l;
+    , loose
+    , cap
+    , bull
+    , b
+    , item
+    , space
+    , i
+    , l;
 
   if (cap = this.rules.newline.exec(src)) {
     src = src.substring(cap[0].length);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by parleur** [comment]
Nice work !
However I think it slows down the lib a bit ( I only performed partial bench because robotskirt doesn't compile straightforward in my system). Have you performed full benchmark and could you publish it?
Also, one test fails:
```
#10. def_blocks.text failed at offset 20. Near: "<blockquote><p>hello</p><p><ahref="foo">1</a>:hell".
Got:
<blockquote><p>hello</p><p><ahref="foo">1</a>:hell

Expected:
<blockquote><p>hello[1]:hello</p></blockquote><hr>
```

joshbruce:
Closing as out of scope for Marked. See #956

PR #741, opened 2016-04-19
feature: make anchor hash name support muti-languages, do not lost chars

humanhuang:
this can be support muti-languages

```
Renderer.prototype.heading = function(text, level, raw) {
  return '<h'
    + level
    + ' id="'
    + this.options.headerPrefix
    + raw.toLowerCase().replace(/[\s\r\n]+/g, '-')
    + '">'
    + text
    + '</h'
    + level
    + '>\n';
};
```

=>>> replaced

```
Renderer.prototype.heading = function(text, level, raw) {
  return '<h'
    + level
    + ' id="'
    + this.options.headerPrefix
    + raw.toLowerCase().replace(/[^\w]+/g, '-')
    + '">'
    + text
    + '</h'
    + level
    + '>\n';
};
```

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -797,7 +797,7 @@ Renderer.prototype.heading = function(text, level, raw) {
     + level
     + ' id="'
     + this.options.headerPrefix
-    + raw.toLowerCase().replace(/[^\w]+/g, '-')
+    + raw.toLowerCase().replace(/[\s\r\n]+/g, '-')
     + '">'
     + text
     + '</h'


toastsgithub:
Hey, is it necessary to replace all `^\w` chars with '-'? I think may be just replace all blanks with '-' will be ok. Plus, my language environment is English, so `\w` will not match the chinese chars, which means that all the titles with chinese in it will not properly parsed, all the chinese chars will lost, I replace this RegExp with /\s+/, I just want to know is that good ? Or is there any better workaround?

joshbruce:
This concept of allowing UTF-8 characters has come up a couple of times while I've been cleaning up the PRs. May be something worth turning into an issue: @Feder1co5oave and @UziTech - thoughts??

Closing due to missing tests - see #956

PR #747, opened 2016-05-05
improve heading render with unicode header content

4Oranges:
When the header is pure unicode content, the id of the header element would be a single dash(-). This patch fixes it.


joshbruce:
See #741 

Closing due to merge conflicts and not the focus at this time per se.

PR #761, opened 2016-06-18
Fix bug in test

parleur:
Bug in test due to case.


joshbruce:
@Feder1co5oave and @UziTech - I think I can merge this, but losing track of some things in-memory. Showdown's converter method is capitalized?? Help me out??

Feder1co5oave:
I think this was already fixed elsewhere. You can skip

PR #780, opened 2016-08-02
GFM paragraph lexer should use GFM heading regex

singpolyma:
GFM paragraph regex was checking for headings in the non-GFM format,
which caused an extra paragraph break to be inserted, for example, in
the text:

```
print_hi('Tom')
#=> prints 'Hi, Tom' to STDOUT.
```


joshbruce:
Closing as relatively stale, missing tests, and merge conflicts.

PR #797, opened 2016-08-27
Fix a bug in lexer which may cause empty table drop last cell

nekolab:
The regex in marked.js:388 tends to trim the trail '\n' in markdown
table by using non marked match(?:). But it also destory the table
structure itself by mistake. So fix it by keeping the table structure.


matt-:
Any tests for this?

joshbruce:
Closing as missing tests - associating with #980 for future reference.

PR #798, opened 2016-08-27
Add Feature - Support `Promise`

crazyguitar:
Hi,

I made this project support `Promise` . Consider the following scenario:

``` js
var marked = require('marked')
    , fs = require('fs');

renderer.heading = function (text, level) {
  var escapedText = text.toLowerCase().replace(/[^\w]+/g, '-');
  // need read file, but we cannot using fs.readFile(filename, enc, cb).
  // We cannot use  any async functions.
},
console.log(marked('# heading+', { renderer: renderer }));
```

then we can solve this by using `Promise` object:

`````` js
var marked = require('./index')
  , fs = require('fs')
  , util = require('util');

var path = "layout.html";

renderer.heading = function (text, level) {
  return new Promise((resolve, reject) => {
    fs.readFile(path, 'utf8', (err, data) => {
      try {
        resolve(util.format(data, text));
      } catch (e) {
        reject(e);
      }
    });
  });
},

marked('# Heading2\n```js\nconsole.log("hi")```\n', { promise: true, renderer: renderer })
.then(function(html) {
  console.log(html);
}).catch(e => {
  console.log(e);
});
``````

and I can get the result like:

``` bash
<h1>Heading2</h1>
<pre><code class="lang-js">console.log(&quot;hi&quot;)
</code></pre>
```

The `Promise` object solve the problem which I cannot call async functions in render function. Hope this feature can make the project better.

thanks,
chang-ning


nchase:
This would be really useful to me.

Feder1co5oave:
This may work in node, but not in the browser. IMO, it is out of scope.

joshbruce:
Closing due to merge conflicts and missing tests (not quite sure how you'd test a promise). Having said this, it might be interesting if Marked could be multi-threaded (more concurrent); not sure if it already is or not.

Issue #336, opened 2014-01-21
Missing pre code style with highlightjs version 0.8 


jlchereau:
highlightjs version 0.8 stylesheets have replaced "pre code" style with ".hljs" style.
So marked.js should at least initialize code blocks as

``` xml
<pre><code class="hljs">
```

and not simply

``` xml
<pre><code>
```

At most, marked should also consider source code language class.

PR #799, opened 2016-08-28
Feature: Raw pass-through of math for MathJax / KaTeX etc.

ViktorQvarfordt:
Add an option `math` which, when enabled, disables all processing of math wrapped in `$..$`, `\(..\)`, `\[..\]`, `$$..$$` or `\begin{}..\end{}`. For use with MathJax / KaTeX etc.

Similar to #722 and #723, but better (in my opinion).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ marked.setOptions({
   pedantic: false,
   sanitize: false,
   smartLists: true,
-  smartypants: false
+  smartypants: false,
+  mathDelimiters: false
 });
 
 console.log(marked('I am using __markdown__.'));
@@ -220,6 +221,21 @@ Type: `boolean`
 Default: `true`
 
 Enable [GitHub flavored markdown][gfm].
+
+### mathDelimiters
+
+Type: `Array`
+Default: `null`
+
+Enable raw pass-through of math wrapped in specified delimiters such as `$..$`, `\(..\)`, `\[..\]`, `$$..$$` and `\begin{}..\end{}`.
+For use with MathJax / KaTeX etc.
+A standard setup would use
+
+```
+mathDelimiters: [['$', '$'], ['\\(', '\\)'], ['\\[', '\\]'], ['$$', '$$'], 'beginend']
+```
+
+Note the special delimiter option `beginend`, this properly handles any `\begin{}..\end{}` with possible nested environments inside.
 
 ### tables
 
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -12,7 +12,7 @@
 
 var block = {
   newline: /^\n+/,
-  code: /^( {4}[^\n]+\n*)+/,
+  code: /^\n( {4}[^\n]+\n*)+/,
   fences: noop,
   hr: /^( *[-*_]){3,} *(?:\n+|$)/,
   heading: /^ *(#{1,6}) *([^\n]+?) *#* *(?:\n+|$)/,
@@ -103,6 +103,12 @@ function Lexer(options) {
   this.tokens = [];
   this.tokens.links = {};
   this.options = options || marked.defaults;
+  // Sort math delimiters, because parsing of $$..$$ must come before $..$ etc.
+  if (Array.isArray(this.options.mathDelimiters)) {
+    this.options.mathDelimiters.sort(function(a, b) {
+      return b[0].length - a[0].length;
+    });
+  }
   this.rules = block.normal;
 
   if (this.options.gfm) {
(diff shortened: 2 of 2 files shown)


ViktorQvarfordt:
Perhaps it would be nice to allow more fine-grained control by changing the option `math` to an `Array` (instead of `Bool`) that accepts delimiters, such as `math: [["$", "$"], ["\\[", "\\]"]]`. However, `\begin{}..\end{}` is more complicated, perhaps the option for this would simply be `"beginend"`, i.e. `math: [["$", "$"], ["\\[", "\\]"], "beginend"]`.

I'll have a go at this later. However, I don't know how to write tests for a non-bool options in with existing test convention.

ViktorQvarfordt:
The commit has been updated to use the above discussed more flexible variant.

ViktorQvarfordt:
If we get this merged, it would be nice to add marked to this list http://docs.mathjax.org/en/latest/misc/mathjax-in-use.html.

anaderi:
looks like a cool addition! any objections for merging it?

akof1314:
@ViktorQvarfordt Thanks. BTW: escape begin delimiter.  eg.  `\$2$`    is not a math.

wookayin:
It looks goods anyway. Any updates, BTW?

j16180339887:
@wookayin I found that [markdown-here](https://github.com/adam-p/markdown-here) forked this project and it support LaTeX.

Then I copied his [marked.js](https://github.com/adam-p/markdown-here/blob/master/src/common/marked.js), then
```
marked.setOptions({
        renderer: new marked.Renderer(),
        gfm: true,
        tables: true,
        breaks: true,
        pedantic: true,
        sanitize: true,
        smartLists: true,
        smartypants: true,
        math: function mathify(mathcode) {
            return '<img src="https://chart.googleapis.com/chart?cht=tx&chl={urlmathcode}" alt="{mathcode}">'
                    .replace(/\{mathcode\}/ig, mathcode)
                    .replace(/\{urlmathcode\}/ig, encodeURIComponent(mathcode));
        }
});
```

Usage: ${YOUR_LaTeX_Code}$
Ex:
`$E = \frac{mc^2}{\sqrt{1-\frac{v^2}{c^2}}}$`

;)

j16180339887:
If you want to keep it offline, you should use this PR.

hgustafsson:
@ViktorQvarfordt Great work!

Since you are already parsing the input, do you think you could return the delimeters and the contained TeX separately? This would allow for easy renderers like:

```javascript
renderer.math = function (begin, tex, end) {
    if ( begin === "$" || begin === "\\(" )
        return katex.renderToString(tex, displayMode = false);
    else if ( begin === "$$" || begin === "\\[" )
        return katex.renderToString(tex, displayMode = true);
    else
        return katex.renderToString(begin + tex + end, displayMode = true);
    }                
}
```

without further parsing.

ViktorQvarfordt:
That's a cool idea! I've updated the PR to reflect this. But I'm lazy and ignored extracting the `begin` and `end` tags properly, see the code, there is a TODO on [line 617](https://github.com/ViktorQvarfordt/marked/blob/master/lib/marked.js#L617) . Feel free to do this!

Note: I also rebased this PR on top of the latest chjj/marked.

joshbruce:
Closing due to merge conflicts - flagging with #936 for future reference.

julien-c:
Only a partial solution to the use case described in this PR, but if you're doing server-side rendering of Math (for instance using KaTeX), why don't you:
- first parse the math,
- replace it with HTML,
- and finally render using marked?

UziTech:
Katex has been something people have been asking for for a while so I just created an extension and published it to npm. [marked-katex-extension](https://www.npmjs.com/package/marked-katex-extension)

Demi1024:
> Hello @chjj, Please solve this problem. I'm suffering `_` rendered to `\<em\>` ... 😅 #336

same here，any update？

Demi1024:
> @Demi1024 Underscores around text are supposed to turn into `\<em\>` tags. This is part of the Commonmark spec. https://spec.commonmark.org/0.30/#emphasis-and-strong-emphasis
> 
> If that is not what you are referring to, could you be more specific, or create a new issue describing your problem?

I don't want to break the syntax of Markdown, but is it possible to skip the parsing of underscores when inserting mathematical formulas in Markdown?

 Parsing underscores as emphasis will cause the conversion of mathematical formulas to fail, as shown in the image below.
image

Is it possible to output mathematical formulas as they are without any conversion, for example, when encountering segments that start and end with `$$`?

UziTech:
Or if you don't want to use katex you can also use code blocks or html blocks, both of which won't convert the contents with markdown.

```html
<span>
_test_
</span>
```

produces:

<span>
_test_
</span>

PR #801, opened 2016-08-31
Support HTML5 element id

Dafrok:
In HTML5, you can take any unicode and use them as values for id
attributes. It’s important for non-english languages. For example:

``` markdown
# 蛤蛤
```

will be complied by this lib into:

``` HTML
<h1 id="-">蛤蛤</h1>
```

however, the majority of us want it to be:

``` HTML
<h1 id="蛤蛤">蛤蛤</h1>
```


joshbruce:
Closing due to merge conflicts and not necessarily the focus at this time - flagging with #981 for future reference.

PR #807, opened 2016-09-12
Don't redeclare parameter variable

jart:
This causes the Closure Compiler to fail in strict mode.


joshbruce:
Closing due to merge conflicts and believe corrected elsewhere.

PR #809, opened 2016-09-15
gfm headings regex should allow no space after hashes

epoch:
```
# heading
#heading
```

headings with no space between hash and words should work with github flavoured markdown


josephernest:
Any news about this @chjj ? This would be super useful.

Feder1co5oave:
This is not compliant with the [spec](http://spec.commonmark.org/0.28/#example-33)

joshbruce:
@Feder1co5oave: Closing on stale, missing tests, and not complient with Commonmark nor GFM: https://github.github.com/gfm/#example-34

Feder1co5oave:
@joshbruce oh wow, another spec sheet to follow! Yay! :laughing:

joshbruce:
lol - See #956 - Been there since I got here. ;)

At least we're following specs!

PR #814, opened 2016-09-25
Parser fix: angle brackets misidentified as tags

### LINKED ISSUE #813: Autolink following an open angle bracket can be misinterpreted as tag
If an angle bracket appears inline before an autolink, and if content follows immediately after the angle bracket, as for example:

```
<1 <http://example.com>
```

the parser will confuse the first angle bracket as the opening of an HTML tag. Instead of producing:

```
&lt;i <a href="http://example.com">http://example.com</a>
```

marked leaves it untouched:

```
<i <http://example.com>
```

thisgeek:
Fixes #813. Tests included.


KostyaTretyak:
Code with your PR not working. Nor test, nor regexp.

joshbruce:
Closing in favor of #977

Issue #266, opened 2013-10-15
Feature request: Table column width


iegik:
So, we have columns, for example:

```
| Some | text | and number
| ---- | -------- | ----
| some | long text | 1
| awesome | another text | 1
```

| Some | text | and number |
| --- | --- | --- |
| some | long text | 1 |
| awesome | another text | 1 |

Now, I need, that second column will be larger and last - smaller. Obviously, that we look at line that goes after headers: if shorter stroke - shorter column.

Solution:
So, count the dashes (with space and column separator) and use percentage width's by the ratio!
For example `table { width:100%;}` (Total: 24chars), `<th style="width:29.17%">` (first column - 7chars),  `<th style="width:45.83%">` (second column - 11chars),  and leve last without width's. http://jsfiddle.net/iegik/3D2Fb/

crifan:
+1 expect can designated table column width, by dash count or something else

joshbruce:
I'm with @kuba-kubula on this one, I think - also @ghost before deleting the first line.

1. Want to keep Marked light and out of the "template engine" business; adding attributes to the markup feel more like template engine (or CSS) business. (I don't know of a Markdown spec that allows for sizing of table columns.)
2. From a UX perspective mobile becomes an interesting problem space - especially if we're using percentages.
3. @jondlm also raises a good point regarding making an extension or override to handle the use case as it's not really markdown in the strictest sense. (Column spanning might be something worth pursuing - don't remember if the Multi-Markdown or CommonMark specs cover that.)

Going to leave this and PR open for now, but not feeling this as being a problem space to pursue solving with Marked (see #956).

UziTech:
This is not part of any spec so it will have to be an extension

maor-rosenfeld:
For anyone facing this issue, just use a &nbsp-s (non breaking space) to extend the header of the column you want to make wider. You can copy it from here: " ".

iegik:
What if i want make short column, where header is wider?

| I want make this column with that long text in header shorter then next column | Long column &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; |
| - | - |
| 1 | 2 |

maor-rosenfeld:
[...]

PR #816, opened 2016-10-12
Table width adjustment by dash character

swcho:
Hi, I have little idea to determine table column width by dash characters.

For example, if you have next table.

```
Header 1|Header 2|Header 3|Header 4
:---    |:--:    |--:     |---
Cell 1  |Cell 2  |Cell 3  |Cell 4
*Cell 5*|Cell 6  |Cell 7  |Cell 8
```

Each column will have 30%, 20%, 20%, 30% width.
The percentage value of width is determined by (row dash count / total dash count) \* 100.
So, if you want to have the same column width, you can adjust dash as below

```
Header 1|Header 2|Header 3|Header 4
:-    |:-:    |-:     |-
Cell 1  |Cell 2  |Cell 3  |Cell 4
*Cell 5*|Cell 6  |Cell 7  |Cell 8
```

Last but not least, to use this feature you can turn on flag "tableWidth" through options.

I hope this is useful.
Thanks.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -213,10 +213,16 @@ Lexer.prototype.token = function(src, top, bq) {
         type: 'table',
         header: cap[1].replace(/^ *| *\| *$/g, '').split(/ *\| */),
         align: cap[2].replace(/^ *|\| *$/g, '').split(/ *\| */),
-        cells: cap[3].replace(/\n$/, '').split('\n')
+        cells: cap[3].replace(/\n$/, '').split('\n'),
+        width: []
       };
 
+      var widthSum = 0;
+      var width;
       for (i = 0; i < item.align.length; i++) {
+        width = item.align[i].split('-').length - 1;
+        widthSum += width;
+        item.width[i] = width;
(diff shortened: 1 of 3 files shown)


joshbruce:
See #980

iegik:
Duplicate of https://github.com/chjj/marked/issues/266

iegik:
I like, that You also see the benefits from this feature.
As already discussed in #266, and my expectations are:
- It should not *depends too much* on *viewport size*
- It should not contain additional formatting
- It should narrow sell with big content if column marked as shorter
- It should leave column longer, even if content of cell is short when column marked as longer

The counting dashes, for me is the best solution for calculating relative sizes of columns.
As I see, the Markdown main goal is represent formatted text in text form. So, total dash + spaces should  represent approximate table width.
`width: 100%` - is strongly depends of viewport size, we can avoid using of this, by reverse solution: limiting cell sizes by `min-width`.
Any ideas?

joshbruce:
Closing as not spec-based (don't believe GFM tables has this functionality). Marking as NFU for possible reconsideration during 2.x development.

Issue #746, opened 2016-05-02
Modularization of library


Nalinc:
There is a rising demand to move towards modularization of **marked ** and separate different components in different files(grammar-rules, lexers, parsers and renderers). This though not have been mentioned explicitly, is evident from issues #743, #717, #704.

A properly modularized code would make maintenance easy and also allow addition of new grammar, implementation of custom parser, renderer and would address separation of concerns properly.

Like may be we can have an `/src` folder to fiddle with the code in development and use `/lib` for deliverable file.

joshbruce:
@Nalinc: See #956 

This idea is intriguing to me, primarily from the perspective of making it easier for the library to communicate what and how it works from an architectural perspective.

I don't know enough about Marked to do this work myself. What would you propose as a full folder structure given what's already here?

Nalinc:
hey @joshbruce 

I worked on marked a bit in past and even changed some of its functionalities based on my needs back then. One thing I found annoying was keeping track of things in the same file even if I was working on a different component. Here's what I did back then:
- I moved each of the primary components in a separate file in `/src`. This way I was left with core.js, grammar.js, lexer.js, parser.js and renderer.js in `/src` folder.
- Each of these components was written in IIFEs so they weren't messing with global scope.
- Finally, I had a grunt task that'd combine these components from `/src` and create a bundled file in  /dist`

I found it easy because having independent components in different files, I could easily change/extend one component without worrying about others. The code became smaller, readable and more manageable too. I feel comfortable with marked and am open to make a PR on this. 
Regarding #956, is there an exiting card for this or should I make one and move it to "under consideration"?

joshbruce:
#420 

#660

joshbruce:
@Nalinc: Sorry it's taken a moment - I totally didn't see (or catch) your comments and questions at the bottom.

That architecture sounds about right to me; given what I've seen in the issues and PRs. Let me tag in a few of the other recently active community members to weigh in as well (@UziTech, @Feder1co5oave, @KostyaTretyak). There's not an existing card beyond this one; so, congrats on that. :)

[...]

PR #818, opened 2016-10-26
Pass header flag to Renderer.tablerow method

ksheedlo:
When styling tables out of a custom renderer subclass, it's useful to
know if the tablerow is in the header or body. Tablecells already get a
header flag that is set if the tablecell is in the header. This change
adds the flag to tablerows.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1009,7 +1009,9 @@ Parser.prototype.tok = function() {
           { header: true, align: this.token.align[i] }
         );
       }
-      header += this.renderer.tablerow(cell);
+      // header flag is not used in Renderer.prototype.tablerow, but passed
+      // here to make available to Renderer subclasses.
+      header += this.renderer.tablerow(cell, {header: true});
 
       for (i = 0; i < this.token.cells.length; i++) {
         row = this.token.cells[i];
@@ -1022,7 +1024,7 @@ Parser.prototype.tok = function() {
           );
         }
 
-        body += this.renderer.tablerow(cell);
+        body += this.renderer.tablerow(cell, {header: false});
       }
       return this.renderer.table(header, body);
     }


joshbruce:
Closing due to missing tests - going to flag with #981 for future reference as we focus on correcting known issues.

Feder1co5oave:
This specific pr is not about parsing, so there's nothing to test here.
I can't see a practical use for this, but I guess it would be nice to have for someone. I'd flag it for the [0.6.0 milestone](https://github.com/chjj/marked/milestone/3)

joshbruce:
Is this a GFM table *extension* - think all the table stuff in GFM is an extension; so, definitely not an immediate need.

Feder1co5oave:
Tables are supported in gfm mode only, and yes, the _whole_ gfm is an extension to markdown, but not github-specific. gfm is **on** by default.

Must remember to adjust the `tablerow` renderer and put in the additional parameter, for sake of "documentation".

joshbruce:
It's possible. I think what I meant when asking if it was an extension is the GFM spec has certain concepts that are explicitly labeled as extensions - not part of their base specification, which is an extension of the original Markdown.py: https://github.github.com/gfm/

Task lists, tables, and autolinks, for example, are explicitly called out as extensions. Think the focus for us right now is on the base specs for CommonMark and GFM - not extensions - at least according to #956.

Think you're probably right about the possibility of helping to inform the 0.5.0 work. Changing.

ksheedlo:
> I can't see a practical use for this

I'm using it in my fork to put a thicker border under the `thead > tr`, like this:

screen shot 2018-01-24 at 2 47 44 pm

I suppose I could hack it by setting that border on the `th` elements instead, but doing it this way seems nicer.

Feder1co5oave:
@joshbruce there is no "gfm markdown", gfm is commonmark + gfm extensions (i.e. tables, task lists, autolinks, etc.)

joshbruce:
@Feder1co5oave: Interesting. Did not realize. Thought they were two separate specs - makes life easier then. :)

@ksheedlo: Seems odd to ask Marked to do that - tightly couples markup and style, if I'm understanding the resulting markup properly. Can you post the markup that gives you that result?

joshbruce:
@Feder1co5oave: #746 - As we consider architecture and extensibility we should start considering what we want Marked to take on beyond the specs a bit more. If we physically separate the parser and whatnot as desicribed by @nalinc it might become more obvious on how to override/extned; thereby, opening the door to folks to do things like what's in #1030 or here instead of asking Marked to do it - and for us to maintain it. Really don't want to end up in the same situation again where none of us really have time to maintain Marked because it is taking on too much, you know?

PR #819, opened 2016-10-26
List Start: Passes start number to render

woodyrew:
Currently there's no way to get the number of the list item.
Backwards compatible whilst adding functionality. Let me know if you want any changes

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -283,7 +283,7 @@ Lexer.prototype.token = function(src, top, bq) {
     if (cap = this.rules.list.exec(src)) {
       src = src.substring(cap[0].length);
 
-      bull = /^\d+/.exec(cap[2]);
+      bull = /^(\d+|[a-h]|[ivx]+)/i.exec(cap[2]);
       bull = bull && bull[0];
 
       this.tokens.push({
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -283,7 +283,7 @@ Lexer.prototype.token = function(src, top, bq) {
     if (cap = this.rules.list.exec(src)) {
       src = src.substring(cap[0].length);
 
-      bull = /^(\d+|[a-h]|[ivx]+)/i.exec(cap[2]);
+      bull = /^\d+/.exec(cap[2]);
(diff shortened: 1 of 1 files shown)


Jiang-Xuan:
Are you sure this can work?

woodyrew:
I'm using this in production, so, yes.

Jiang-Xuan:
Are you trying to do ol element is generated with a start attribute?But I did not see you to modify the Renderer class.

PS: I am a China boy is not at English.Please don't keep in mind if i say something wrong😀.

woodyrew:
Yes, you need to modify the renderer, e.g.
```js
renderer.list = function (body, ordered) {
    const type = ordered ? 'ol' : 'ul';
    const attrs = {
        type : '',
        start: ''
    };

    if (+ordered || +ordered === 0) {
        if (+ordered !== 1) {
            attrs.start = `start="${ordered}"`;
        }
    }
    else {
        const type_case = (ordered === ordered.toUpperCase()) ? 'A' : 'a';
        attrs.type = `type="${type_case}"`;
        const start_letter = ordered.toString().toUpperCase().charCodeAt(0) - 64;
        attrs.start = (start_letter > 1 && start_letter < 9) ? `start="${start_letter}"` : '';
    }
    const attr = [attrs.type, attrs.start].filter(x => x).join(' ');

    return `<${type} ${attr}>
    ${body}
    </${type}>
    `;
};
```

Jiang-Xuan:
I understand.You are awesome👍, thank you.

joshbruce:
Closing due to merge conflicts.

Also, I will adopt the majority view on this; however, my interpretation of Markdown is more imperative/functional in its design. Write the markdown -> convert to html. There aren't a lot of options to play with it using a parser. Instead, we can interact with the output HTML after it has been rendered.

Feder1co5oave:
Ok, the idea was nice but I think the PR breaks the list rendering.
I would have added a third argument to the list renderer containint the actual bullet, and left the `ordered` one unchanged (so it is backward-compatible).

I'd flag this for extensibility / developer experience

PR #821, opened 2016-11-11
Add markdown extension mechanism and syntax

rumkin:
I've added markdown extension mechanism. It uses plugins with such syntax:

```
@plugin()
    Plugin data block
    ...
    is indentation based
    so it could include any data
    like csv, xml and even markdown itself
```

Examples:
```
@youtube(Hy-sVByUHqE)

@csv(Id, Name)
    1, Alice
    2, Bob

@uml(sequence)
    Alice->Bob: Hi, Bob.
    Bob->Alice: Hi, Alice.

@datetime()

@iframe(src)
```

Plugins are defined with Renderer:

```javascript
let renderer = new marked.Renderer({
    plugins: {
        video(params, block) {
              return `<video src="${params}"/>`; 
        }
    },
});
let html = marked.parse('...', {plugins: true, renderer});
```

By default if plugin not found it should be ignored.


OverFlow636:
using this on my fork and it works great!

jHoldroyd:
Just tried this out and its great but there are a few issues.

Firstly it throws `Cannot read property 'replace' of undefined` on line 204 of lib/marked.js which seems to be triggered when you have tags without a line break under them due to it looking for block content.

The second issue we've run into is that if you have a number of the same tags after each other it "eats" them. That is to say without block content under each of them they disappear. An example is:

```
@link(small, tag, Inventory, /inventory)

@link(small, shopping-cart, Cart, /cart)
```

Will only show the first link and if there were additional items under the second one, in our case 8 in total it would still only show the first one.

Apart from these issues the system works really well and if I get a chance I'll look into a resolution if you don't get there first.

rumkin:
@jHoldroyd Thanks. Fixed this.

naivefun:
Nice. Maybe add a plugin examples folder for demo?

rumkin:
@naivefun Do you expect to see exact plugin example?

naivefun:
@rumkin actually it is very straightforward, but still if there are very common use cases like @youtube, @vimeo, even @tweet, that would be great for beginners to learn.

rumkin:
@naivefun Done.

jHoldroyd:
Just following up on the previous problem we reported, happy to say the initial error has been resolved since you pulled in my suggestion, the "eating" of tags is still occasionally happening though.

I managed to get around it by updating the plugin regex on line 98-100 to:

``` js
block.extra = {
  plugin: /^@!?([A-Za-z0-9_-]+) *\( *((.+) *)?\).*?\n((\s+){2,}.*\n(((\5).*\n)|\n)*)?/
};
```

This allows the script to catch all indented text below a plugin tag but nothing that isn't nested. An example of this fix can be seen here: https://regex101.com/r/76YMYB/1

rumkin:
@jHoldroyd There is one issue with that regexp it will not match single space.

```
@text
 Hi there
```

This will not be matched, it mean that single tab would not be matched too. Also I've replaced `(\s+){2,}` with `(\s{2,})`.

Thanks for help.

rumkin:
I realise no-plugin block which should return html/text if plugin is not defined:

```
@!plugin()
  There is no plugin "plugin".
@!youtube()
   Video on <a href="http://youtube.com/asd123">Youtube</a>.
```

It currently doesn't support markdown itself.

joshbruce:
Closing due to merge conflicts.

See #1026 as well.

PR #826, opened 2016-11-29
Fix the regexes for thematic breaks

kythyria:
Per the [CommonMark spec](http://spec.commonmark.org/0.26/#thematic-breaks) and GFM, you get a `<hr>` only if all three characters are the same ([Example 26](http://spec.commonmark.org/0.26/#example-26)).

This makes that change.


Feder1co5oave:
+1
This is correct, I'd like to review this, resolve conflicts and reopen it.

joshbruce:
Do it up @Feder1co5oave - would love to merge this in - I always thought `<hr>` was seriously overlooked in defining documents.

PR #828, opened 2016-12-06
Allow the own escape/unescape functions as an option

tagoh:
I want to have the own escape/unescape functions instead of the internal one.


abrahamcuenca:
@UziTech This PR should probably be closed as well as part of the issues cleanup.

UziTech:
I was keeping this open since this might be something we want to consider in the future, but If anyone still needs this feature they can create an issue and we can discuss adding the option.

Issue #514, opened 2014-11-19
How to add custom html tags


lordfriend:
I want to add some custom html tags like this:

`````` html
<example>
    <file name="index.html">
    ```html
      <div class="example">
      example code here
      </div>
    ```
    </file>
</example>
``````

but it seems `<example>` and `<file>` tag are not recognised as a valid html tag. marked think they are paragraph elements and add a couple of `<p>` around them

the compiled content is:

``` html
<p><example>
    <file name="index.html"></p>
    <pre><code class="lang-html">
      &lt;div class="example"&gt;
      example code here
      &lt;/div&gt;
    </code></pre>
    <p></file>
</example></p>
```

This behave make the `<example>` and `<file>` tag fail to work

brunowego:
This is very inconvenient, I am creating a documentation for my project which involves displaying the element, however unfortunately is automatically added a paragraph around.

My Angular directive:

```html
<pf-panel>
    <pf-body>
        <p>Exemplo de painel com corpo.</p>
    </pf-body>
</pf-panel>
```

The results:

```html
<p><pf-panel>
    <pf-body>
        <p>Exemplo de painel com corpo.</p>
    </pf-body>
</pf-panel></p>
```

Perhaps not the purpose of this project ...

brunowego:
@chjj if you have a little time, please take a look on this issue :+1:

joshbruce:
#985

ghdoergeloh:
Sorry to comment on this old issue, but I cannot find what happend to this issue/feature request.
I found some Issues like #283 and the PR #833 but that was not merged.
I still have the problem, that custom html tags are not recognised by the parser. Why aren't they handled like normal html?

ghdoergeloh:
Okay, partially it is working:
```
> marked.parse('<div-test></div-test>') // didn't work
'<p><div-test></div-test></p>\n'

> marked.parse('<div-test>\n</div-test>') // worked
'<div-test>\n</div-test>'

> marked.parse('<div-test>test</div-test>') // didn't work
'<p><div-test>test</div-test></p>\n'

> marked.parse('<div-test>\ntest\n</div-test>') // worked
'<div-test>\ntest\n</div-test>'
```

UziTech:
That is because there is a difference between an [HTML block](https://spec.commonmark.org/0.29/#html-blocks) and [inline HTML](https://spec.commonmark.org/0.29/#raw-html).

[...]

Issue #283, opened 2013-11-12
Does not recognise custom html elements


timoxley:
### tldr;

[Custom html elements](www.html5rocks.com/en/tutorials/webcomponents/customelements/) are coming, and `marked` does not currently recognise them as valid html. Custom elements may have different escaping requirements for their content based on whether the elements contains more elements (like `div`) or plain text (like `script`). Perhaps it would be possible to pass which html tags to escape/not escape as configuration properties.
### Backstory

I've been experimenting with custom elements and I wanted to inject some into [my markdown presentation](https://github.com/timoxley/shadowdom-talk-slides).

`````` markdown
## Shadow DOM is a sub-tree inside a DOM node
#### Like a mini-iframe that replaces an Element.

<h3>Light DOM</h3>

```html
<h3>Light DOM</h3>
`` `
<play-script>
var heading = currentSlide.querySelector('h3')
var root = heading.createShadowRoot()
root.innerHTML = '<a href="#">Shadow DOM</a>'
</play-script>
``````

[Source](https://github.com/timoxley/shadowdom-talk-slides/blame/b260937e10764f363aba28c386cc40ad8f93c9ba/slides.md#L19-L31)

[`play-script`](https://github.com/timoxley/shadowdom-talk-slides/blob/b260937e10764f363aba28c386cc40ad8f93c9ba/client.js#L95-L111) is essentially a `script` tag that renders its content, highlighted with [highlight.js](https://github.com/isagalaev/highlight.js), and executes only when you click on it.

Problem: **currently `marked` does not recognise tags with hyphens in them as valid html**. To get around this quickly, I loosened the html detection regex to also match tags with a single hypen:

[Source](https://github.com/timoxley/marked/compare/feature;custom-html-elements#diff-81ab0a5aa39b7a91951fc28d2d151496L44)

``` diff
-  + '|span|br|wbr|ins|del|img)\\b)\\w+(?!:/|@)\\b';
+  + '|span|br|wbr|ins|del|img)\\b)[\\w-]+(?!:/|@)\\b';
```

This successfully recognised my custom tag `play-script` as html, but it raised a new issue:

My custom tag's content was text as opposed to more xml tags, so **marked would html-escape all of my script contents**. This was no good since I wanted that text to remain as executable script (perhaps I should have looked at unescaping).

So as a hack to make it work, I reluctantly added my specific tag, (and the `template` tag) as special-cased to avoid escaping, along with `style`, `pre` and `script`:

``` diff
[...]

PR #833, opened 2016-12-13
Allow html tags to contain hyphens

gavmck:
We need HTML tags with hyphens for a project with some new Angular components in so I've updated the regex for tag detection.

Should help out with #514 and #283


Feder1co5oave:
#985

joshbruce:
Closing due to merge conflicts, not focus for 0.x, missing tests, and open for reconsideration during 1.x or 2.x.

PR #836, opened 2016-12-27
Remove redundant description of direct access to lexer/parser.

jschulenklopper:
The README contained two descriptions for "direct access to the lexer and parser if you so desire." One of those can be removed.


joshbruce:
Seems fixed elsewhere.

PR #841, opened 2017-01-14
Hacked the parser to not wrap in P tags if only one line

jhnsntmthy:
Maybe this would be better as a configurable option, but I hacked the parser to NOT wrap the block in <p> tags unless there is more than one line, then it functions as normal. 

Using this for my i18n CMS so that language editors can edit blocks inside header tags and others, where we know it will be one line and having a P tag is undesireable.

Welcoming your thoughts and feedback.


dbaldon:
I like this idea and would use it if the PR is approved. I do think it should be configurable.

joshbruce:
Closing due to merge conflicts, missing tests, and believe this might actually break the CommonMark spec. Might be something worth considering again during 1.x or 2.x.

PR #846, opened 2017-01-21
Added modular plugin system

Drew-S:
Created a basic modular plugin system for adding in custom modules.

This needs to be tested more. I created a single custom test to see how it operated. Was unable to figure out how to add it to the standard test suite. This arose, I believe, because my system cannot simply be run strait like marked does in the currently existing test setup, it requires making a call to register a new module before running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,143 @@ Type: `function`
 
 The callback function to call when using an async highlighter.
 
+## Custom modules
+
+Using marked registrar you can add custom modules with:
+
+`marked.register(module);`
+
+### custom module structure
+
+Your custom module has three sections:
+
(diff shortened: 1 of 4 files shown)


rumkin:
@Drew-S Good idea, vote for this! That's what I needed implementing plugins #821.

KostyaTretyak:
This can be made even easier for users. Checkout [this one](https://github.com/KostyaTretyak/marked-ts#example-of-setting-a-simple-block-rule)

joshbruce:
Interesting idea. Might be worth considering as it relates a bit to #746 

Closing due to staleness and merge conflicts. Suggestion remains open to consideration.

PR #852, opened 2017-02-08
Make erroneous / dangerous links display as text only

jeremyblalock:
**TL;DR:** Leaves link text but removes hyperlink.

## Explanation

Currently when there is a link that either cannot be parsed, or is detected to be potentially malicious, **marked** simply hides the entire link. For example, if I had the following markdown:

```
For help: [Click Here](data:text/html,lkjasdjflkasd==).
```

the output would be the following:

> For help: .

_This is especially problematic_ when URLs are automatically converted to links. For example:

```
http://example.com
```

Would automatically become:

> http://example.com

If for some reason the URL cannot be parsed, then the URL simply does not display at all. This happens if it has un-urlencoded characters etc., or if it otherwise cannot be parsed, but _was_ identified as a URL by the separate lexer.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -873,10 +873,10 @@ Renderer.prototype.link = function(href, title, text) {
         .replace(/[^\w:]/g, '')
         .toLowerCase();
     } catch (e) {
-      return '';
+      return text;
     }
     if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0 || prot.indexOf('data:') === 0) {
-      return '';
+      return text;
     }
   }
   var out = '<a href="' + href + '"';
--- a/test/tests/gfm_links_invalid.sanitize.html
+++ b/test/tests/gfm_links_invalid.sanitize.html
@@ -1,0 +1,1 @@
+<p>This should not be linked: http://example.com/%ff</p>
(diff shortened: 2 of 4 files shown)


matt-:
I like this idea. It looks like a simple change as well, but do you mind adding a test?

jeremyblalock:
@matt- Yeah, I can add a test for it.

jeremyblalock:
@matt- I updated the invalid link test / added a new one for the `decodeURIComponent` failure. 

On another note, noticed that we are not quite following the recommended test setup as described in the documentation, though. I don't see `links.sanitize.html` in `tests/new`. And so when you run `node test --fix` it will delete this test case, which I'm guessing is not the desired behavior. Are we ditching the `test/original` / `test/new` structure in favor of just putting everything in `/test/tests`? Let me know, I can move the files around if necessary get the `node test --fix` working properly.

jeremyblalock:
@matt- @chjj: any decision on this? This will produce the same behavior that github has in their markdown rendering support for data: urls.

egeste:
+1

Insayt:
<ul><li>123123123</li></ul>

Insayt:
# Hello world

Insayt:
# 

> ``

Feder1co5oave:
This looks good to me. +1

PR #856, opened 2017-02-15
def support single quote

Jiang-Xuan:
I think def regexp should be support single quote.

PS: I am a Chinese boy is not good at English and javascript.Please don't keep in mind if I say something wrong.


joshbruce:
Closing in favor of #1018 - thank you for your contribution.

PR #862, opened 2017-03-03
Not generate html links for url in code block

baonhan:
Auto generating links for url in code block doesn't work properly when the url is followed by special characters. For example:

```
<a href="http://test.com">
```
would generate a link for http://test.com&quote

GFM doesn't seem to require url in code block to be converted to links.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -591,7 +591,7 @@ InlineLexer.prototype.output = function(src) {
     }
 
     // url (gfm)
-    if (!this.inLink && (cap = this.rules.url.exec(src))) {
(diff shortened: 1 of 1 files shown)


styfle:
Hi @baonhan thanks for the PR!

I don't believe this is an issue in the latest version.

[See demo](https://marked.js.org/demo/?text=%60%60%60%0AThis%20is%20a%20%3Ccode%3E%20block%0A%0Ahttp%3A%2F%2Ftest.com%0A%0A%3Ca%20href%3D%22http%3A%2F%2Ftest.com%22%3E%0A%60%60%60) for an example.

PR #878, opened 2017-03-28
Fixes for displaying multiple line breaks

deepthy4poornam:
Hi Team,

I am creating a pull request for the modifications done for the following :

-- to display multiple line breaks
-- to fix display issue with urls having a single %. (Eg : http://api.google.com/q?exp=a%b)

Could you please review it?

Thanks.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -374,6 +374,13 @@ Lexer.prototype.token = function(src, top, bq) {
         href: cap[2],
         title: cap[3]
       };
+ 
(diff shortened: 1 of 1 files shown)


styfle:
Thanks for the PR but this appears to be fixed in the latest version.
[See the demo](https://marked.js.org/demo/?text=to%20fix%20display%20issue%20with%20urls%20having%20a%20single%20%25.%20(Eg%20%3A%20http%3A%2F%2Fapi.google.com%2Fq%3Fexp%3Da%25b)) to see it working.

Issue #252, opened 2013-09-23
compression


adius:
Why is there no option to print out compressed html?
I think that's definitely a "must have" feature!

rudolfschmidt:
If you compress your regular html it would be nice if the html output from markdown would be also compressed. I agree, its a must have feature. Some workaround already?

joshbruce:
Was going to mark this as stale, but I'm curious if this is still a desired feature for folks actively using marked.

I see what @jonathanong is saying. I also think there are issues that come with interspersing Markdown *and* HTML. In some cases, it's better to just pick HTML. Or, compile and inject. With template string literals this could be pretty straightforward. But, I'm new around here still; so, I'll have to use pseudo-code.

If your markdown is coming from a database query, it shouldn't have HTML, in my opinion.

```js
let markdown = 'Hello, World!`;
let html = marked.toHtml(markdown);
let final = `<div style="inline-block">${html}</div>`
```

That HTML is definitely strange but, looking at the broader point, knowing what the output is and the possible side-effects of that...anyway....

See #956

adius:
Still desired.

joshbruce:
The drawback I see is the potential cost-benefit of doing something like this. For `<code><pre></pre></code>` we need to maintain whitespace. I don't know what the overhead would for stripping whitespace, on the fly, while ignoring the triple-backtick case.

adius:
I'd implement the easy solution first. i.e. strip all whitespace between adjacent tags with no content in between. 

```html
<div>
  <p>  Test  </p>
</div>
```
becomes

```html
<div><p>  Test  </p></div>
```
This shouldn't break anything and be easy to implement

UziTech:
This can easily be done with other tools after getting the uncompressed output from marked.

```js
const minify = require('html-minifier').minify;
const marked= require('marked').minify;

const md = "some markdown";

const result = minify(marked(md));

// result is now minified html
```

mccraveiro:
I agree with @UziTech! There are better tools for the job and we should leave this out of marked. 

The "dumb" way of implementing this feature would be to just add `html-minifier` in the bundle. However, this would just add complexity and size to the library. 

I would go with adding @UziTech's example to the readme as an "official" solution for the problem.

UziTech:
I'm going to close this since we will not be adding it to core but will most likely create an extension for it.

Issue #52, opened 2012-05-20
Support contenteditable better with \u00a0 characters


OscarGodson:
As you know from my ticket, contneditables will convert spaces to `\u00a0`. From: http://www.fileformat.info/info/unicode/char/A0/index.htm

It's decomposition is actually a `\u0020`, or, a space. It'd be nice if Marked checked for both of those characters since they mean almost the same thing. I can't imagine someone wanting to keep a no-break space but convert spaces.

Thoughts?

PR #897, opened 2017-05-22
Remove substitution of non-breakable spaces

### LINKED ISSUE #363: Marked removes non-breaking spaces in the original text
Consider, say, a text like this:
`2 125 euro`
I've put a non-breaking space between `2` and `125` so that it would always end up on the same line.

Marked pre-parses the text and completely removes the original non-breaking characters that I've put there:

```
Lexer.prototype.lex = function(src) {
  src = src
    .replace(/\r\n|\r/g, '\n')
    .replace(/\t/g, '    ')
    .replace(/\u00a0/g, ' ')
    .replace(/\u2424/g, '\n');
  return this.token(src, true);
};
```

This is where the devil hides: `.replace(/\u00a0/g, ' ')`

Here is more on why invisible non-breaking space characters are cool: http://destroytoday.com/findings/fix-widows-with-non-breaking-spaces/

### LINKED ISSUE #640: Non-breakable spaces don't work in marked
If I insert a non-breakable space character, it gets to output as a regular space.

This is terribly wrong, we don't want to convert non-breakable space to a regular one.

As a workaround one could use `&nbsp;`, but that looks ugly in Markdown files.

kephas:
Fixes #640 
Fixes #363

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -137,7 +137,6 @@ Lexer.prototype.lex = function(src) {
   src = src
     .replace(/\r\n|\r/g, '\n')
     .replace(/\t/g, '    ')
-    .replace(/\u00a0/g, ' ')
     .replace(/\u2424/g, '\n');
 
   return this.token(src, true);
--- a/test/tests/space.html
+++ b/test/tests/space.html
@@ -1,0 +1,1 @@
+<p>This colon follows un unbreakable space : this should be preserved. There are other spaces, like the en space or the em space. There are also a figure space or a punctuation space.</p>
--- a/test/tests/space.text
+++ b/test/tests/space.text
@@ -1,0 +1,1 @@
+This colon follows un unbreakable space : this should be preserved. There are other spaces, like the en space or the em space. There are also a figure space or a punctuation space.


joshbruce:
@kephas: Is it possible add tests to this PR??

davisjam:
@kephas

I don't know much about this part of the code...

1. Can you comment on the potential impact on the rest of the lexer? I imagine that it will all "just work" and the `\u00a0` will be preserved in the generated HTML to be rendered by the browser?
2. @UziTech If `\u00a0` is a distinct Unicode character with special behavior not equivalent to plain whitespace, why would we eliminate it? We could instead replace `\u00a0` with `&nbsp`, I suppose, but that seems worse than the approach @kephas has proposed.

UziTech:
@davisjam 
It seems this issue happens when taking input from a content editable div. Typing two or more consecutive spaces turns the extra spaces into `\u00a0` which @OscarGodson (#52) believes should be interpreted as whitespace and @kephas (#640) believes should be interpreted as non-breaking space characters.

Martii:
So here's a thought... if someone uses https://github.github.com/gfm/#indented-code-blocks then the "content editable div" will convert those 4 normal typed spaces into 1 normal space and 3 `\u00a0` *(non-breaking spaces)*  and then *marked* currently converts those back to regular spaces *(as if they were typed that way and they were from what I understand reading this)*. Removal of that seems like it would break that part of the spec if they weren't normal spaces. Just something to chew on I hope.

UziTech:
I agree with @Martii, I would think that most people who enter a `\u00a0` mean for it to be a regular space and not a non-breaking space character.

If someone is going to enter `\u00a0` intentionally they can just as easily enter `&nbsp;`

davisjam:
So let's see if I understand the situation: we can't tell with complete certainty whether a `\u00a0` was generated automatically from whitespace (so replacing it with whitespace is fine) or whether it was deliberately entered (so replacing it with whitespace is surprising).

If the majority of the `\u00a0` we see are auto-generated from whitespace, then I think returning them to whitespace by default (i.e. the current behavior) is fine.

I suspect that most of the `\u00a0`'s we see are auto-generated. Markdown is generally produced by humans typing into text editors, and none of the humans I know inject unusual Unicode characters into their markdown files. And @UziTech is correct that an `&nbsp` is a viable way to insert non-breaking spaces if someone really wants them.

kephas:
I fear that you are basing your assumptions on english writers. In French, for example, it is a requirement to have a space before double punctuations like [;:?!]. Some text editors actually do this for you and do this, with good reason, not with a normal space but an unbreakable space, because that punctation should not be orphaned, typographically speaking.

UziTech:
Possible solutions that work for both parties:

1. Leave marked as is and recommend anyone who needs `\u00a0` to be a non-breaking space to use
   ```js
   marked(markdownString.replace(/\u00a0/g, '&nbsp;'))
   ```
2. Commit this PR and recommend anyone who needs `\u00a0` to be whitespace to use
   ```js
   marked(markdownString.replace(/\u00a0/g, ' '))
   ```
3. Create an option to let the user convert `\u00a0` to `' '` or `'&nbsp;'`

Martii:
3)) NON-BREAKING Could work but a default of 1)) :+1: 

1)) NON-BREAKING :question:  Couldn't those few languages that punctuate uniquely write their own `Lexer.prototype.lex` or is that not possible? If it is possible it would make more sense for them to utilize that with the French language.

2)) BREAKING  :-1: 

Misc note: Russian, Spanish, etc. doesn't use an extra space and it's definitely not English. So moot point.

A)) NON-BREAKING  :question:  Seems like CSS and/or the browser default language via  should handle this although I've never seen a rule like this.

B)) NON-BREAKING :question: Some html metadata... Ex. http://nimbupani.com/declaring-languages-in-html-5.html

davisjam:
@kephas You are correct that I was only considering English-language practices in my summary.

@UziTech Thanks for the proposals.

I like the "offer an option for this" approach (3), and I think @Martii's suggestion of using substitution as the default is the right choice.

I don't want to ask users to supply a mix of options and ad hoc substitutions, as suggested in (1).

mccraveiro:
Option 1 means we would never be 100% spec compliant as stated on #1398 
Option 2 makes a lot of sense and goes in the same direction as #252. I mean: the fewer config options we have the better; and everything thats better done outside the marked core shouldn't be included on the core.

Also, we could create a FAQ section on README to include this case and #252.

PR #899, opened 2017-05-24
Add line number support to Lexer.token()

techird:
Hi, I add line number output support to the lexer. That's useful when I want to tell the document author which about warning and suggestions on line level.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -147,7 +147,7 @@ Lexer.prototype.lex = function(src) {
  * Lexing
  */
 
-Lexer.prototype.token = function(src, top, bq) {
+Lexer.prototype.token = function(src, top, bq, ln) {
   var src = src.replace(/^ +$/gm, '')
     , next
     , loose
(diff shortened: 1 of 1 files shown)


lib/marked.js
```diff
@@ -875,7 +912,7 @@ Renderer.prototype.link = function(href, title, text) {
     } catch (e) {
       return '';
     }
-    if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0 || prot.indexOf('data:') === 0) {
+    if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0) {
```
matt-: Why would you remove this XSS protection?

techird: By no means would I remove this. I will revert it.

kristiandupont:
Is there any progress on this? I'd love to use the feature..

joshbruce:
@kristiandupont: Please see #956

I'm not sure how much demand is out there for this. I'm also sure it fits the current, more strict focus as we correct things. 

1. If there are no tests, it won't be merged. 
2. Defects will take precedence over new features.
3. I don't know enough about the codebase to make truly component decisions about certain; therefore, I'm leaning on the community to determine whether marked should even still be a thing.

lib/marked.js
```diff
@@ -159,49 +159,70 @@ Lexer.prototype.token = function(src, top, bq) {
...
+    function() { return 0; };
+
+  var tokens = this.tokens;
+  var push = function(token) {
+    if (lineNumber) {
+      token.ln = ln;
+    }
+    tokens.push(token);
+  };
+
```
fishbar: instead of defining a new push function, override  `this.tokens.push()` seems more graceful:

```
var originPush = this.tokens.push;
this.tokens.push = function (token) {
  if (lineNumber) {
     token.ln = ln;
  }
 originPush.call(this, token);
};
```

otherwise test case is necessary

PR #911, opened 2017-07-10
added option targetBlank

qnp:
added an option `targetBlank` which default is `false` so that the rendered links can be open in a new window. (Marked thus renders `<a>` tags with attribute `target="blank"` )

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -875,13 +875,16 @@ Renderer.prototype.link = function(href, title, text) {
     } catch (e) {
       return '';
     }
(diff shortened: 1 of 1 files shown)


brycedorn:
+1, but I believe it should be `'target="_blank"'` (rather than `'target="blank"'`)

qnp:
Exact ! Fixed. Thanks

hilleer:
Would love to have this feature as well :)

joshbruce:
Closing due to merge conflict and not sure whether either of the two targeted specifications listed in #956 address this.

Issue #956, opened 2017-12-01
Hi, I'm Josh, I'll be your co-pilot for at least a bit


joshbruce:
Hello,

If you've read #951 or #938 or chatter elsewhere, then you know I (and others) have been trying to revive the active status of this project. We're still working out the finer details, but for at least the next 30 days, @chjj and I have agreed to give me the ability to perform merges and publish to NPM on the main project.

Having said that, you all know the codebase, the history, and have more direct knowledge of what's going on than I do. So, I'm asking for your help in this endeavor. 

This goes for anyone using Marked, not just the collaborators.

## Current vision:

Marked is a high-speed, lightweight, and low-level JavaScript library for converting markdown to HTML. 

(Correct me, if this is off-base in some way.)

Note: This is not the same as a full-featured HTML generator. Instead, it is an HTML generator with a specific and limited scope (the markdown specification).

## Current actively supported Markdown specifications

- [CommonMark](http://commonmark.org)
- [GFM](https://github.github.com/gfm/#task-list-items-extension-) (no extensions)

## Current reviving approach

1. If there are no tests, it won't be merged. 
2. Fixing defects will take precedence over new features.
3. I don't know enough about the codebase to make truly competent decisions about certain things; therefore, I'm leaning on the community to determine whether marked should even still be a thing.

## Roadmap

We believe all XSS vulnerabilities have been addressed and released. Therefore, we have established 3 milestones based on tickets identified. They are as follows:

1. No known issues (https://github.com/chjj/marked/milestone/1): This milestone is about focusing on issues flagged as `broken`, `annoying`, and `refactoring & re-engineering`. The goal is to address all of these issues - most likely resulting in a 0.4.0 release.
2. Architecture and extensibility (https://github.com/chjj/marked/milestone/2): This milestone focuses on some community concerns and comments related to how the project itself is setup. Once these issues and proposals have been considered there will most likely be a 0.5.0 release.
3. Improve developer experience (https://github.com/chjj/marked/milestone/3): Issues related to this milestone are about improving the consumability and contributability of Marked and will most likely result in a 0.6.0 release.

Help us out by linking duplicate issues and PRs...going through one issue and PR at a time is pretty time consuming. :)

joshbruce:
Added option to preserve list numbering from input

rgbkrk:
Thanks for taking up the mantle and being so thorough @joshbruce!

joshbruce:
Hello,

The position from which this is written:

[...]

PR #961, opened 2017-12-01
Release 0.3.7

### LINKED ISSUE #925: Cross-site Scripting (XSS) via URI
Browsers support both lowercase and uppercase `x` in hexadecimal form of HTML character entity (tested on Chromium && FF).
But marked unescape only lowercase:
```js
// https://github.com/chjj/marked/blob/master/lib/marked.js#L1096-L1108

function unescape(html) {
	// explicitly match decimal, hex, and named HTML entities 
  return html.replace(/&(#(?:\d+)|(?:#x[0-9A-Fa-f]+)|(?:\w+));?/g, function(_, n) {
    n = n.toLowerCase();
    if (n === 'colon') return ':';
    if (n.charAt(0) === '#') {
      return n.charAt(1) === 'x'
        ? String.fromCharCode(parseInt(n.substring(2), 16))
        : String.fromCharCode(+n.substring(1));
    }
    return '';
  });
}
```

This allow attacker to create link with javascript code.
For example, this code:
```js
var marked = require('marked');
marked.setOptions({
  renderer: new marked.Renderer(),
  sanitize: true
});

text = `
lower[click me](javascript&#x3a;...)lower
upper[click me](javascript&#X3a;...)upper
`;

console.log(marked(text));
```

Will render:
```html
<p>lowerlower
upper<a href="javascript&#X3a;...">click me</a>upper</p>
```

Browser example: https://www.buglloc.com/marked.html
Tested on Marked v0.3.6 + Chromium 60.0.3112.90 and Firefox 55.0.1

### LINKED ISSUE #926: Cross-site Scripting (XSS) via autolink (require disabled mangling)
When mangling is disabled via option `mangle` marked don't escape target href. This allow attacker to inject arbitrary html-event into resulting `a` tag.

For example, this JS code:
```js
var marked = require('marked');
marked.setOptions({
  renderer: new marked.Renderer(),
  sanitize: true,
  mangle: false
});

text = `
<bar"onclick="alert('XSS')"@foo>
`;

console.log(marked(text));
```

Will render:
```html
<p><a href="mailto:bar"onclick="alert('XSS')"@foo">bar"onclick="alert('XSS')"@foo</a></p>
```

Tested on marked v0.3.6
Online demo: http://www.buglloc.com/marked-mangle.html

Fixes #925

### LINKED ISSUE #937: Vulnerable Regular Expression
The following regular expression used in parsing the input markdown content is vulnerable to ReDoS:
```js
/^(`+)\s*([\s\S]*?[^`])\s*\1(?!`)/
```
The slowdown is very serious (for 1,000 characters around 6 seconds matching time). I would suggest one of the following: 
- remove the regex,
- anchor the regex,
- limit the number of characters that can be matched by the repetition,
- limit the input size.

If needed, I can provide an actual example showing the slowdown.

joshbruce:
Fixes #937 
Fixes #925 
Fixes #926


joshbruce:
#956

PR #967, opened 2017-12-05
update travis ci

UziTech:
@joshbruce I'm not sure if you are able to enable travis-ci on this repo or if @chjj would need to do that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
 node_js:
   - "4"
-  - "7"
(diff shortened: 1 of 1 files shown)


joshbruce:
@UziTech: I do not have access to the "Settings" tab; therefore, cannot.

Further, I believe Travis/Circle/etc., have a fee eventually.

I think having an integrated check would definitely be beneficial.

Tagging and assigning to @chjj

joshbruce:
#956

PR #968, opened 2017-12-05
#269

jguffey:
WRT escaping quotes that appear in HTML

Implements solution suggested in https://github.com/chjj/marked/issues/269, also breaks tests.


joshbruce:
@jguffey: Thanks for the PR!

Could you do me a favor and write unit tests demonstrating this in use? I'm not familiar with the library enough to understand what would be the optimum solution, the test will help me, and others, understand what's going on.

Thanks!

joshbruce:
Closing as merge conflicts, missing tests, and no response in the last 30 days.

PR #974, opened 2017-12-18
Fixed https://github.com/chjj/marked/issues/465

KostyaTretyak:


joshbruce:
@matt- or @UziTech - Can you check this out.

KostyaTretyak:
I split my commit into two commits. To make it clearer what I am fixing, see https://github.com/chjj/marked/pull/974/commits/98ac7a43958014832788b9d01580f394c84bfb27

UziTech:
LGTM

The bug that code was meant to fix (https://github.com/chjj/marked/issues/77) seems to still work

KostyaTretyak:
@UziTech, when I do the following (with this my PR):

```js
const marked = require('./');

const str = `[link][1]

* List

  [1]: http://example.com
`;

const md = marked(str);
console.log(md);
```

marked output this:

```html
<p><a href="http://example.com">link</a></p>
<ul>
<li>List</li>
</ul>
```

What is exactly awaits @tlvince.

KostyaTretyak:
By the way, and without my PR, it works as expected.

lib/marked.js
```diff
@@ -39,10 +39,6 @@ block.list = replace(block.list)
   ('def', '\\n+(?=' + block.def.source + ')')
   ();
 
-block.blockquote = replace(block.blockquote)
```
Feder1co5oave: this is plain wrong, cannot remove this without changing the block.blockquote regex

test/tests/def_blocks.html
```diff
@@ -6,7 +6,8 @@
 <hr>
 
 <blockquote>
-  <p>hello</p>
+  <p>hello
+[2]: hello</p>
```
Feder1co5oave: from def_blocks.text:

   > hello
   [2]: hello

I'm not sure this should/should not be parsed as a link definition. It is _outside_ a blockquote.

test/tests/def_blocks.html
```diff
@@ -24,5 +25,6 @@
 <blockquote>
   <p>foo
 bar
+[1]: foo
```
Feder1co5oave: There's a problem in https://github.com/KostyaTretyak/marked/blob/98ac7a43958014832788b9d01580f394c84bfb27/test/tests/def_blocks.text :
The `[1]` at line 2 is interpreted as a shortcut link to link definition at line 20, so it can be confusing. However, I think it (line 2) should be parsed as a link def itself, because it follows the correct syntax, event if it is inside a blockquote, because it is allowed by the [commonmark spec](http://spec.commonmark.org/0.28/#example-181).

KostyaTretyak: As we now know, this edit was correct because we have rule [A link reference definition cannot interrupt a paragraph.](http://spec.commonmark.org/0.28/#example-178).

test/tests/def_blocks.html
KostyaTretyak: As we now know, this edit was correct because we have rule [A link reference definition cannot interrupt a paragraph.](http://spec.commonmark.org/0.28/#example-178).

Feder1co5oave: http://spec.commonmark.org/0.28/#example-210
Ok, because of the laziness rule, '[2]: hello' is part of the blockquote, too, and of the same paragraph of the previous line. So it is not a link def. This is correct then!

test/tests/def_blocks.html
Feder1co5oave: There is no broad consensus on this one:
http://johnmacfarlane.net/babelmark2/?normalize=1&text=%3E+foo%0A%3E+bar%0A%5B1%5D%3A+foo%0A%3E+bar

However, commonmark.js parses it as you wrote.

Feder1co5oave:
I have worked on fixing rules for link defs, they should pass the def_blocks test as written by @KostyaTretyak (which we've deemed correct per commonmark), and also almost every example from commonmark. I will submit a pr once I've reviewed and rearranged commits.

PR #987, opened 2017-12-26
add a test case for links defs & paragraphs

gromnitsky:
this is how it should work according to CommonMark

### DIFF
--- a/test/tests/links_reference_style.html
+++ b/test/tests/links_reference_style.html
@@ -43,6 +43,16 @@
 
 <p>Backslashing should suppress [this] and [this].</p>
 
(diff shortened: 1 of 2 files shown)


Feder1co5oave:
+1 merge this

joshbruce:
@Feder1co5oave: Not sure when I will be able to confirm - does this test pass or fail? In other words, is it just testing to block a possible future regression; or, is it confirming that a specification is not being following without actually solving the issue?

gromnitsky:
@joshbruce the test does pass & it's just testing to block a possible future regression, for I accidentally found out that the prev rel (0.3.7) exhibits a diff behavior that doesn't conform to the CommonMark spec.

PR #991, opened 2017-12-27
Declare undeclared variables

learykara:
Fix for https://github.com/chjj/marked/issues/990

These two variables are undeclared and throw an error in strict mode.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -5,6 +5,7 @@
  */
 
 ;(function() {
+'use strict';
 
 /**
  * Block-Level Grammar
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1144,8 +1144,8 @@ function resolveUrl(base, href) {
     return base + href;
   }
 }
-baseUrls = {};
-originIndependentUrl = /^$|^[a-z][a-z0-9+.-]*:|^[?#]/i;
+var baseUrls = {};
+var originIndependentUrl = /^$|^[a-z][a-z0-9+.-]*:|^[?#]/i;
 
 function noop() {}
 noop.exec = noop;


joshbruce:
@learykara: Thank you for the PR. Is there a way to write a test for this? 

Want to make sure we don't get a regression.

I don't know enough about the Marked test suite to know and I think it's somewhat custom.

Feder1co5oave:
Adding `"use strict";` at the top of lib/marked.js puts the engine in strict mode and doesn't compile without this fix. I don't know the consequences of doing this, though. Might need to hear someone about this.

joshbruce:
Let's tag in @matt- and @UziTech, see if they can weigh in.

felixfbecker:
I agree that adding `'use strict'` would be best (and if tests pass that should mean everything is alright) but it would be great to get this fix released asap. The last working version 0.3.6 has a security vulnerability that is currently alerted by GitHub with the recommendation to update to 0.3.9, which breaks everyone in a strict mode environment at runtime.

felixfbecker:
This can also be prevented with eslint (no-unused-variable)

UziTech:
LGTM @learykara could you add `"use strict";` at the top of the function (line 8)

learykara:
done @UziTech !

joshbruce:
All right. Lots of good conversation here.

Here's the plan:

1. Merging this.
2. Prepping a 0.3.10 release - please flag other PRs you think fix defects that should be included in that release within the next two days. So, plan is to have a release on Monday.
3. Starting a ticket to discuss adding a linter to Marked. See #999

Anything else we need to do?

PR #993, opened 2018-01-01
Support older regex engines

smhg:
Recent changes broke IE8 support. This makes things compatible with old regex engines (in IE8, old Safari and old Opera).
More information: https://stackoverflow.com/questions/10843428/what-is-meaning-of-in-javascript-regexps/10843677#10843677


Feder1co5oave:
+1

joshbruce:
LGTM! No way to truly test I don't think - unless we added some type of browser compatibility testing harness, yeah?

Issue #963, opened 2017-12-03
[Vision alignment]: Comparative benchmarking - high-speed


joshbruce:
What other markdown compilers are there for JS now?

Would be worthwhile to put together a list of compilers (markdown-it, for example) and a complex (but sane) markdown file to do comparative benchmarking. 

If we aren't faster than them, then we aren't achieving the primary purpose of marked...high-speed.

See also #956

worker8:
I found this site jsperf.com for comparing javascript library performance.

Someone wrote a test for comparing the performance for a list of markdown compilers:

https://jsperf.com/showdown-js-vs-marked/10

![screen shot 2018-01-05 at 12 39 29 pm](https://user-images.githubusercontent.com/1988156/34594626-8ddf4b74-f215-11e7-9257-7bedf2245c56.png)

According to the tests, there are a few libraries ranked higher than marked.

Feder1co5oave:
newer version:
https://jsperf.com/showdown-js-vs-marked/19
also:
https://github.com/commonmark/commonmark.js#performance

calculuschild:
First, is there a good online benchmark anymore? The jsperf links above seem to be broken.

Running the Marked bench script on my own machine is showing Commonmark and Marked-It as both faster and more compliant. Can you point to any techniques that used to give you the edge that aren't working anymore, or recent updates that introduced a lot of slowdown? I'd be interested in taking a look.

UziTech:
I don't know of any online benchmarks.

I think #1686 made a bit of a slow down but also markdown-it has been getting faster with recent releases. I don't know of any simple ways to increase the speed.

calculuschild:
> I think #1686 made a bit of a slow down

I'm sure it did, there was a lot going on in there. I wonder if... there's a way to show kind of a benchmark score per release? Run some kind of automated benchmark on all releases up to this point, and collect that info somewhere useful so it could be plotted and compared, maybe as a page on the documentation?

UziTech:
We tried to integrate the benchmarks into our testing but test servers are extremely inconsistent

PR #998, opened 2018-01-05
get rid of warning on 'npm run bench'

Feder1co5oave:
This fixes https://github.com/chjj/marked/pull/977#issuecomment-355463708

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -300,6 +300,8 @@ function runBench(options) {
   } catch (e) {
     console.log('Could not bench markdown.js.');
   }
+
+  return true;
 }
 
 /**


UziTech:
Do we want to change it to return false if we couldn't bench something?

Feder1co5oave:
As in?

UziTech:
like if showdown or robotskirt fails? (like they are now)

or maybe set a threshold and if the marked bench is higher than that it should fail? that way we could actually use it in our automated testing and flag if a pr makes it too much slower..

Feder1co5oave:
Benching against other parsers is optional so I wouldn't exit 1.
About the threshold... I don't really like it.

UziTech:
It seems like there is not really any point in benching then.

joshbruce:
From the product/marketing perspective: We claim to be built for speed, which raises the question, compared to what? I think Apple does this pretty well. When they release a new version of Safari, for example, they do comparitive analysis against the previous version and the other major browsers. 

@worker8 posted an interesting link over on #963 that does some comparitive analysis.

If we don't bench, can we really make the claim honestly?

From the developer perspective I think what I would like to see as a consumer of Marked would be the bench results for the supported parsers - and if one fails or times out a message saying something like: This parser failed. If you want to use this parser, please submit a GitHub issue. Maybe include the URL to the issues page.

Re threshold: The issue there is device dependence, I think. @Feder1co5oave ran his benches and they were all 10 seconds longer than mine. I can't think of an elegant way to deal with that to choose a quality threshold. For 8fold Component - at least when I was still doing benchmarking - I used 3ms...no matter the scenario under test. I do appreciate the idea of a threshold because, like @UziTech said, without a target, why bench...just not sure if that is a separate issue compared to the benches throwing the error or not.

My throwing copper moment on this for now.

PR #1002, opened 2018-01-06
Fix tests

UziTech:
Changes to tests

- Markdown for tests are written in `.md` files instead of `.text`
- No need to `--fix` the tests as it is done automatically before you run the tests
- Added a `--no-fix` arg for the tests in case you don't want it to automatically fix the tests before testing
- tests that need specific options set can use [front-matter](https://www.npmjs.com/package/front-matter) at the top of the `.md` file to specify the options instead of the file name
- fixed benchmarks
  - removed robotskirt
  - fixed showdown
  - added markdown-it
- updated readme to include new benchmarks and testing description

fixes #998


joshbruce:
Wow. @UziTech, this is very impressive. Thank you so much.

Given the spread of changes I would like a second pair of eyes and some questions confirmed. Maybe @Feder1co5oave, @matt-, or @KostyaTretyak can have a look.

Questions/confirmation:

- [ ] Only affects test harness and patterns (doesn't add the property setting feature requested elsewhere).
- [ ] Most, if not all of the files changed or deleted are to convert them from text or HTML files to MD.
- [ ] The Marked lib itself was not touched; therefore, all previous tests are in tact, benchmarks remain the same, and no new functionality was added or removed from the main lib.

Again, thank you so much, this should make it much easier to do comparitive analysis and support different specs in the future.

test/index.js
```diff
@@ -320,31 +324,32 @@ function time(options) {
  */
 
 function fix() {
-  ['tests', 'original', 'new'].forEach(function(dir) {
+  ['compiled_tests', 'original', 'new'].forEach(function(dir) {
     try {
-      fs.mkdirSync(path.resolve(__dirname, dir), 0755);
+      fs.mkdirSync(path.resolve(__dirname, dir), 0o755);
```
KostyaTretyak: `0o755` - it's not typo?

test/index.js
```diff
@@ -261,45 +253,52 @@ function runBench(options) {
...
+  try {
+    bench('markdown-it', (function() {
+      var MarkdownIt = require('markdown-it');
+      var md = new MarkdownIt();
+      return function(text) {
+        return md.render(text);
+      };
+    })());
+  } catch (e) {
+    console.log('Could not bench markdown-it.');
```
KostyaTretyak: Here we can show error like this: `console.log('Could not bench markdown-it. (Error: %s)', e);`

test/index.js
UziTech: That is an [octal literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Integers)

UziTech:
@joshbruce 
- Yes this only changes the test harness. `lib/marked.js` was not touched at all
- Most changed files were from removing the `test/tests` folder and changing all `test/*/*.text` files to `test/*/*.md` as well as setting test options with front-matter
- All test remain the same (only the way options are specified was changed)
- No new tests were added as functionality did not change

test/index.js
UziTech: I was thinking of doing something like that.

Would we want to also check for `e.code === 'MODULE_NOT_FOUND'` and display a message like `Please install x to run bench`?

KostyaTretyak: I think it's better to make it simpler - just error message.

UziTech: done

joshbruce:
All right. LGTM! Easy to change in future without affecting our users...rollback or something else. And confirmed.

Again, awesome work. Thank you.

test/index.js
joshbruce: @UziTech: Maybe create an issue for the idea of "module not found" - I like the idea of decoupling us from packages and giving users a message like that. I'd flag it for 0.6.0 milestone as it improves the developer experience.

UziTech: The way I set it up now if they don't have a module installed they will see:

```
Could not bench showdown. (Error: Cannot find module 'showdown')
```

joshbruce:
#746

Feder1co5oave:
I recently peeked into the test script and I must say it looks _hacky as hell_. The `fix()` function especially, since its job is to make the `original` tests from `markdown.pl` work on marked. Should we consider updating these test files to work with marked out of the box? I'm actually trying to incorporate test cases from commonmark to improve out test baseline and I'm facing similar difficulties.

For instance, one of the features that marked has is automatic id generation for headings, which look quite cumbersome (and painful to hand-write) in test cases. Should we consider disabling this in testing?

KostyaTretyak:
> For instance, one of the features that marked has is automatic id generation for headings

@Feder1co5oave, do you mean this method?

```js
Renderer.prototype.heading = function(text, level, raw) {
  return '<h'
    + level
    + ' id="'
    + this.options.headerPrefix
    + raw.toLowerCase().replace(/[^\w]+/g, '-')
    + '">'
    + text
    + '</h'
    + level
    + '>\n';
};
```

It would be better?

```js
Renderer.prototype.heading = function(text, level, raw) {
    var id = this.options.headerPrefix + raw.toLowerCase().replace(/[^\w]+/g, '-');
    return `<h${level} id="${id}">${text}</h${level}>\n`;
};
```

KostyaTretyak:
As for me, `marked` is rather uncomfortable when it removes all the spaces for testing.

Feder1co5oave:
No, I meant `` `<h${level}>${text}</h${level}>` ``, without any id at all.

Spaces are too tricky to make sure they match perfectly.

KostyaTretyak:
I have rewritten this functionality in myself:

```text
#9. failed def_blocks.md

Expected:
~~~~~~~~> in /var/www/marked-ts/test/tests/def_blocks.html:25:4

'<p><a href="http://spec.commonmar'


Actual:
--------> in /var/www/marked-ts/test/tests/def_blocks-actual.html:25:4

'<p>[spec]</p>'


Excerpt tokens: [ { line: 24, type: 'blockquoteEnd' },
  { line: 25, type: 'paragraph', text: '[spec]' },
  { type: 'blockquoteStart' },
  { type: 'text',
    text: '[spec]: http://spec.commonmark.org/0.28/#example-181' },
  { line: 28, type: 'blockquoteEnd' } ]
links: { '4': { href: 'hello', title: undefined } }
```

joshbruce:
Throwing copper...

I believe we're using a sort of homegrown solution for testing, yeah?

I don't see JSUnit, for example, as a dev dependency. Writing code to solve our specific problem is great, but might be nice for other developers to not have to learn some new way to write tests as well.

Feder1co5oave:
I don't think unit testing is suitable for this. The current testing concept (input-vs-expected-output) is the right one, it just needs to be more homogeneous.

UziTech:
I thought about moving to a modern testing framework (jasmine, mocha, etc.) when rewriting the tests but I agree with @Feder1co5oave. The input output approach is much easier than writing unit tests with fixtures.

We could add some sort of unit tests on top of the current test to provide more control. At the moment there is no way to run a test that changes the options in between tests without writing multiple tests.

joshbruce:
Just to deal with the language thing to make sure I'm on the same page with y'all. I tend to lump tests into three groups: unit, integration, and application.

Unit: Testable in isolation of a single function or class to confirm input-output.
Integration: Requires other objects to confirm input-output.
Application: Tests how a user actually interacts with a system.

In my brain, we'd being doing more of an integration test - I'm with y'all - not concerned the parts in isolation (sorry if the JSUnit mention got us on the "unit" testing thing). Want to verify we get the desired output based on a specific input. Doesn't mean we can't use a more standard testing library see 8fold Component: https://github.com/8fold/php-html-component/blob/master/tests/ComponentTest.php

The only issue from a developer experience perspective, would be in the markdown versus HTML files. But, I defer to @UziTech more than anyone, because he seems the most familiar and I don't have the bandwidth to dive too deep on things. (Thanks again, y'all have been a pleasure to work with.)

PR #1004, opened 2018-01-07
sort tests

UziTech:
Sorting tests alphabetically for more deterministic tests.

@joshbruce this was supposed to go with #1002

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -598,6 +598,12 @@
         }
       }
     },
+    "glob-to-regexp": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/glob-to-regexp/-/glob-to-regexp-0.3.0.tgz",
+      "integrity": "sha1-jFoUlNIGbFcMw7/kSWF1rMTVAqs=",
+      "dev": true
+    },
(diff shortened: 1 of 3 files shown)


joshbruce:
@UziTech: Interesting. This is alphabetizing the console output, yeah? Just to make sure I'm understanding...then I should probably go to sleep.

UziTech:
And alphabetizing the way the tests are ran. So we can run tests in a specific order if we need to.

joshbruce:
Cool. Do we have a way to run a single test case? Just curious.

UziTech:
No we don't. Do you want me to set that up?

joshbruce:
@UziTech: If it's not difficult, it would probably be helpful - not that it takes that long for our tests to run, of course...we just have a lot and might be easier for someone adding a new feature to test their single piece of new functionality, you know?

UziTech:
I added running only certain tests by specifying a glob pattern

``` bash
node test --glob gfm*
```

that will only run tests that start with 'gfm'

joshbruce:
Thanks, brother!

joshbruce:
#746

PR #1005, opened 2018-01-07
Add message: token type cannot be found

KostyaTretyak:
It seems to me that such a message will be appropriate. If I did not write very English, the wishes for corrections are accepted.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1082,6 +1082,14 @@ Parser.prototype.tok = function() {
     case 'text': {
       return this.renderer.paragraph(this.parseText());
     }
+    default: {
+      var errMsg = 'Token with "' + this.token.type + '" type was not found.';
+      if (this.options.silent) {
+        console.log(errMsg);
+      } else {
+        throw new Error(errMsg);
+      }
+    }
   }
 };
 


joshbruce:
@KostyaTretyak: Can you give me an example of a type? I don't know enough about marked to knwo what we mean by "token" or "type".

KostyaTretyak:
Forgive me, but you should not maintain this project if you do not know what a token is and what a type of token is. By the way, it is elementary looking in the project.

joshbruce:
@KostyaTretyak: Forgiven. I will ask you forgive me as well, if this seems harsh or offbase.

I help maintain enough projects to know that they all use language differently. What makes language difficult is that we have to encode and decode messages, which comes from understanding the language. I'm only asking for courtesy and assistance from who knows the language of the community and the library because I don't know; in the time it took to write your reply - you probably could have answered the question (thereby, helping me become smarter). 

To put it bluntly, I don't know if I can help with the English question of this PR because I'm missing the possible word replacement possibilities in the submission...but, I would like to help, as I'm pretty good with language and user experience.

Also, to be fair, I don't "maintain" the project - I coordinate and publish the exceptional efforts of the maintainers such as yourself, @UziTech, @Feder1co5oave, and so on. If you believe I am doing a poor job or am underqualified, please let me know and I will begin the stepping away protocol eluded to in #956.

Again, just asking for help in increasing my knowledge and understanding by interacting with someone who knows - I'm more of a people person in that regard. 

If there was a test demonstrating the expectation and result, the answer would be obvious without me (or another developer) needing to know how marked works (to be fair, I don't know if writing a test is possible for this PR). That's what our tests offer our users - documentation on usage and expectation without needing a high-level of knowledge about the marked lib.

Feder1co5oave:
This seems pointless at the current situation, but might be useful once we make marked a bit more easily extensible

joshbruce:
@Feder1co5oave: Agreed. Flagged for 0.6.0.

lib/marked.js
```diff
@@ -1082,6 +1082,14 @@ Parser.prototype.tok = function() {
     case 'text': {
       return this.renderer.paragraph(this.parseText());
     }
+    default: {
+      var errMsg = 'Token with "' + this.token.type + '" type was not found.';
```
styfle: I think `this.token.type` can point to something different that what was used in the `switch` statement. Perhaps you should store it as a variable before calling `switch` to use in this message.

Also, what markdown input would cause `default` to be executed?

styfle: Oh second thought, this looks correct.

Issue #1006, opened 2018-01-08
Any `def` string in code block prevents block detection


rndstr:
When there is a `def` at the beginning of any line within a code block, it is not recognized as such.

Sample inputs

A
```
`
def`
```

B

    ```ruby
    def foo(x)
      return 3
    end
    ```
## Expectation

The code block should be recognized as code block.

## Result

The code block is not recognized and just rendered as paragraphs.

A
```
<p>`</p>
    <p>def
    `</p>
```
B
```
<p>`
    ruby</p>
    <p>def foo(x)
      return 3
    end
    `</p>
```

## More information
This broke by commit https://github.com/chjj/marked/pull/976/commits/98ac7a43958014832788b9d01580f394c84bfb27
which seemed to have originated in https://github.com/chjj/marked/pull/974
and was supposed to fix https://github.com/chjj/marked/issues/465

Feder1co5oave:
I think we need to revert and review #974 as soon as possible. I don't recall how that could have slipped my checking.
This can be quite tricky since we changed the whole test system, so expect to manually resolve some conflicts.

KostyaTretyak:
Agree, should do revert  https://github.com/chjj/marked/commit/98ac7a43958014832788b9d01580f394c84bfb27 and https://github.com/chjj/marked/pull/974/commits/a477d1d0e285e1b0aed91ea9048bc66e85324b2a  commits. They are incorrect.

joshbruce:
@Feder1co5oave and @KostyaTretyak: Agreed.

I know of three ways to do that - and seem to only be able to do one of them myself. (Note: Beleive it or not, I've never had to a reversion.)

1. Copy-paste and submit a new PR...not revert.
2. https://help.github.com/articles/reverting-a-pull-request/ - in theory, there should be a button at the bottom of the PRs to allow reversion - there is not.
3. Using git via terminal - https://git-scm.com/docs/git-revert.html - not comfortable enough to do that one myself for a live project in production.

KostyaTretyak:
@joshbruce, please do the following via terminal from root the project:

```bash
git checkout master
git revert a477d1d --no-edit
git push origin master
```

It should go through without problems. And now I will pull request for revert `98ac7a4`.

Feder1co5oave:
@KostyaTretyak can you not open a PR to revert both commits? That would be neater.

KostyaTretyak:
[...]

PR #1007, opened 2018-01-08
Revert 98ac7a4 and a477d1d

KostyaTretyak:
See https://github.com/chjj/marked/issues/1006

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -40,6 +40,10 @@ block.list = replace(block.list)
   ('def', '\\n+(?=' + block.def.source + ')')
   ();
 
+block.blockquote = replace(block.blockquote)
+  ('def', block.def)
+  ();
+
 block._tag = '(?!(?:'
   + 'a|em|strong|small|s|cite|q|dfn|abbr|data|time|code'
   + '|var|samp|kbd|sub|sup|i|b|u|mark|ruby|rt|rp|bdi|bdo'
--- a/test/new/def_blocks.html
+++ b/test/new/def_blocks.html
@@ -6,8 +6,7 @@
 <hr>
 
 <blockquote>
-  <p>hello
-[2]: hello</p>
+  <p>hello</p>
 </blockquote>
 
 
@@ -25,6 +24,5 @@
 <blockquote>
   <p>foo
 bar
-[1]: foo
 bar</p>
 </blockquote>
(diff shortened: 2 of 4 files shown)


joshbruce:
@Feder1co5oave: Real quick, this looks good to me, but I would like a second pair of eyes - just to be able to add it to the release we are trying to do today.

Thank you @KostyaTretyak!

Feder1co5oave:
NO WAIT

Feder1co5oave:
Looks like this won't solve #1006. Check out my last comments.

KostyaTretyak:
Now marked works this way:

```md
`
def ?`

```

Output:

```html
<p>`</p>
<p>def ?`</p>
```

With this pull request marked output:

```html
<p><code>def ?</code></p>
```

KostyaTretyak:
However, when we return everything as it was before https://github.com/chjj/marked/commit/98ac7a43958014832788b9d01580f394c84bfb27, then [this rule](http://spec.commonmark.org/0.28/#example-178) will not be works.

KostyaTretyak:
However, in two of your three comments, you were wrong with the rules. Now we know that my test edits were correct, but I was incorrect when I deleted rule with blockquote from `marked.js`.

joshbruce:
Two questions mainly for @KostyaTretyak, @Feder1co5oave, @UziTech:

1. Do we merge this before doing the release of [0.3.12](https://github.com/chjj/marked/releases/tag/untagged-fd441601fc06eba749f4)? 
2. Should we push the release off a day or two to figure this out?

Feder1co5oave:
I believe it's better if we just revert this one, push out important security issues and in the meanwhile work on these, as you suggest @joshbruce.

@KostyaTretyak I admit I'm crossing my eyes over regexes and markdown examples, can you point out in the review what complies to the spec and what doesn't?

Thanks to both of you

joshbruce:
@Feder1co5oave: I believe that's the best as well. Having said that, still not entirely comfortable doing the revert myself using `git revert` if this one doesn't do that. Help a brother out?

Feder1co5oave:
You only need to merge this pr

joshbruce:
@Feder1co5oave: Thanks, brother. Just really wanted to make sure I was looking at the right thing and understanding the conversation that was happening. :)

joshbruce:
Text-base communication because the lowest bandwidth communication and all that. ;)

PR #503, opened 2014-10-18
fix backticks in inline code (fixes #312)

Feder1co5oave:
Fixes #312.
Some simple test cases are included in linked issue.

PR #1013, opened 2018-01-14
fix backticks in inline code (replaces #503, fixes #312)

### LINKED ISSUE #312: Single backtick in inline code
``` sh
$ marked
`` ` ``
^D
<p><code>` </code></p>
```

There is a remaining white space after the backtick that shouldn't be there !

http://daringfireball.net/projects/markdown/syntax#code
At the "A single backtick in a code span" part, the output doesn't have that trailing space.

Feder1co5oave:
Fixes #312:

~~~html
$ marked
`` ` ``
^D
<p><code>` </code></p>
~~~

Tests are about checking for whitespace, which is ignored by our test system for now.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -458,7 +458,7 @@ var inline = {
   nolink: /^!?\[((?:\[[^\]]*\]|[^\[\]])*)\]/,
   strong: /^__([\s\S]+?)__(?!_)|^\*\*([\s\S]+?)\*\*(?!\*)/,
   em: /^\b_((?:[^_]|__)+?)_\b|^\*((?:\*\*|[\s\S])+?)\*(?!\*)/,
-  code: /^(`+)([\s\S]*?[^`])\1(?!`)/,
+  code: /^(`+)(\s*)([\s\S]*?[^`]?)\2\1(?!`)/,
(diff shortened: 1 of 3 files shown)


lib/marked.js
```diff
@@ -664,7 +664,7 @@ InlineLexer.prototype.output = function(src) {
     // code
     if (cap = this.rules.code.exec(src)) {
       src = src.substring(cap[0].length);
-      out += this.renderer.codespan(escape(cap[2].trim(), true));
+      out += this.renderer.codespan(escape(cap[3].trim(), true));
```
UziTech: @Feder1co5oave @joshbruce We have to watch out for changing capturing parentheses. This can be a breaking change. https://github.com/firede/msee/pull/59/files

Feder1co5oave:
Ah, I see, they hook between the lexer and the parser and they process
tokens straight out of it. To be honest this change required a new
capturing group so I was forced to shift the others

Il lun 19 feb 2018, 01:08 Tony Brix <notifications@github.com> ha scritto:

Feder1co5oave:
They actually re-wrote the parser, while using its internal regex rules, so I maintain this should not be considered a breaking change as suggested in #1059

joshbruce:
#1077

Issue #907, opened 2017-06-17
Setting options interferes with other "require"-calls to marked.


nknapp:
In my NodeJS-project, `marked` is used by the main program and by a dependency.
When the main-program calls `setOptions`, the same options are also applied to the instance that is used by the dependency, because both instances are the same object.

I think, a better way to apply default options, would be like the `request`-package is doing it.
They create a new instance that is configured with the default options, so the original instance is untouched.

What I will try now, is to clear the cache before and after requiring `marked`, but that seems like a weak workaround.

Feder1co5oave:
idk if this is viable for you, but you could avoid setting options with `setOptions` and instead pass options to `marked` as second argument.

I do agree with you that setOptions should return a new marked instance instead! But that is an API change.

joshbruce:
Again, I'm not too familiar with working directly with Marked; so, pardon the ignorance. Also, this sounds like the problem with Singletons (especially if they're configurable).

```javascript
var marked1 = Marked().setOptions();
var marked2 = Marked().setOptions();
```

Is that not the way things work? Sounds like each package would want to create the instance, at which point you would access via something like:

```javascript

var package = Package();

package.marked.setOptions();
```

In either case I'm not sure how much this is on Marked to correct. Will flag it for the 0.6.0 milestone; closing.

UziTech:
A work around is to use

```js
const marked = require('marked');
marked.setOptions(marked.getDefaults());
```

to make sure that defaults are set to the initial defaults.

Andarist:
The proposed workaround can break those other marked "instances" which is also not good overall.

UziTech:
True. You could also send options as a second parameter to marked to override the defaults.

```js
const marked = require('marked');
marked(markdown, marked.getDefaults());
```

john-doherty:
I ran into this. As @Feder1co5oave said, the solution is to avoid using `marked.setOptions({ renderer: renderer })` and instead pass options inline `marked(md, { renderer: renderer })` - worked for me.

DLiblik:
I hate to rain on a parade, but this is a really clumsy issue.  We did not realize this was happening (singleton instance in the background) and to make matters worse, our test pipeline turns over our server as it runs test cases.  We had 3 extensions being added when the module was imported (which occurs on each test-reset) and found we had 1,926 extensions installed because marked has no way to reset existing options and it runs as a singleton.

[...]

PR #1015, opened 2018-01-15
fill unspecified options with defaults for lexers and parser

solugebefola:
This ensures that if a custom Parser or Lexer is made with an options object as an argument, any unspecified options will be set by `marked.defaults` rather than being undefined.

One note:  I ran the test suite before making changes, and there were two failures

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -103,7 +103,7 @@ block.tables = merge({}, block.gfm, {
 function Lexer(options) {
   this.tokens = [];
   this.tokens.links = {};
-  this.options = options || marked.defaults;
+  this.options = merge({}, marked.defaults, options || {});
   this.rules = block.normal;
 
   if (this.options.gfm) {
@@ -519,7 +519,7 @@ inline.breaks = merge({}, inline.gfm, {
  */
 
 function InlineLexer(links, options) {
-  this.options = options || marked.defaults;
+  this.options = merge({}, marked.defaults, options || {});
   this.links = links;
   this.rules = inline.normal;
   this.renderer = this.options.renderer || new Renderer;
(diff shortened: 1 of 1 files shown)


KostyaTretyak:
You can use `marked.setOptions()` to merge your options with `marked.defaults`.

Your Pull Request significantly reduces performance. Just try run `node test -t` before and after you PR.

Feder1co5oave:
It's true, I had a 10% performance drop with this. Seems like `merge`ing 1000 times is an expensive task, compared to the actual parsing and lexing.

Additionally, the task of setting unspecified options is taken care of by the `marked()` function, so if you use that (as do 99% of the users) you are fine.

solugebefola:
Thanks for the feedback.  I will see if I can increase the performance while maintaining the functionality for custom Parsers and Lexers.

joshbruce:
Flagging for 0.5.0 to concentrate on fixing known issues with handling different specs...we're getting there! Very exciting.

Feder1co5oave:
I think the right way to fix this would be to use the solution proposed in #907.
You `require('marked')`, you set the options thereby instantiating the lexer and parser with those options, and you can reuse those instances on multiple inputs.
Unfortunately this is a fundamental API change that would require a major version (or minor, since we're still in 0.x?) upgrade.

joshbruce:
Definitely think it would be in the 0.5.0 minor/major camp as tagged in the #907 issue. Fix the defects for those using Marked's current APIs; then change the APIs to let people opt-in to those breaking changes without having something we know is already broken. :)

joshbruce:
Believe we can close this in favor of the solution by @UziTech in #1203. Feel free to reopen.

PR #1018, opened 2018-01-19
[commonmark] Make link definitions commonmark compliant

Feder1co5oave:
I've made link definitions (almost 100%) compliant with the [commonmark spec], and added the related test cases, from the spec itself. They all pass, but example 167 has some issue with escaping that needs some attention to be fixed.

This (conflicts with) closes #856, which goes against the cm spec.

[commonmark spec]: http://spec.commonmark.org/0.28/#link-reference-definitions


joshbruce:
LGTM! (Which probably means I'm missing something somewhere. ;) )

Feder1co5oave:
Uhm just to be precise: single quotes in link titles are **not** against the spec. In fact, they are supported by this PR. I forgot about this :thinking: 
So, everything's good in the end.

Feder1co5oave:
About [this](https://github.com/chjj/marked/pull/1023#issuecomment-359206278) failing test, it was introduced in c3e00590:

I confirm. I should have put a note about this.
It is caused by the fact that "matched" double quotes in link titles, as in

    Foo [bar](/url/ "Title with "quotes" inside").

are not supported anymore.
This is actually by design since it complies with commonmark.
They decided to go against "matching" quotes in titles and break compatibility with `markdown.pl` (a sensible choice, IMO). Now quotes inside titles need to be backslash escaped.

The rationale is [here](http://spec.commonmark.org/0.28/#example-480)

Feder1co5oave:
Now that I think about this, it creates a slight incoherence in marked, because it was designed to behave exactly as `markdown.pl`, **when in `pedantic` mode**.

PR #1019, opened 2018-01-19
bench against commonmark instead of showdown

Feder1co5oave:
If we want to be competitive on performance nowadays, our competition is markdown-it and commonmark. Showdown is the slowest converter out there, and on my slow machine it takes a lot to bench.
I think it's pointless to compare against the notably slowest one, but we definitely should compare against commonmark. This PR does that.

See #963


joshbruce:
Good call. Flagging for 0.6.0 for now.

Feder1co5oave:
Also showdown brought in a lot of (dev)dependencies that we can now drop.
Currently my dependency tree on my branch looks like this:
```
$ npm list
marked@0.3.12 /home/federico/marked
├─┬ commonmark@0.28.1
│ ├── entities@1.1.1
│ ├── mdurl@1.0.1
│ ├── minimist@1.2.0
│ └── string.prototype.repeat@0.2.0
├─┬ front-matter@2.3.0
│ └─┬ js-yaml@3.10.0
│   ├── argparse@1.0.9 deduped
│   └── esprima@4.0.0
├── glob-to-regexp@0.3.0
├─┬ markdown@0.5.0
│ └─┬ nopt@2.1.2
│   └── abbrev@1.1.1
├─┬ markdown-it@8.4.0
│ ├─┬ argparse@1.0.9
│ │ └── sprintf-js@1.0.3
│ ├── entities@1.1.1 deduped
│ ├─┬ linkify-it@2.0.3
│ │ └── uc.micro@1.0.3 deduped
│ ├── mdurl@1.0.1 deduped
│ └── uc.micro@1.0.3
└─┬ uglify-js@3.3.10
  ├── commander@2.14.1
  └── source-map@0.6.1
```

Feder1co5oave:
I think we should put out somewhere in the README to install via npm with the following command:

    npm install marked --production --save

This way no (dev)Dependency is brought in when installing marked, since 99% of our users do no need to lint, test or bench against other parsers!

joshbruce:
@Feder1co5oave: Can we add the README update to this PR and then I will merge this and #1046 and publish 0.3.13 - which just leaves the lint thing over at #1020...no rush on that one.

joshbruce:
Sorry! Wrong button.

styfle:
@Feder1co5oave Would you like to move forward with this PR?

I think it makes sense and we could also use this as a opportunity to upgrade all the versions of `devDependencies` since these are pretty outdated.

PR #1022, opened 2018-01-20
Fixed example of setting Renderer

KostyaTretyak:
If my English is not very correct, please tell me how to write correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,14 @@ console.log(marked('I am using __markdown__.'));
 // Outputs: <p>I am using <strong>markdown</strong>.</p>
 ```
 
-Example setting options with default values:
+Create instance of `Renderer` only **after** you setting the necessary options.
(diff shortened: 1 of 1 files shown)


KostyaTretyak:
As tests have shown, for the marked this PR is not relevant.

Feder1co5oave:
I'm confused: what's this about?

KostyaTretyak:
@Feder1co5oave, currently `marked` options merge and are taken into account even **after** setting them, for example, to the Renderer's constructor.

For example, if we do that:

```js
marked.setOptions({renderer: new marked.Renderer});
```

And then do:

```js
marked.setOptions({ highlight: () => 'some text' });
```

The options passed to the `Renderer` constructor will have a `highlight` because the `setOptions()` method changes same object.

In my fork, `setOptions()` method returns new object with options every time I call this method.

PR #1023, opened 2018-01-20
[commonmark] Make blockquotes commonmark compliant

Feder1co5oave:
I've made blockquotes (almost 100%) compliant with the commonmark [spec], and added the related test cases, from the spec itself. They all pass, but I was not able to pass example 198 and 200 because they are related to list and code fences rules. This will need to be fixed in the future if anyone's up to it.

Many thanks to @KostyaTretyak who helped in solving related issues before

[spec]: http://spec.commonmark.org/0.28/#block-quotes

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -19,12 +19,12 @@ var block = {
   heading: /^ *(#{1,6}) *([^\n]+?) *#* *(?:\n+|$)/,
   nptable: noop,
   lheading: /^([^\n]+)\n *(=|-){2,} *(?:\n+|$)/,
-  blockquote: /^( *>[^\n]+(\n(?!def)[^\n]+)*\n*)+/,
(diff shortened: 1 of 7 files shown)


joshbruce:
1 failing test, just for tracking:

```
#36. literal_quotes_in_titles failed at offset 6. Near: "<p>Foo[bar][].</p><p>Foo<ahref="/url".


Got:
<p>Foo[bar][].</p><p>Foo<ahref="/url


Expected:
<p>Foo<ahref="/url/"title="Titlewith
```

Believe this test is actually incorrect. Think there should be a space `<a href=` otherwise browser may look at this in an odd manner.

Feder1co5oave:
This is caused by a compatibility break in the spec. The missing space you see is just caused by how the test system currently works - comparing after stripping all whitespace.

Read https://github.com/chjj/marked/pull/1018#issuecomment-359207001

PR #1029, opened 2018-01-22
Fix failing case

Feder1co5oave:
As explained [here](https://github.com/chjj/marked/pull/682#issuecomment-359208055) a failing test was introduced in #1018 as a result from merging #456. This pr fixes it.


joshbruce:
Thank you, sir!

PR #1030, opened 2018-01-22
Add linksInNewTab config option

ocdtrekkie:
All credit due to @drmuey, reference is #659 for his work. This should merge against the current codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ marked.setOptions({
   sanitize: false,
   smartLists: true,
   smartypants: false,
-  xhtml: false
+  xhtml: false,
+  linksInNewTab: false
 });
 
 console.log(marked('I am using __markdown__.'));
@@ -281,6 +282,13 @@ Type: `boolean`
 Default: `false`
 
 Self-close the tags for void elements (&lt;br/&gt;, &lt;img/&gt;, etc.) with a "/" as required by XHTML.
+
+### linksInNewTab
+
+Type: `boolean`
+Default: `false`
+
+Opens links in new tabs.
 
 ## Access to lexer and parser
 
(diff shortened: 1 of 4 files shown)


ocdtrekkie:
You wanted a big green button. I am looking at how to add a test and will add it to this branch.

joshbruce:
Taking 'em one at a time. I appreciate the zeal. Let's see where it goes. :)

ocdtrekkie:
I wrangled the testing setup a bit, the Linux VM I had handy is... outdated. I got an NPM error before and after adding my test, but after tweaking it a bit (I missed adding the `<p>` tags originally), it shows as completed.

ocdtrekkie:
I completely understand the desire for brevity. I believe this PR is very brief, and it touches an extremely small amount of code. The believe the likelihood of this PR leading to bugs is small. The only thing I can think of that someone will need to be aware of, is that if code is added to add any other properties to the `<a>` structure, they will need to have a leading space, as this code does (and the `title` tag also does), before the new property. It should coexist with any new properties added similarly to itself.

styfle:
I think you would want to add a few additional attributes to avoid [cross-origin vulnerabilities](https://mathiasbynens.github.io/rel-noopener/) such as `rel=noopener`. Also what about `rel=nofollow`? It's hard to know what the user wants in this scenario.

ocdtrekkie:
@styfle The original PR for this feature was created in 2015. I have pretty much abandoned all hope for marked at this point.

UziTech:
I think this feature would be better as an extension. it shouldn't be too hard to provide a renderer that does this.

styfle:
@UziTech Agreed, there are too many variables here that it doesn't make sense to be in core.

PR #1033, opened 2018-01-23
Make legacy test literal_quotes_in_titles pass

Feder1co5oave:
As I explained [here](https://github.com/chjj/marked/pull/1018#issuecomment-359207001):

With the new commonmark compliant rule for link definitions, literal (matched) doubles quotes in titles are not supported anymore, so this legacy test from markdown.pl failed. I re-added support for this (even if it is against the cm spec).
Backslash-escaped (non-matching) quotes are still supported.

I did not consider it worth to create a separate `pedantic` rule for this.
(pedantic means it is going to replicate markdown.pl behavior exactly)

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -29,7 +29,7 @@ var block = {
 };
 
 block._label = /(?:\\[\[\]]|[^\[\]])+/;
-block._title = /(?:"(?:\\"|[^"])*"|'\n?(?:[^'\n]+\n?)*'|\([^()]*\))/;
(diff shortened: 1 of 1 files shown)


joshbruce:
Once we get into 0.6.0 and beyond, we can probably get into other specs as well - including updates to pedantic and whatnot, I think. Multimarkdown - almost covered by GFM extensions, I think. There was another spec that I'm not sure if it is already part of the base CM or GFM extensions that did insertions and deletions and whatnot, for example. Cheers for this.

Issue #1041, opened 2018-02-02
Dependencies check-up


Feder1co5oave:
Follow-up from https://github.com/chjj/marked/pull/1020#issuecomment-362465220

I just removed the node_modules directory and run `npm update` on a fresh working copy.
This is what I get:

```
$ npm up
npm WARN deprecated gulp-util@3.0.8: gulp-util is deprecated - replace it, following the guidelines at https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5
npm WARN deprecated graceful-fs@3.0.11: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated graceful-fs@1.2.3: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
marked@0.3.12 /home/federico/marked
├── eslint@4.16.0 
├── eslint-config-standard@11.0.0-beta.0 
├── UNMET PEER DEPENDENCY eslint-plugin-import@>=2.2.0
├── eslint-plugin-node@5.2.1 
├── UNMET PEER DEPENDENCY eslint-plugin-promise@>=3.5.0
├── eslint-plugin-standard@3.0.1 
├── front-matter@2.3.0 
├── glob-to-regexp@0.3.0 
├── gulp@3.9.1 
├── gulp-concat@2.6.1 
├── gulp-uglify@1.5.4 
├── markdown@0.5.0 
├── markdown-it@8.4.0 
└── showdown@1.8.6 

npm WARN eslint-config-standard@11.0.0-beta.0 requires a peer of eslint-plugin-import@>=2.2.0 but none was installed.
npm WARN eslint-config-standard@11.0.0-beta.0 requires a peer of eslint-plugin-promise@>=3.5.0 but none was installed.
```

I think this needs some checking.
First of all, I think peerdependencies need to be explicitly installed alongside their dependants, so I'm gonna ~~add `eslint-plugin-import` and `eslint-plugin-promise` to #1020`~~.
This was taken care of by @UziTech in Feder1co5oave/marktex#1. Thanks!

Here's the updated section in `package.json`:

```json
  "devDependencies": {
    "eslint": "^4.15.0",
[...]

PR #1020, opened 2018-01-20
[linting] Use eslint to lint code. + minor refactoring + Travis CI setup

Feder1co5oave:
See the previous discussion at #999.

closes #1014 and closes #1028 .
closes #967 

See the commits list for the history of edits.

Apart from some unnoticeable edits, I've changed some minor code style (see `var` declarations) and refactored the `replace()` function in a OOP way. I believe it looks neat and self-documenting, I'd like to hear from others on this, and accept different proposals.

With this PR, eslint is completely happy with the source code as it is, and reports no warning or errors.
The programmer can run `npm run lint` to autofix code style, and get linting feedback before submitting new PRs.

Future steps: linting `test/index.js` :smile: 

Thanks a lot to @smhg and @joshbruce for their contributions.

Issue #984, opened 2017-12-25
[Links]: Consolidate issues


joshbruce:
This issue consolidates all issues related to the generating of links. Issues will be closed, PRs will remain open - unless they have merge conflicts.

Feder1co5oave:
I have to be honest Josh, I don't see the rationale behind this methodology. I'm trying to help resolve the issues, but how can I keep track of which of them are solved/unresolves if you close all of them indiscriminately?
For me it would be best to close resolved or non-issues, mark duplicates and leave open those who might lead to fixing marked the right way.

Federico

joshbruce:
@Feder1co5oave: That's fair and well-appreciated feedback. I might have been looking at it primarily from me scoping things out alone and thinking of what might be stopping people from participating in the conversation or doing what I was doing. 700+ issues across 15+ pages...versus 100 issues 6 of which are there just to group things into themes...solve the analysis paralysis problem. 

I was also thinking that we would be able to *remove* the references once we had triaged them (never tried this methodology before)...that's apparently not possible.

We could create more labels as a filtering mechanism - that's really all this methodology gives us beyond the reduction - the "one list" so to speak (this one spans 500 issue identifiers; so, maybe it makes the original intent more obvious).

How about this...in the comments of the issues you think might actually help and aren't duplicates of something else - say so - I will reopen those issues but leave the rest closed (I think I've only seen the ones you thought were non-issues so haven't done that yet). I will also create labels for the groupings we have so far to allow simple grouping of future issues. We'll still leave the PRs open by default - we might even be able to update the submission with "Fixes" to close a related issue.

Sound fair and reasonable?

Feder1co5oave:
Yes Josh, that sounds awesome. I see that your intent was to filter out and group together related issues, and that's good. IMO labels work great for that, since discussions relative to the issues mostly happen in the issue pages themselves. Thank you for your help!

Feder1co5oave:
Just to recap link-related issues:

- [x] #482 should hopefully be fixed by my #996
- [ ] #366 #574 #690 #619 #736 #859 #586 still to be fixed (balanced/escaped parens in URLs - this is a priority). There's PR #600 but it looks awful to me.
- [ ] #498 is to be fixed to comply with the spec (not a priority though)
- [ ] #857 is really about headings, and there's no perfect way to generate IDs; still we could improve the default behaviour
- [ ] #890 is to be considered
- [ ] 872 haven't looked into this
- [ ] 829 haven't looked into this
[...]

PR #1046, opened 2018-02-09
Use uglifyjs CLI instead of Gulp to minify

Feder1co5oave:
Following from #1041.
By dropping gulp we get rid of a lot on unused dependencies.
[Uglify-js3] now takes care of minification. A shortcut to minify is `npm run build`.

Make sure to `npm prune && npm install` after pulling.

If you want to make sure that the minified version behaves as expected, run `node test --minified` to test it.

[uglify-js3]: https://github.com/mishoo/UglifyJS2


joshbruce:
@Feder1co5oave: 

Can you expand on this bit??

> Make sure to npm prune && npm update after pulling.

Feder1co5oave:
To update the dependencies, as you would on a fresh checkout. Pruning is not mandatory, it gets rid of everything related to gulp. You can even `rm -r node_modules` and `npm install`.

joshbruce:
Okay, just checking to make sure. Never saw the `prune` command before; thanks for the introduction.

Given this would help reduce the overall size of the installation footprint, I'm thinking we'll go ahead and merge this. Maybe do the lint as well. Then publish despite not strictly part of the 0.4.0 goal (don't want to lose too much of the consistent flow of updates we've had (seems like people are taking notice that the project is alive again). Thoughts??

Tagging @UziTech as well.

UziTech:
I think this is a step in the right direction but I would ultimately like to start using ES6+ to modularize the codebase so it is cleaner. At that point we will have to replace this with babel or some other transpiler.

joshbruce:
@UziTech: Fair. Then let's go ahead and merge this in as that step. Maybe start an Issue on the ES6 question to aid the 0.5.0 milestone??

Feder1co5oave:
There's #746 for discussing modularization! Let's move over there

Feder1co5oave:
@joshbruce Wow, I wrote the following message right after [your comment](https://github.com/chjj/marked/pull/1046#issuecomment-364305435) but totally forgot to send it out! :man_facepalming: 
Anyway, here it is for y'all to enjoy :laughing: 

---

Actually I think a simple `npm install` takes care of everything inside node_modules.

Anyway, this PR, #1020 and #1019 take care of everything about dependencies we discussed in #1041. I wouldn't bother publishing for just these ones, but we have 62 commits since v0.3.12 and we fixed [quite a lot](https://github.com/chjj/marked/compare/v0.3.12...master) of issues!

I terms of commonmark compliance, we fixed
- link definitions
- blockquotes
- autolinks (gfm fashion too)
- thematic breaks

Also from the user perspective we "fixed" the "replace is not a function" stupid error message so hopefully we won't get any new issue opened because of that.
I'm currently working on fixing links (so #984 here I come) but that could take a while, and also some other things for the future.
Your choice to release or wait. Anyhow it's good to send out the message that we're alive and fighting.

PR #1053, opened 2018-02-15
Fix GitHub language via override

styfle:
GitHub shows marked as HTML when in fact, it is a JS project.
This adds an [override](https://github.com/github/linguist#overrides) so the test folder is ignored when detecting the primary language because that is where all the html lives.

### Before

before

### After

after


joshbruce:
Nice! Merging now.

PR #1060, opened 2018-02-19
revert number of capturing parens in #1013

### LINKED ISSUE #1059: Semver 1.0 release
I noticed that `marked` [starts fresh](https://github.com/markedjs/marked/releases/tag/0.3.14) in this year. However this release broke a dependent. Noticing that is 0.3.14, I wonder if it would be in the spirit of "starting fresh" and go for the 1.0 release and in that sense start following semver (breaking change: semver: major)?

UziTech:
This reverts the number of capturing parentheses in `inline.code` grammar from #1013

The regex should still function the same.

fixes #1059


lib/marked.js
```diff
@@ -468,7 +468,7 @@ var inline = {
   nolink: /^!?\[((?:\[[^\]]*\]|\\[\[\]]|[^\[\]])*)\]/,
   strong: /^__([\s\S]+?)__(?!_)|^\*\*([\s\S]+?)\*\*(?!\*)/,
   em: /^_([^\s_](?:[^_]|__)+?[^\s_])_\b|^\*((?:\*\*|[^*])+?)\*(?!\*)/,
-  code: /^(`+)(\s*)([\s\S]*?[^`]?)\2\1(?!`)/,
+  code: /^(`+)(?:\s*)([\s\S]*?[^`]?)(?:\s*)\1(?!`)/,
```
Feder1co5oave: There's no need to capture then!
`` /^(`+)\s*([\s\S]*?[^`]?)\s*\1(?!`)/ ``

Feder1co5oave:
+1, thanks

This looks more [commonmark](http://spec.commonmark.org/0.28/#code-spans) compliant too (they don't require that leading and trailing spaces match)

lib/marked.js
UziTech: good catch

joshbruce:
This is a merge + publish (0.3.15), yeah?

If so, makes me think we should have another owner who can publish...I’m at dinner, but if one of y’all could publish, we could increase our flow rate for hot-fix type things.

Thoughts??

UziTech:
yeah, merge + publish

I would be ok with more permission.

joshbruce:
@UziTech @Feder1co5oave - Any tests we can write to guard against future regression?? (Really looking forward to the CI thing.)

Publishing

styfle:
Is anyone working on setting up CI? Maybe that's something I can help with.

Feder1co5oave:
It's been set up in #1020, head over there and give us some advice if you
wish!

PR #1061, opened 2018-02-19
[WIP] Update read me with TOC & other changes

joshbruce:
Working on preparing to add @Feder1co5oave and @UziTech to the list of folks who can publish to NPM as well as possibly breaking up the README into separate files to help users out a bit as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,40 @@
 
 [![NPM version](https://badge.fury.io/js/marked.svg)][badge]
 
-## Install
+<ul>
+  <li><a href="#install">Installation</a></li>
+  <li><a href="#usage">Usage</a></li>
+  <li><a href="#marked"><code>marked()</code></a></li>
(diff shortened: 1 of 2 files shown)


joshbruce:
screen shot 2018-02-18 at 10 31 38 pm

README.md
```diff
@@ -17,7 +50,7 @@ or if you want to use the `marked` CLI tool (not necessary when using npm run-sc
 npm install -g marked
 ```
 
-## Usage
+<h2 id="usage">Usage</h2>
```
Feder1co5oave: I think github auto generates IDs on headings? It should be safe to reference #usage inside the document

joshbruce: Could be fair. Was mentioned in the review on #1081 as well. I've always run into issues there; so, normally end up doing it by hand. Also, could make it more difficult to use in a non-GFM setting; at least here we know the TOC will always work.

PR #1062, opened 2018-02-19
Release 0.3.15

joshbruce:
Something seems odd. The `.min` didn't seem to change. Also, not running from a fork, and I should be...sorry.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "marked",
-  "version": "0.3.14",
+  "version": "0.3.15",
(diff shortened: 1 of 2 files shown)


UziTech:
We need to get the build process on prepublish or something like that

Feder1co5oave:
You need to `npm run build` before every release, or set up the prepublish (or prepack?) hook

joshbruce:
I swear I did run it. Wrote the instructions while I was doing them.

Gonna see about doing the pre-publish thing as well...removes a step anyway.

PR #1063, opened 2018-02-19
[WIP] Updates to readme and license

joshbruce:
Working on preparing to add @Feder1co5oave and @UziTech to the list of folks who can publish to NPM as well as possibly breaking up the README into separate files to help users out a bit as well.

screen shot 2018-02-18 at 10 31 38 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,13 @@
 [![NPM version](https://badge.fury.io/js/marked.svg)][badge]
 
 <ul>
+  <li><a href="#philosophy">Philosophy</a></li>
   <li><a href="#install">Installation</a></li>
   <li><a href="#usage">Usage</a></li>
   <li><a href="#marked-call"><code>marked()</code></a></li>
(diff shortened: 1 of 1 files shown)


styfle:
I wonder if we should make the README smaller (since this gets published to NPM) and move most of the docs into the `/docs`. Then this folder can be used for GitHub Pages users can get a nice web view of the docs as well as a working demo. Then we can add `/docs` to `.npmignore` file so it doesn't get published.

This pattern can be seen here: https://github.com/styfle/copee

joshbruce:
@styfle: I've seen a similar pattern to allow GitHub Pages. Also been thinking a bit about the whole GitHub Pages thing as well.

Having said all that, totally agree on breaking up the README. It's on my list of things to do. Maybe it will become a GH Pages site here or in a different repo (not that we have the org thing going on); not sure.

In either case - it look like GitHub and I got very out of sync somehow. So, going to close this and work on a resubmission.

Issue #1051, opened 2018-02-15
Still the co-pilot. So, now what?


joshbruce:
All right, we have an org. I can see the settings tab - what now.

1. This is the new home of Marked. Having said that, it's an organization account, which gives us more flexibility related to having more than just the Marked repo.
2. We are sticking with the milestones already established; therefore, still considered in beta mode using the zero major to indicate that to the world: https://github.com/markedjs/marked/milestones
3. We will use Travis CI for now as I believe @UziTech had plans and ways of doing this. Let me know what I can do to make that whole thing easier.
4. @UziTech, @Feder1co5oave, and @styfle have been added as collaborators on the repo; this solves the PR merging problem. Recommend maintaining the `master` is always deployable (publishable to NPM) model. Please be considerate - again going back to the idea of having separate CONTRIBUTING, LICENSE, and CODE OF CONDUCT markdown files in the repo. I'm probably going to keep things strict for now and open as needed/requested.

***

About me. I am the owner of a company called [8fold](https://8fold.pro). For my day job I am a consultant working on various IT projects and programs in various capacities. I've been looking for an excuse to start running CI with my own projects, but could not justify the expense - given that it's just me (a lot of my projects are private repos - guess I could have used some of the public one); however, I'm also testing out a new principle I would like to see become a bigger thing in the software development profession: If I create it, or acquire it, I help maintain it.

Marked and this crew are a good reason to start these experiments. Further, I'm helping another crew open source one of their internal projects and we are looking to use Circle CI for that as well - so, this is kind of like off the job training for me.

Built my first site in 1998...all I wanted to do was, what we now call UX, unfortunately, at the time, that term hadn't really become a thing; so, I sort of let myself get pushed into development. Now, I'm working toward being a coach on Agile Software Development teams after freelancing full time from 2007 to 2010 as an all around business consultant with a focus on UX and design. I'm also a value-based, principle-driven human; so, that's me. :)

Tagging in @chjj and @KostyaTretyak as well.

joshbruce:
Very exciting to see that settings tab. Thanks @chjj!

joshbruce:
I've created my fork to make good on the promises of modifying the contributors list:

https://docs.npmjs.com/files/package.json#default-values

> "contributors": [...]
If there is an AUTHORS file in the root of your package, npm will treat each line as a Name <email> (url) format, where email and url are optional. Lines which start with a # or are blank, will be ignored.

[...]

PR #1066, opened 2018-02-20
Simplify script’s footer

mina86:
1. ‘typeof foo === "undefined"’ is just a convoluted way of writing
   ‘foo === undefined’.  The slight risk of ‘undefined’ being redefined
   (since it’s not a keyword) can be easily averted by using ‘void 0’.
   In the end, ‘typeof foo === "undefined"’ becomes ‘foo === void 0’
   which is shorter and does the same thing.

2. There is really no need to use Function.prototype.call.  Since the
   wrapper function uses it’s ‘this’ variable exactly once, whatever
   function run as call argument does can be moved inside of the wrapper.

This shortens and simplifies script’s footer.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1369,14 +1369,11 @@ marked.inlineLexer = InlineLexer.output;
 
 marked.parse = marked;
 
-if (typeof module !== 'undefined' && typeof exports === 'object') {
+if (module !== void 0 && typeof exports === 'object') {
   module.exports = marked;
 } else if (typeof define === 'function' && define.amd) {
   define(function() { return marked; });
 } else {
-  this.marked = marked;
+  (this || (window !== void 0 ? window : global)).marked = marked;
 }
-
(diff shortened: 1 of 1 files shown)


UziTech:
Thank you for your interest in marked.

We will be making a lot of changes to the code base in the future to make marked more modular so I'm sorry to say that these changes won't be needed. Read #1051 if you haven't already.

Feder1co5oave:
@mina86 

I'd like to answer about the technicalities of the pull request:

1. It seems the `typeof === 'undefined'` check is the de-facto standard and is considered the 100% safe way to check for undefinedness so we'll keep it that way. Plus I really don't think doing `=== void 0` is more readable than the former, but actually worse.

2. I recognize the umd boilerplate currently in place is awful and I rewrote it in my PR #1020.
Look at ad6484b90 to get an idea and feel free to put in any feedback you have.

Cheers

PR #1076, opened 2018-02-24
Release 0.3.17 (template updates) - ??

joshbruce:
## Description

- Fixes #1075 

## Submitter

- [x] `$ npm version` has been run.
- [x] Release notes in [draft GitHub release](https://github.com/markedjs/marked/releases) are up to date
- [x] Reviewer checklist is complete.
- [ ] Merge PR.
- [ ] Publish GitHub release using `master` with correct version number.
- [ ] `$ npm publish` has been run.
- [ ] Create draft GitHub release to prepare next release.

Note: If merges to `master` occur after submitting this PR and before running `$ npm pubish` you should be able to

1. pull from `upstream/master` into the branch holding this version,
2. run `$ npm run build` to regenerate the `min` file, and
3. commit and push the updated changes.

## Reviewer

- [x] Version in `package.json` has been updated (see [RELEASE.md](https://github.com/markedjs/marked/blob/master/RELEASE.md)).
- [x] The `marked.min.js` has been updated; or,
- [ ] release does not change library.
- [x] case_insensitive_refs is only failing test (remove once CI is in place and all tests pass).
- [x] All lint checks pass (remove once CI is in place).

PR #1081, opened 2018-02-25
Update docs

joshbruce:
## Description

This PR helps break up the README into more consumable chunks.

## Submitter

- [ ] Test(s) exist to ensure functionality works (if no new tests added, list which tests cover this functionality); or,
- [x] no tests required for this PR.
  
## Reviewer

- [x] Draft GitHub release notes have been updated.
- [x] case_insensitive_refs is only failing test (remove once CI is in place and all tests pass).
- [x] All lint checks pass (remove once CI is in place).
- [x] Merge PR

PR #1082, opened 2018-02-26
Setup .editorconfig

joshbruce:
**Marked version:** 0.3.16

**Proposal type:** project operations

Tabs are rendered as literal tab characters, not spaces using Sublime Text 3.

Was working on #1076 and noticed the code samples were rendering funny when it came to whitespace and text wrapping. Realized it was because my editor was using literal tabs instead of spaces.

A `.editorconfig` file following similar rules as lint options...specifically, spaces not tabs.

See also http://editorconfig.org

Could be accepted before or after #1076

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ insert_final_newline = true
 indent_style = space
 indent_size = 2
 
-[*.md]
+[*.md, !test/*md]
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*, *.json, *.js]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.md]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+translate_tabs_to_spaces = true
+indent_style = space
+indent_size = 4


joshbruce:
@UziTech: "Let the editor figure it out" - agreed...that's why I fell in love with `.editorconfig` - it let's the project tell the editor what to do. 😊 

I have a project that is a blend or PHP, JS, Sass, and so on. PHP uses spaces not tabs - four spaces for each (if you are following the PHP-fig coding standards, which is common for the community). The JS and CSS communities seem more inclined to 2. As long as contributors use an editor that understands them, everyone is doing the same thing without having to wait for the linter to tell them they're wrong. 😆 

So, yeah, `.editorconfig` for the win.

.editorconfig
```diff
@@ -0,0 +1,17 @@
+root = true
+
+[*, *.json, *.js]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.md]
```
Feder1co5oave: I wouldn't impose rules on markdown files because we use them to make tests, and sometimes we want to break out of all these rules. Is there a way to apply this only to documentation files?

Feder1co5oave:
Personally I use tabs for indentation with 4 spaces width (or 2 spaces width for javascript most of the time) but it doesn't really matter to me either.
I use Sublimetext3 and I configured it by clicking on "tab size: ?" in the bottom right corner. There you click on "use spaces for indentation" and "tab width: 2". If you screwed something up before setting this, you can click on "convert indentation to spaces" and it'll magically fix everything for you.

The beauty is that if you use tabs in `marked.js` or `test/index.js`, eslint complains:
```
$ npm run lint

> marked@0.3.12 lint /home/federico/marked
> eslint --fix lib/marked.js test/index.js


/home/federico/marked/test/index.js
  119:2  error  Mixed spaces and tabs                 no-mixed-spaces-and-tabs
  119:4  error  Unexpected tab character              no-tabs
```

.editorconfig
styfle: I agree. This could cause problems if it is not limited the docs.
Maybe remove this check for *.md and it can be added when we create a `/docs` folder with GitHub Pages.

Feder1co5oave:
Note: you need I plugin to have this working in SublimeText, Atom, Emacs, gedit, Eclipse, Notepad++, Vim, VSCode and others.

Feder1co5oave:
Also, should we include a _brief_ "coding style" section in our "contributing" documentation (be it in README on in some other file)?

joshbruce:
Agreed on the markdown believe we can do exclude...will resubmit shortly. Also agreed on the needing a plug-in in most cases (vendors not supporting it natively); however, we don't even have the option to support it right now. :)

I was first introduced to it only a couple of years ago and now I don't leave home without it. 😉 

https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig - It's gaining traction in a lot of projects I trust - especially the A11Y.

joshbruce:
Also, found out the VIM can't support it period while working #1083 - so, I've updated #1081 with a checkpoint to run `npm test` prior to submitting the PR, which will run `npm test:lint` now and instructions related to our NPM scripts, namely running `npm run lint` to have the linter fix any lint issues.

See also #1089 - just created it.

Issue #1089, opened 2018-02-27
Coding standards


joshbruce:
**Marked version:** n/a

**Proposal type:** project operations

## What pain point are you perceiving?

More contributions and a lack of defined and documented coding standards.

See also #1082, #1081, #1083, and a few other conversation areas where we've expressed having this pain point.

## What solution are you suggesting?

1. We have lint checks now.
2. We have Travis now.
3. We have NPM scripts that fix lint issues for us.
4. #1081 has an updated checklist for contributing and a list of our NPM scripts and what they do - not merged yet.

What else do we need or want to help contributors know what is expected in their code?

intcreator:
As a new contributor, should I list a few things here that I don't immediately know?

One of the biggest things is whether or not contributors should use ES6/ES7 features (like arrow functions, object spread, filter, map, reduce, etc.).

I tried hacking the code to customize it a bit and I really have no idea what I'm doing despite having two years of experience working with JavaScript.  Maybe if someone wrote a doc explaining the underlying mechanics of how the library works newcomers could read it and understand the vision better.  This kind of architecture documentation (and indeed an architect who is in charge of the design) is vital to software projects.

UziTech:
No ES6+ until we figure out a transpile build step. We still support older browsers.

joshbruce:
That's a fair question...let's go over to #1096

joshbruce:
Oh! Since you mentioned architecture, check out #746. Get some idea of where we're planning on going. Fortunately, and unfortunately, we have a large user base; so, we're really trying to satisfy the rest of the zero-major promises, before doing anything drastic.

UziTech:
I think eslint covers this pretty well.

PR #1096, opened 2018-02-28
[WIP]: Update contributing

joshbruce:
**Marked version:** 0.3.17

## Description

Moving this [comment](https://github.com/markedjs/marked/issues/1089#issuecomment-369114440) from @intcreator here to continue the broader conversation on helping new contributors get up to speed with marked.	

## Contributor

## Committer

- [ ] cm_autolinks is the only failing test (remove once CI is in place and all tests pass).
- [ ] All lint checks pass (remove once CI is in place).
- [ ] Merge PR


joshbruce:
@intcreator: If you would like to talk privately, we can find a different venue, but you're one of the first to ask regarding new contributor on-boarding; so, wanted to acknowledge that and get to know the issues you're having a bit more. We can also try something other than text-based communication, if you would prefer.

intcreator:
That looks like a good start.

**Browser compatibility**: We should probably support modern browsers (Chrome, Firefox, Edge, and Safari current and previous two versions) and (sigh) IE 11.  As far as I know twice as many people still use IE as Edge.  I think that set of browsers is pretty standard.  [Vue](https://github.com/vuejs/vue#browser-compatibility) and [React](https://reactjs.org/docs/react-dom.html#browser-support) support that same set of browsers except they go back to IE 9.  If anyone really wants to, we can make it happen, but IE 11 is probably ~~painful~~ far enough.

**Node support**: I know less about what versions of Node are currently in widespread use.  Perhaps we can support the oldest version [supported by Node.js](https://github.com/nodejs/Release#release-schedule) itself?  Node 4 will be in maintenance until April so I wouldn't mind passing it by.  Node 6 will enter maintenance in April and continue until April 2019 so maybe we can start there.

joshbruce:
@intcreator: Thanks!

I updated the *Submitting PRs continued* section to account for where we are right now. Let me know your reaction to that.

Also added the *The marked architecture* section. For that one, if you had to guess, how would you describe each of the parts that have the ellipses (...)?

Tagging @styfle as well for situational awareness and silent partner for now.

UziTech:
Right now I'm pretty sure marked works all the way back to IE8 (some people just don't have enough XP) and our travis-ci runs node 0.10. I think those are good limits for 0.x

Node 6 is the [first to support](https://node.green) 99% of ES6. I agree with following the lifecycle of node.

I would say that starting with 2.x we only support the latest evergreen browsers (i.e. not IE) and follow node lifecycles (so probably Node 6)

styfle:
> I would say that starting with 2.x we only support the latest evergreen browsers (

Which browsers should be supported for 1.x then?

UziTech:
same as 0.x

joshbruce:
This is interesting to me from the perspective of the CLI (not a browser)...we're not really dependent on the browsers per se, yeah? Instead we are dependent on the JS engine running marked, just a thought (could totally be wrong, if I am let me know).

UziTech:
You are correct. The browser/node version just comes with a certain JS engine that runs marked.

That is why browsers started auto updating so we can just say we support the latest stable version of each browser.

Wouldn't it be awesome if node started auto updating 🤣

intcreator:
I think it would also be a good idea to mention which Markdown specifications Marked will adhere to and whether or not Marked can add features with flags or maintain itself as an easily extensible base.

The README mentions that both CommonMark and GFM are supported, but I would go into a little more detail on the mechanics.  For example, these questions come to mind:

1. Do we always start to adhere to the latest version of the spec immediately or wait to adopt it at a certain version/milestone?
2. What happens when a user uses CommonMark vs GFM under the hood?
3. Do we have plans for supporting other flavors/dialects of Markdown?

joshbruce:
@UziTech: Funny story. I'm on a project where the platform automatically updated Node for security - broke all our stuff...so, not sure I can come with you on the autoupdate; unless we're strictly speaking about the JS engine.

@intcreator: Let's put a pin in that. Gonna create an issue.

joshbruce:
@intcreator: Just to make sure this was seen and is covered (we sometimes get sidetracked and lose a request in the excitement of the conversation and collaboration).

> Also added the *The marked architecture section*. For that one, if you had to guess, how would you describe each of the parts that have the ellipses (...)?

styfle:
> same as 0.x
@UziTech If we're not dropping support for anything in 1.0.0 then what is the breaking change? And what is the timeline to release 1.0.0?

UziTech:
> what is the breaking change? 

Most likely a bunch of stuff

A lot of projects that I have seen depend on marked use a specific version (i.e. "0.3.17") because something could easily break while everything is changing (spec support, etc.)

The idea behind v1.x is that dependents can finally use "^1.0.0" and not have to worry about us deciding that we are or aren't going to support a spec.

intcreator:
The [first two FAQs](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase) in Semantic Versioning suggest to me that 1.0 doesn't _require_ breaking change but rather signifies a stable release/API.

Nevertheless, I'm sure if we get Marked where we want by 1.0, it will break something.  It's software after all...

joshbruce:
@intcreator: Fair.

@UziTech and @styfle: Anyone we can ask (or beg)? Maybe @Nalinc from #746 

Just to make it easier to see the ask:

> What is the single responsibility or purpose of the things identified with ellipses (...)??

joshbruce:
All right @intcreator, @UziTech, @styfle, and @davisjam - My guesses on what the single responsibility of the objects are in...am I wrong, do you think. See also #1102 

Gonna be done for the day. Need to catch up on some things. Let me know if something explodes. 😂

CONTRIBUTING.md
```diff
@@ -56,6 +56,80 @@ Marked provides templates for submitting both pull requests and issues. When you
...
+4. `var originIndependentUrl`: ...
+
+Functions:
+
+1. `escape()`: ...
+2. `unescape()`: ...
+3. `edit()`: ...
+4. `resolveUrl()`: ...
+5. `noop()`: ...
+6. `merge():` Merges one or more objects to create a new object. (Why not use Object.assign??)
```
intcreator: `Object.assign` is an ES6 feature [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Specifications), so it'll have to wait until 2.0

joshbruce:
@styfle: Gonna close and make a note to update it later. Thinking Roadmap will be part of the docs?? Trying to minimize the places a developer has to go to learn about all the things.

joshbruce:
See #1162

Issue #1092, opened 2018-02-27
Auto-linking - cm_autolinks test


joshbruce:
**Marked version:** 0.3.17

**Proposal type:** other

## What pain point are you perceiving?

Travis CI is still in untrustworthy state because we have a failing test: `cm_autolinks`

I don't think auto-linking is a feature from a specification is it? If not, is it something marked took on at some point and we're trying to keep alive for now to maintain backward compatibility?

I don't know about the rest of y'all but I keep getting emails from Travis saying, "Build X is still failing".

## What solution are you suggesting?

1. How are we on fixing autolinking?
2. _Should_ we fix autolinking and why?
3. Can/Should we remove the test for now until it gets fixed - create an issue so we know it's something we need to do?

styfle:
Original Markdown spec: https://daringfireball.net/projects/markdown/syntax#autolink
CommonMark spec: http://spec.commonmark.org/0.27/#autolink

intcreator:
Failing tests kind of drive me insane, so I'd like to investigate this a bit further.  Feder1co5oave mentioned that he had a branch that addressed this issue [in a comment](https://github.com/markedjs/marked/pull/1065#issuecomment-366810372), but the [actual branch](https://github.com/Feder1co5oave/marktex/commits/encode_url) is a personal fork.  My questions are:

1. Is there enough value to justify salvaging this branch?
2. If so, what additional tests need to be added (per Feder1co5oave's comment)
3. If not, is there an easy alternative fix?

Here is the [diff of Feder1co5oave's branch with Marked master](https://github.com/markedjs/marked/compare/master...Feder1co5oave:encode_url).

joshbruce:
1. I think so?? (He does really great work.)
2. Which comment, not sure I'm following??
3. I do not know and not sure who to tag.

Also, updated `good first issues` with the consolidation issues Feder1co5oave was, for better or worse, taking on almost entirely by himself. See #984, #1036, and #1043 for how our process and interactions were working there. 

Once we get the architecture stuff a little bit more figured out from #1096, I might be able to help with straight [refactoring](https://www.amazon.com/Refactoring-Improving-Design-Existing-Code/dp/0201485672) (things that only improve readability) code submissions; thereby, not breaking or trying to fix anything before I actually know what's going on better.

intcreator:
1. Cool let's do it.
2. It's the same comment that I linked in the description
3. Let's focus on the branch then

[...]

Issue #1043, opened 2018-02-02
[Consolidated proposal] Features to deprecate


joshbruce:
Over the years the community has added features to Marked based on issues and user feedback; while not always a bad thing, sometimes it can cause a project to go beyond sustainable scope for itself (as we kinda saw). To try to rein things back in a bit, there are some features we may want to deprecate because they take Marked beyond its intended scope; thereby, making it more difficult for a small band to maintain.

This Issue is for capturing those suggestions, which will most likely be implemented during the 0.5.0 and beyond releases:

- [ ] ??

Use comments to discuss, will update the task list as we gain consensus and whatnot...that way we don't add too many issues that are future-planning.

joshbruce:
Think we should deprecate header ids being handled by Marked. I don't think any of the supported flavors of specifications actually has this functionality and it's something that can be better performed as a secondary pass on the DOM, in my experience. Also, could be something a use could add as extended functionality to reduce our maintenance overhead.

Tagging @UziTech and @Feder1co5oave to make sure y'all see the ticket.

Feder1co5oave:
My vote goes to deprecate the `pedantic` option :laughing:
I know chjj was in love with it, but it just doesn't make sense now that we have commonmark. Would be curious to see how many users actually enable it.

joshbruce:
That's fair re pedantic. I don't know how many peole know the Daring Fireball spec is even still out there. How different is it from CommonMark though?? Don't we get most of it "for free" by virtue of having CommonMark. Might be an extend + override path.

Pedantic = All these properties
CommonMark = Has Pedantic + these properties - these Pedantic properties.
GFM = Has CommonMark + these properties.
MultiMarkdown (this is totally made up at this point, just an example) = Has GFM

Know what I mean??

Martii:
@joshbruce 

Re: 
> Think we should deprecate header ids being handled by Marked.

I believe we currently utilize extending header ids to our own flavor for easier document traversal.

> Will we be supporting syntax highlighting of languages such as `c` or `r`?

We have the ability to utilize the full compliment of languages however most ppl use `js` and `diff`. We match GitHubs implementation as close as possible which includes other identifiers of that nature.

joshbruce:
Hey @Martii thanks for showing up! 

Just to make sure I'm following, don't think we've interacted much, so pardon the bumps in communication, really glad to see a new face (though I feel like I should know you...been a lot of peopling for recently, sorry).

> we currently utilize... and We have the ability to utilize

[...]

PR #1102, opened 2018-02-28
🚧[Experimenting]: Refactoring marked js🚧

joshbruce:
**Marked version:** 0.3.17

## Description

Experimenting with the codebase to learn more and try to make marked easier to read without fundamentally breaking things. 

**Again, this is experimental, please do not merge!**

Having said that, this should demonstrate how we can effectively alter how easy it is to read marked without actually changing the functionality. Please see code comments for questions I have regarding why certain decisions were made and whatnot.

I've also noted areas where I do not know if it is possible to test a unit of functionality with our current testing harness; therefore, tagging @UziTech.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -223,6 +223,8 @@ var originIndependentUrl = /^$|^[a-z][a-z0-9+.-]*:|^[?#]/i;
 
 /**
  * Block-Level Grammar
+ *
+ * Note: if you need to call edit(), please place member after object definition. 
  */
 
 var block = {
@@ -239,18 +241,19 @@ var block = {
   table: noop,
   lheading: /^([^\n]+)\n *(=|-){2,} *(?:\n+|$)/,
   paragraph: /^([^\n]+(?:\n?(?!hr|heading|lheading| {0,3}>|tag)[^\n]+)+)/,
-  text: /^[^\n]+/
+  text: /^[^\n]+/,
+  bullet: /(?:[*+-]|\d+\.)/,
+  item: /^( *)(bull) [^\n]*(?:\n(?!\1bull )[^\n]*)*/,
+  _label: /(?:\\[\[\]]|[^\[\]])+/,
+  _title: /(?:"(?:\\"|[^"]|"[^"\n]*")*"|'\n?(?:[^'\n]+\n?)*'|\([^()]*\))/
+
 };
 
-block._label = /(?:\\[\[\]]|[^\[\]])+/;
-block._title = /(?:"(?:\\"|[^"]|"[^"\n]*")*"|'\n?(?:[^'\n]+\n?)*'|\([^()]*\))/;
 block.def = edit(block.def)
   .replace('label', block._label)
   .replace('title', block._title)
   .getRegex();
 
-block.bullet = /(?:[*+-]|\d+\.)/;
-block.item = /^( *)(bull) [^\n]*(?:\n(?!\1bull )[^\n]*)*/;
 block.item = edit(block.item, 'gm')
   .replace(/bull/g, block.bullet)
   .getRegex();
@@ -284,14 +287,12 @@ block.blockquote = edit(block.blockquote)
   .replace('paragraph', block.paragraph)
   .getRegex();
 
-/**
- * Normal Block Grammar
- */
-
 block.normal = merge({}, block);
 
 /**
  * GFM Block Grammar
+ *
+ * See inline??
  */
 
 block.gfm = merge({}, block.normal, {
@@ -308,6 +309,8 @@ block.gfm.paragraph = edit(block.paragraph)
 
 /**
  * GFM + Tables Block Grammar
+ *
+ * See inline??
  */
 
 block.tables = merge({}, block.gfm, {
@@ -316,9 +319,10 @@ block.tables = merge({}, block.gfm, {
 });
 
 /**
- * Inline-Level Grammar
+ * Regex for inline-level HTML elements 
+ *
+ * Note: if you need to call edit(), please place member after object definition. 
  */
-
 var inline = {
   escape: /^\\([\\`*{}\[\]()#+\-.!_>])/,
   autolink: /^<(scheme:[^\s\x00-\x1f<>]*|email)>/,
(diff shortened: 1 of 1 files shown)


joshbruce:
I really don't trust our test bed to accurately reflect the desired behavior and use cases, to be honest.

joshbruce:
@intcreator: Interesting findings and notes in here, might help. Maybe we should slow down on trying to fix issues and work on making this thing more readable...?? Or, I'll do that, you do you, and we'll meet in the middle??

intcreator:
It looks like the diff got messed up so it's really hard for me to tell what's going on still.  As far as fixing vs. making the code legible, I vote we get the builds passing (just fixing #1092) then we can focus on understanding the code more.  It's just way easier to check if things are still okay if builds are passing.

lib/marked.js
```diff
@@ -7,8 +7,223 @@
...
+
+/**
+ * @param  src string
+ * @param  opt (optional) Object
+ * @param  callback (optional) Object
+ *
+ * @return string
+ */
+function marked(src, opt, callback) {
+  // ^^ Can't we do something like marked(src, opt = {}, callback = function())
```
intcreator: Default parameters are ES6 [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Specifications).  Maybe in 2.0.

lib/marked.js
```diff
@@ -7,8 +7,223 @@
...
+
+    // If we don't have a callback honestly passed in, opt becomes null,
+    // then we set it to a merged of the defalts + opt, or an empty object??
+    opt = merge({}, marked.defaults, opt || {});
+
+    // Not sure what all is happening here.
+    var highlight = opt.highlight,
+        tokens,
+        pending,
+        i = 0;
```
intcreator: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator), the comma operator evaluates each operand and returns the value of the last operand.  Here, `opt.highlight`, `tokens`, `pending`, and `i = 0` are all being evaluated and (in theory) `0` is bing returned.  Intuitively (without testing) it appears that `highlight` is being set to `0` in a very convoluted way.  Maybe there's more to it.

joshbruce:
@intcreator: On the diff, don't know if this is what you're referring to exactly, but the moving `marked()` to the top causes the entire file to considered new according to the diff...basically a change to the whole file. I'm trying to make each commit a refactor; so, it might be easier to read one commit at a time.

lib/marked.js
joshbruce: Concur with the assessment of the commenter. Wonder if we can add that to the lint (@UziTech)? Don't do this. :)

lib/marked.js
joshbruce: ES5 is gonna probably start to annoy me soon. :)

lib/marked.js
UziTech: I agree, I hate this style, but that is how chjj did it. I prefer initializing the variables as close to when you use them as possible.

In this case since the `=` operator has a [higher presedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table) than the `,` operator the expression is essentially:

```js
var (highlight = opt.highlight),
    (tokens),
    (pending),
    (i = 0);
```

joshbruce: To be fair, I don't know who necessarily did it that way, just that I think it should be changed. Totally agree that putting things as close to their contextual relevance is kind of a big deal.

joshbruce: There actually seems to be a lot of "clever" decisions that were made, which may not actually bring the desired ROI compared to writing human readable code. We spend way more time reading than we do writing, you know?

lib/marked.js
intcreator: Perhaps we can find a transpiler for 1.0.

lib/marked.js
```diff
@@ -7,8 +7,223 @@
...
+      return err
+        ? callback(err)
+        : callback(null, out);
+    };
+
+    // Highlight option check, not set or less than three bail
+    //
+    // TODO: Don't some highlighter plugins allow for `js`, effectively being less than three
+    //       despite wanting to have the highlighter run??
+    if (!highlight || highlight.length < 3) {
```
intcreator: I think there are better tests than the length of the highlight.  Agree with the possibility of `js`.  Will we be supporting syntax highlighting of languages such as `c` or `r`?  Highlighting doesn't appear to be in the CommonMark spec.

joshbruce: Agreed. See #1043

lib/marked.js
```diff
@@ -7,8 +7,223 @@
...
+          }
+          token.text = code;
+          token.escaped = true;
+          --pending || done();
+        });
+      })(tokens[i]);
+    }
+
+    return;
+  }
```
intcreator: This function is almost 100 lines long.  I highly recommend we break it down into pieces.

joshbruce: Tagging #1043 so UziTech can get a better feel on internal pull for the updates to the test harness request. No pressure. Just informational purposes.

lib/marked.js
```diff
@@ -102,73 +318,246 @@ block.tables = merge({}, block.gfm, {
...
+  .getRegex();
 
-Lexer.rules = block;
+inline.normal = merge({}, inline);
 
 /**
- * Static Lex Method
+ * Override normal grammar rules to follow original markdown.pl
+ *
+ * TODO: Place this in a Gruber object??
```
intcreator: Gruber object?

lib/marked.js
joshbruce: Meant #1103

lib/marked.js
```diff
@@ -7,8 +7,223 @@
...
+}
+
+/**
+ * TODO: Not sure how to test
+ *
+ * @param  src String
+ * @return boolean
+ */
+marked.receivedOnlySource = function(src, opt, callback) {
+  return (marked.receivedCallback(opt, callback));
```
joshbruce: Should be return `!(marked.receivedCallback(opt, callback));` only source means we did *not* receive those other things.

lib/marked.js
```diff
@@ -454,111 +843,221 @@ Lexer.prototype.token = function(src, top) {
...
+/**
+ * Parsing & Compiling
+ */
 
-inline.reflink = edit(inline.reflink)
-  .replace('inside', inline._inside)
-  .getRegex();
+function Parser(options) {
+  this.tokens = [];
+  this.token = null;
```
joshbruce: Is having it be `null` necessary? Given that this becomes and object with members might be easier for a new person to understand if they saw something like this (or make a Token object):

```js
{
  type: '',
  ...
}
```

Was going to make a more complete example; however, it appears the Token is just a random set of members...like an undefined type. You can see it in the `switch` inside `Parser.prototype.tok()` ~ line 930

joshbruce:
> different things in different files

Strictly speaking, is there anything about ES5 (or some other gotcha) that stops us from separating `marked.js` *before* 1.0 or later?

lib/marked.js
joshbruce: Maybe something more akin to this from traditional OO languages...using Swift highlighting, but not the syntax, strictly speaking.

```swift
abstract class Token {}

class Heading {}

// then the switch could be something like
switch (typeof this.token) {
  case 'Heading':
...
```

But, if the `type` member is the only thing in common between all the different paths for Token...then there's really no reason for the `type` member or Token. ??

joshbruce: I wonder if there are more "hidden" objects inside of here.

lib/marked.js
```diff
@@ -454,111 +843,221 @@ Lexer.prototype.token = function(src, top) {
...
+  if (this.options.gfm) {
+    if (this.options.breaks) {
+      this.rules = inline.breaks;
+    } else {
+      this.rules = inline.gfm;
+    }
+  } else if (this.options.pedantic) {
+    this.rules = inline.pedantic;
+  }
+}
```
joshbruce: The Renderer is just below here, it's essentially a string concatenation object...there's a separate one for both inline and block elements...think it could actually be much smaller.

When we get there, I'll probably want to work on this one...leverage the work from [here](https://github.com/8fold/php-html-component) and [here](https://github.com/8fold/php-html). More than former than the latter since we need to accommodate client-side and most likely we are the extension of someone else's package, being used in, yet again, someone else's project.

We really should try to make Marked compile to something tiny - that's probably why some of the "clever" decisions were made, in fact. Not sure if the project started prior to the prevalence of minify for JS.

joshbruce:
It seems like what Marked is trying to do is sound, but it never got the chance iterate this far...could be totally wrong here, but looking at the code, it's starting to feel familiar based on other projects:

```bash
M⬇ --- parse ---> {} --- render ---> HTML
```

As we consider the future refactoring and unit tests, might be worth keeping the following HTML element type definition in mind:

```js
{
  // h1, h2, etc.
  elem: string,

  // the attribute list, which is usually a set of key-value pairs JSON, in JS,
  // where you can easily convert keys to strings to generate the desired `key=value` string output. 
  attr: array, 

  // this gets tricky because some HTML elements can only take a string; or, there's the child element
  // problem (what marked seems to use "inline" for). If we create a text object that only holds a string
  // we can always pass in an array of element objects (what we seem to call tokens at the moment).
  inner: string|array,

  // this phrasing comes from the HTML5 spec when looking at the definition of an individual element,
  // the first time I created this type definition I called `isSelfClosing` but that proved confusing.
  omitEndTag: boolean
}
```

This is almost the model used by React, I think, but not explicitly just through their `createElement` method: https://reactjs.org/docs/react-api.html#createelement

Maybe we need an Element object?? 

Then the parser can just keep creating those. Various Markdown flavors are also only able to use so many of the global attributes afforded by HTML5 that we might be able to do some optimization there as well.

joshbruce:
Closing as edification was achieved, conflicts, and will most likely happen over a series of PRs not just one.

PR #1110, opened 2018-03-02
ESLint ignore node_modules

joshbruce:
**Marked version:** 0.3.17

## Description

Our builds keep failing because of lint...??

From what I understand chatting with @UziTech, the linter is running against `node_modules` as well. 

The ESLint docs discuss an `.eslintignore` file. However, the sample has a comment that says `node_modules` is ignored by default. Confused.

This PR makes the ignore explicit but I'm not sure that will actually solve the issue and is necessary.

Leaving the "build" ignores in just for completeness with the example.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,7 @@
+# /node_modules/* and /bower_components/* ignored by default
+/node_modules/*
+/bower_components/*
+
+# Ignore built files except build/index.js
+build/*
+!build/index.js


UziTech:
> when indent_style is set to tab, it may be desirable to leave indent_size unspecified so readers may view the file using their preferred indentation width

from http://editorconfig.org/#file-format-details

.eslintignore
```diff
@@ -0,0 +1,7 @@
+# /node_modules/* and /bower_components/* ignored by default
```
UziTech: /node_modules/* is ignored by default so I don't think this is the issue.

UziTech:
Looks like the issue is that eslint doesn't support node 0.10

I'll set up travis build stages so it only check eslint on the latest version of node

PR #1104, opened 2018-03-01
Testing approach concept

joshbruce:
**Marked version:** 0.3.17

## Description

http://spec.commonmark.org/0.28/#example-12

Been thinking about priorities and testing, given the analysis of our current test bed. Further, there are a lot of issues that are more spec-related that may or may not include the Markdown and expected HTML output...

Anyway, long story short and we can expound in the comments.

Thinking we should possibly stick to our guns (or appear to adjust the) [priorities](https://github.com/markedjs/marked/blob/master/CONTRIBUTING.md#priorities).

We don't have any known security issues. 

We know we have a lot of issues submitted by users. Most of those seem related to not being compliant with the spec, which is a higher priority than user-specific things. 

We also have the base for creating the simplest complete spec-compliant test bed ever...even someone who has no idea how marked works could do this.

1. Go to the tests.
2. Look for a flavor and example (cm_example_12, in this case).
3. If it's there, that gives users confirmation and trust that we have that covered.
4. If it's not there:
  * Create an md and HTML file in `/new` (would be really nice to be able to put tests in a folder named after the flavor `/commonmark`, for example).
  * Literally copy and paste.
  * Run `npm test`
5. Submit the PR.

Or, if it's broken and you know how to fix it, fix it. Otherwise, we have the PR available for someone else to come in and pick it up.

I think if we can confirm spec compliance with the testing harness we have, might be easier for us to close them out.

Maybe we can get random people to help out...none of us need to know *how* marked works to do this.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regresstion (if no tests added, list tests covering this PR); or,
- [ ] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] Draft GitHub release notes have been updated.
- [ ] CI is green (no forced merge required).
- [ ] Merge PR

Issue #1103, opened 2018-02-28
Update testing harness to allow for unit, integration, and application tests


joshbruce:
**Marked version:** https://github.com/markedjs/marked/pull/1102/commits/fb28a9538e27c5c97988bf2423cd257def243a60

**Proposal type:** other

Sorry, one more, then I'll be done...sometimes my brain latches on to things and I can't do anything else until I get it done (think @chjj has that type o' thing going on as well 😊).

## What pain point are you perceiving?

I typically bucket test types into three major areas:

1. Unit: Object is able to be tested in complete (or darn close) isolation from other objects and environments...at the public API level (individual public properties and methods).
2. Integration: Objects can be tested to make sure they play well with each other.
3. Application: The entire application (all objects) are leveraged to test things.

I know there's like 500 different labels for testing types, which is why I tend to just bucket them this way.

I think `marked.js` has a lot of large functions and methods that can be broken down using the extract method refactoring. However, our current test harness only seems to be able to run what I would consider application tests (some might even call it end-to-end).

Markdown + marked() + HTML + are they equal? pass/fail

It would be nice if I could test each method on the marked object though. Really beat the heck out of that one method to make sure it's clean.

## What solution are you suggesting?

1. Add ability to have unit and integration style testing (think we have application testing pretty well covered).
2. Have ability to run the one or two types of tests separately (being able to run a single test would be awesome, but not required, our tests run pretty quickly right now); maybe something like `npm test:unit` or something.
3. Add to Travis checks.
4. Maybe update CONTRIBUTING with the test types and instructions (technically, I think that would be part of the checklist but maybe we want to test it out first - make sure it has the ROI I think it will).

Tagging @UziTech as Dr. DevOps and @davisjam as a Node Nerd (new title?? I think you're a NodeJS cat, yeah??)

joshbruce:
Oh! And @UziTech, here's the Angular 5 Circle CI config thing I mentioned somewhere when we were got on the subject of Travis supporting separate builds and things...not sure if it helps with Travis: https://github.com/joshbruce/sam-design-system-site/blob/master/.circleci/config.yml

But, think you're on to something with that...though not sure exactly what you were thinking - maybe separate out the lint checks and the test tests?? Maybe have a build for each one - first to fail - fails the PR?? Anyway, yeah that's the config and project.

joshbruce:
Just want to make the implicit explicit...sorry for the lack of emoticon and intent behind the comment:

> Node Nerd

[...]

PR #1111, opened 2018-03-03
[editorconfig]: All md files except in test use tab

joshbruce:
**Marked version:** 0.3.17

## Description

tl;dr: PR changes editorconfig to use tabs not spaces while still ignoring MD files in test directory.

Looking at the CommonMark specification and considering the future regarding #1104 and https://github.com/markedjs/marked/issues/1103#issuecomment-369957173 a lot of the examples use honest tabs...not spaces in Markdown...makes sense given the nature of plain text editors (this was why @davisjam's initial contributions didn't pass lint - he uses VIM, which doesn't do spaced tabs).

So, not too sure what to do here regarding ignoring that `/test` folder. Or maybe ignoring markdown files all together like Federico had mentioned in the initial PR #1082 - think I forgot to put it back into WIP status or something and we hadn't really talked more about the GH Pages thing yet.

Plus, that's why we call it software and not hardware - can be easy to change in the future, if we do it right.

Thoughts?

Tagging @intcreator as well, as it's something to consider as we move toward creating all the tests ever.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,5 +13,5 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 translate_tabs_to_spaces = true
-indent_style = space
+indent_style = tab
 indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,5 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
-translate_tabs_to_spaces = true
 indent_style = tab
 indent_size = 4


.editorconfig
```diff
@@ -13,5 +13,5 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 translate_tabs_to_spaces = true
-indent_style = space
+indent_style = tab
 indent_size = 4
```
UziTech: > when indent_style is set to tab, it may be desirable to leave indent_size unspecified so readers may view the file using their preferred indentation width

from http://editorconfig.org/#file-format-details

.editorconfig
```diff
@@ -13,5 +13,5 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 translate_tabs_to_spaces = true
```
UziTech: What is this rule? it doesn't show on up https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties

joshbruce: May be either a deprecated feature - or a Sublime Text specific thing (see first sentence after code): https://github.com/sindresorhus/editorconfig-sublime/issues/12

Automatically converts tabs to spaces on new files. Probably a horrible idea all around anyway. Good catch. ;)

Removed.

.editorconfig
```diff
@@ -12,6 +12,5 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
```
UziTech: kind of funny that this file doesn't have a new line at the end 😆

.editorconfig
joshbruce: Ran into issues with that, I think - most likely because of the asterisk. *all* files have an indent size of 2...need to override to say something else. We could remove the asterisk, I suppose. Basically saying we only want these specific file types to have these specific qualities. I tried doing something like this, and it didn't seem to work very well - but it could just be my editor + plug-in interaction (tested in both Sublime and VSC, with same unexpected results):

```
[*]
charset = utf-8
end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true

[*.json, *.js]
indent_style = space
indent_size = 2

[*.md, !test/*md]
indent_style = tab
```

For some reason, when I tried various permutations of this inheritance stack I always got spaces in MD files...?? Could just be something in my environment, though it's not obvious to me what.

.editorconfig
joshbruce: Dude! I thought that was hilarious. I was literally expecting it to happen as soon as I hit safe...so disappointed. 🤣

.editorconfig
joshbruce: At one point, I was also tempted to just do this:

```
[*.json, *.js]
charset = utf-8
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 2
```

And call it a day...literally only effect JS and JSON. Not sure what would have things like YML though. Also not sure how broad our lint rules go either. Would like to get this to match those as much as possible where possible - just not a deal-breaker if they don't or can't.

joshbruce:
Does eslint only do JS? (Revealing my over-dependance `editorconfig` right there.)

If so, could literally get rid of the other rules and just apply them to JS files. The objective is keep developers from failing lint, if their editor can apply the config. Give them the check with the `NPM lint:check` by @UziTech and a quick way to fix by Federico. And a way to protect the core by indicating with Travis. At that point, we copy-paste from the docs and point them there. Eventually people see it enough and should reduce how often it happens. :)

joshbruce:
es = EcmaScript

I'm retarded sometimes... https://youtu.be/j9ndS6yu3Kk

UziTech:
by default eslint only lints .js files. You can specify a [`--ext`](https://eslint.org/docs/user-guide/configuring#specifying-file-extensions-to-lint) parameter to add other extensions. `.json` is a pretty common one that is added

PR #1113, opened 2018-03-03
travis build stages

UziTech:
**Marked version:** 76d75e19c7717e954c91a4a4c4c6499aa5f7b20a

## Description

Adds [build stages](https://docs.travis-ci.com/user/build-stages/) to travis-ci to make it easier to see what failed.

image

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,17 @@
 language: node_js
+
 node_js:
-  - "0.10"
-  - "4"
-  - "lts/*"
-  - "node"
-script: |
-  if [ `node --version | cut -d . -f 1,2` = "v0.10" ]; then
-    sed -i s/0o755/0755/ test/index.js;
-    npm test;
-  else
-    npm run test:lint && npm test;
-  fi
+  - v0.10
+  - v4
+  - lts/*
+  - node
+
+jobs:
+  include:
+    - stage: lint
+      node_js: lts/*
+      script: npm run test:lint
+
 cache:
   directories:
     - node_modules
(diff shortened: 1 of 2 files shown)


joshbruce:
I really like, and am assuming here, that lint comes last; so, we get the really really important sorted first, yeah?

.travis.yml
```diff
@@ -1,16 +1,17 @@
...
-    sed -i s/0o755/0755/ test/index.js;
-    npm test;
-  else
-    npm run test:lint && npm test;
-  fi
+
+jobs:
+  include:
+    - stage: spec tests 👩🏽‍💻
+      node_js: v0.10
```
styfle: It looks like this ran properly but why is the first `node_js` line missing the `-` hyphen prefix?

UziTech: that is the same as json:

```json
[
      {
        "stage": "spec tests 👩🏽‍💻",
        "node_js": "v0.10"
      },
      {
        "node_js": "v4"
      }
]
```

the hyphen is the start of an array item.

UziTech:
@joshbruce The jobs in each stage run in parallel but each stage runs in series so it will only run Lint if all Test jobs pass.

I was thinking it would be good to add stages for unit and integration tests after they are added.

We could also do a minify stage that would minify and commit after every master branch commit

joshbruce:
> We could also do a minify stage that would minify and commit after every master branch commit

Could be a good thing. If we do that, we should make a note to update the RELEASING docs.

joshbruce:
#1119

UziTech:
this ☝️ should be on #1132

PR #1121, opened 2018-03-04
[WIP] Jasmine 3.1.0

### LINKED ISSUE #1103: Update testing harness to allow for unit, integration, and application tests

UziTech:
**Marked version:** latest

## Description

Adds Jasmine for unit and integration testing

### TODO:
- [ ] Actually write unit and integration tests
- [ ] Any other types of test?
- [ ] Document new test types?
- [ ] code coverage?
- [ ] Use jasmine to show errors with spec tests instead of custom test harness.
- [ ] Add Unit and Integration build stages on travis

Fixes #1103

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,6 +22,7 @@
   "env": {
     "node": true,
     "browser": true,
-    "amd": true
+    "amd": true,
+    "jasmine": true
   }
 }
--- a/jasmine.json
+++ b/jasmine.json
@@ -1,0 +1,11 @@
+{
+  "spec_dir": "test",
+  "spec_files": [
+    "**/*-spec.js"
+  ],
+  "helpers": [
+    "helpers/**/*.js"
+  ],
+  "stopSpecOnExpectationFailure": false,
+  "random": true
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -857,6 +857,20 @@
       "integrity": "sha1-jpQ9E1jcN1VQVOy+LtsFqhdO3hQ=",
       "dev": true
     },
+    "glob": {
+      "version": "7.1.2",
+      "resolved": "https://registry.npmjs.org/glob/-/glob-7.1.2.tgz",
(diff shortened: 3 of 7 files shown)


davisjam:
> Are we planning to convert all tests to Jasmine (even the MD + HTML ones)

Certainly all the things tested by the MD-HTML tests should appear in "unit test"-like form. The existing MD-HTML tests are good integration/whatever-you-call-it tests.

davisjam:
Ah, I think I may have been slightly hasty in merging this.

I think having the new Jasmine infrastructure in place is good so that more than just @UziTech can contribute.

However, is Travis configured to run both the old tests and the new tests during the interim? I'm not familiar with Travis configurations so I can't tell.

PR #1129, opened 2018-03-04
WIP: test: jasmine-ify original tests

davisjam:
Jasmine-ifying original tests.

So far:

amsps_and_angles_encoding
auto_links
links_inline_style
inline_html_simple

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "uglify-js": "^3.3.10"
   },
   "scripts": {
-    "test": "jasmine --config=jasmine.json",
+    "test": "node test && jasmine --config=jasmine.json",
     "test:unit": "npm test -- test/unit/**/*-spec.js",
     "test:specs": "npm test -- test/specs/**/*-spec.js",
     "test:integration": "npm test -- test/integration/**/*-spec.js",
--- a/test/integration/marked-spec.js
+++ b/test/integration/marked-spec.js
@@ -1,5 +1,62 @@
 var marked = require('../../marked.min.js');
 
-it('should run the test', function () {
-  expect(marked('Hello World!')).toBe('<p>Hello World!</p>\n');
+describe('integration suite', function () {
+  it('should put plaintext into a paragraph', function () {
+    expect(marked('Hello World!')).toBe('<p>Hello World!</p>\n');
+  });
+
+  describe('links inline style', function () {
+
+    it('should handle regular links', function () {
+      var descs = [{'text': 'simpleString',
+                    'link': 'simpleString'},
+                   {'text': 'string with spaces',
+                    'link': 'http://github.com'},
+                   {'text': 'string with \[backslash \] brackets',
+                    'link': 'http://github.com'},
+                   {'text': 'string with [inline brackets]',
+                    'link': 'http://github.com'},
+                   {'text': 'url has space',
+                    'link': '/url has space'},
+                   {'text': 'empty',
+                    'link': ''},
+                  ];
+      descs.forEach(function(desc) {
+        expect(marked(`[${desc.text}](${desc.link})`)).toBe(`<p><a href="${desc.link}">${desc.text}</a></p>\n`);
+      });
+    });
+
+    it('should handle links with titles', function () {
+      var descs = [{'text': 'simpleString',
+                    'link': 'simpleString',
+                    'title': 'simpleTitle',
+                    'beforeTitle': ' ',
+                    'afterTitle': ''},
+                   {'text': 'url and title',
+                    'link': 'simpleString',
+                    'title': 'title preceded by two spaces',
+                    'beforeTitle': '  ',
+                    'afterTitle': ''},
+                   {'text': 'url and title',
+                    'link': 'simpleString',
+                    'title': 'title preceded by a tab',
+                    'beforeTitle': '	',
+                    'afterTitle': ''},
+                   {'text': 'url and title',
+                    'link': 'simpleString',
+                    'title': 'title has spaces afterward',
+                    'beforeTitle': ' ',
+                    'afterTitle': '  '},
+                   {'text': 'url and title',
+                    'link': '/url/has space/',
+                    'title': 'url has space and title',
+                    'beforeTitle': ' ',
+                    'afterTitle': ''},
+                  ];
+      descs.forEach(function(desc) {
+        expect(marked(`[${desc.text}](${desc.link}${desc.beforeTitle}"${desc.title}"${desc.afterTitle})`)).toBe(`<p><a href="${desc.link}" title="${desc.title}">${desc.text}</a></p>\n`);
+      });
+    });
+
+  });
 });
--- a/test/integration/marked-spec.js
+++ b/test/integration/marked-spec.js
@@ -25,7 +25,7 @@ describe('integration suite', function () {
       });
 
     });
-  });
+  }); // inline html
 
   describe('links inline style', function () {
 
@@ -53,7 +53,7 @@ describe('integration suite', function () {
       });
     });
 
-  });
+  }); // links inline style
 
   describe('auto links', function () {
     it('should handle auto-linking', function() {
@@ -101,5 +101,5 @@ describe('integration suite', function () {
       });
     });
 
-  });
+  }); // auto links
 });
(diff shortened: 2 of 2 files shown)


test/integration/marked-spec.js
```diff
@@ -1,5 +1,62 @@
 var marked = require('../../marked.min.js');
 
-it('should run the test', function () {
-  expect(marked('Hello World!')).toBe('<p>Hello World!</p>\n');
+describe('integration suite', function () {
+  it('should put plaintext into a paragraph', function () {
+    expect(marked('Hello World!')).toBe('<p>Hello World!</p>\n');
+  });
```
UziTech: We don't need to keep this test if we don't want to. I just added it as an example

davisjam: Eventually it'll go under a describe('plaintext') or something.

davisjam:
a693f23: implements suggestion from [1121](https://github.com/markedjs/marked/pull/1121#issuecomment-370260099). `npm test` will now run the old tests first.

davisjam:
@UziTech 

So these tests will be prettier if we can write code like this:
```js
expect(marked(`[${desc.text}](${desc.link})`)).toBe(`<p><a href="${des
```

which is not "old node"-compliant (hence the travis failure).

Thoughts?

UziTech:
Ya we need to stick to es5 in the tests until we figure out a build step or drop node <6

UziTech:
I think the idea is v2.0 we will support only evergreen browsers and node >=6. At that time we will switch everything to ES6+

joshbruce:
@davisjam: Can't really read that. Newb-light is target.

davisjam:
@joshbruce Acknowledged.

joshbruce:
@davisjam: Try one test per example??

See also: https://github.com/markedjs/marked/pull/1124#issuecomment-370240670

davisjam:
@joshbruce Hesitant to do that since there would be a ton of boilerplate. Thoughts?

joshbruce:
@davisjam: I might need to slow down a bit in response time...really jazzed to see all the cooperation and collaboration...y'all are awesome!

...question:

Define "boilerplate" in this context?

Considerations (imho):

1. Each test should be discreet (seem to remember there a mnemonic here STAR?? I think, COMET...I don't know): `it('should pass CM 123'` - this should help isolate which specific test scenario is failing without being the general heading "inline elements" and should be easier to migrate should the spec ever change...added new example -> add new test...changed specific example -> update.
2. If we get an issue with a sample of MD and HTML - we can write `it('should pass issue 1234` - to protect against regression and make triaging fixes easier. 
3. The size of the test bed shouldn't matter, if that's the concern related to boilerplate. It's mainly there to act as a technical specification for users to see how to use marked and get an idea of what to expect. If the test is in there, we say we can do it; if it's not...well. (Maybe we would do something like `it('should fail CM 789'` to let people know, "We don't support that example" (see the emphasis problem). 

Having said that, I'm not a big fan of writing [tests for their own sake](https://github.com/8fold/php-html-component/blob/master/tests/ComponentTest.php)...the fact we're doing this against a spec makes it a little more awkward than what I was doing there at the moment...partially, I think because our renderer is slightly complex compared to the type definition of an HTML element per https://github.com/markedjs/marked/pull/1102#issuecomment-370098125 - which is what I ended up with after starting with a PHP associative array (JSON). It makes the NPM package size bigger, but that's not what people are using operationally in their sites...if I have a question on why something is happening; instead of running to GitHub, I can run to the tests or spec (make sure I'm suing the right combination, make sure it's supported, then maybe post a PR to make the test pass).

Something like that...??

davisjam:
> Each test should be discreet

Yes. I'm writing tests at the level of "marked honors inline html" or "marked honors inline links". Spec-by-spec.

davisjam:
Okie dokie, I'll leave this for now.

@joshbruce @UziTech Let me know how the current version looks.

davisjam:
Currently 'npm run lint' does not scan the new tests.

```
"lint": "eslint --fix lib/marked.js test/index.js",
```

Can we just replace "test/index.js" with "test/"? Seems to work.

UziTech:
@davisjam really we should add an `.eslintignore` file for `marked.min,js` (I'm pretty sure that is the only files we want to ignore) and use `eslint --fix .`

package.json
```diff
@@ -38,7 +38,7 @@
     "uglify-js": "^3.3.10"
   },
   "scripts": {
-    "test": "jasmine --config=jasmine.json",
+    "test": "node test && jasmine --config=jasmine.json",
```
UziTech: `test/specs/specs-spec.js` already takes care of running the `test/index.js` tests in jasmine so there is no need to run `node test`.

test/integration/marked-spec.js
```diff
@@ -1,5 +1,133 @@
...
+      var samples = [{'md': 'AT&T has an ampersand in their name', 'html': '<p>AT&amp;T has an ampersand in their name</p>\n'},
+                     {'md': 'AT&amp;T is another way to write it', 'html': '<p>AT&amp;T is another way to write it</p>\n'},
+                     {'md': 'This & that', 'html': '<p>This &amp; that</p>\n'},
+                     {'md': 'Inline [link](/script?foo=1&bar=2) with ampersands', 'html': '<p>Inline <a href="/script?foo=1&amp;bar=2">link</a> with ampersands</p>\n'},
+                     {'md': '[link with ampersands](http://example.com/?foo=1&bar=)', 'html': '<p><a href="http://example.com/?foo=1&amp;bar=">link with ampersands</a></p>\n'},
+                     {'md': '[link title with ampersands](http://att.com "AT&T")', 'html': '<p><a href="http://att.com" title="AT&amp;T">link title with ampersands</a></p>\n'},
+                    ];
+
+      samples.forEach(function(sample) {
+        expect(marked(sample.md)).toBe(sample.html);
```
UziTech: We should figure out how to get a better error message other than `expected 'some really long string' to equal 'some other really long string'`.

We could try something like [jasmine-diff-matchers](https://www.npmjs.com/package/jasmine-diff-matchers)

package.json
UziTech: `test/specs/specs-spec.js` already takes care of running the `test/index.js` tests in jasmine so there is no need to run `node test`.

test/integration/marked-spec.js
```diff
@@ -1,5 +1,133 @@
...
+
+  describe('special characters', function () {
+    it('should handle ampersands', function () {
+      var samples = [{'md': 'AT&T has an ampersand in their name', 'html': '<p>AT&amp;T has an ampersand in their name</p>\n'},
+                     {'md': 'AT&amp;T is another way to write it', 'html': '<p>AT&amp;T is another way to write it</p>\n'},
+                     {'md': 'This & that', 'html': '<p>This &amp; that</p>\n'},
+                     {'md': 'Inline [link](/script?foo=1&bar=2) with ampersands', 'html': '<p>Inline <a href="/script?foo=1&amp;bar=2">link</a> with ampersands</p>\n'},
+                     {'md': '[link with ampersands](http://example.com/?foo=1&bar=)', 'html': '<p><a href="http://example.com/?foo=1&amp;bar=">link with ampersands</a></p>\n'},
+                     {'md': '[link title with ampersands](http://att.com "AT&T")', 'html': '<p><a href="http://att.com" title="AT&amp;T">link title with ampersands</a></p>\n'},
+                    ];
```
UziTech: I think it would be easier to just pull these from their current files. I hate hard coding strings in javascript. Too much possibility for errors.

We could construct each test from the files and add the `describe` and `it` descriptions to the front-matter in the .md file

UziTech: or maybe reorganize these tests and pull the `describe` description from the folder name

joshbruce:
See #1159

UziTech:
I think we should try to converge on a similar test pattern as #1160, adding a json file for the original test similar to [commonmark.json](https://github.com/8fold/marked/blob/ac80808612219127bfb9ecef67e1e32a9f4b0311/test/integration/commonmark.json)

joshbruce:
I think if we're going to Jasminify the legacy tests, we might be covered elsewhere. #1160, for example, should/would cover the CM tests; so, they can be removed. The GFM ones would be covered by the GFM tests that haven't been created yet. There may also be conflicts between the specs and what marked does (`autolinks`, for example, might include a mechanism not covered by CommonMark).

UziTech:
I like the idea of just having CommonMark and GFM tests and deprecating all other features in favor of making marked more extensible and less bloated.

We should probably keep the tests around until the options are removed (in 1.x or 2.x)

UziTech:
I'm closing this in favor of #1444

PR #1132, opened 2018-03-05
More build stages

UziTech:
**Marked version:** latest

## Description

This adds separate build stages for unit, spec, and integration tests as well as a minify build step which is only run when the master branch is pushed to.

The minify build step requires a github token be stored on travis (this should be done securely) and only given `public_repo` access

I'm not sure what we want to do with the different node versions. I don't think we need to run all tests against all node versions.

Later we can add a browser test stage.

image

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,12 @@ language: node_js
 
 jobs:
   include:
-    - stage: spec tests 👩🏽‍💻
-      script: npm run test:specs
-      node_js: lts/*
-
     - stage: unit tests 👩🏽‍💻
       script: npm run test:unit
       node_js: lts/*
 
-    - stage: integration tests 👩🏽‍💻
-      script: npm run test:integration
+    - stage: spec tests 👩🏽‍💻
+      script: npm run test:specs
       node_js: v0.10
     - node_js: v4
     - node_js: lts/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ jobs:
           git config --global user.name "Travis-CI"
           git config credential.helper "store --file=.git/credentials"
           echo "https://${GITHUB_TOKEN}:@github.com" > .git/credentials
-          git commit -am 'minify [skip ci]'
+          git commit -am '🗜️ minify [skip ci]'
           git push origin HEAD:${TRAVIS_BRANCH}
         fi
       node_js: lts/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,20 @@ jobs:
       script: npm run test:lint
       node_js: lts/*
 
+    - stage: minify 🗜️
+      script: |
+        npm run build
+        if ! git diff --quiet; then
+          git config --global user.email "travis@travis-ci.org"
+          git config --global user.name "Travis-CI"
+          git config credential.helper "store --file=.git/credentials"
+          echo "https://${GITHUB_TOKEN}:@github.com" > .git/credentials
+          git commit -am 'minify [skip ci]'
+          git push origin HEAD:${TRAVIS_BRANCH}
+        fi
+      node_js: lts/*
+      if: branch = master AND type = push
+
 cache:
   directories:
     - node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,23 @@ language: node_js
 jobs:
   include:
     - stage: spec tests 👩🏽‍💻
+      script: npm run test:specs
+      node_js: lts/*
+
+    - stage: unit tests 👩🏽‍💻
+      script: npm run test:unit
+      node_js: lts/*
+
+    - stage: integration tests 👩🏽‍💻
+      script: npm run test:integration
       node_js: v0.10
     - node_js: v4
     - node_js: lts/*
     - node_js: node
 
     - stage: lint ✨
-      node_js: lts/*
       script: npm run test:lint
+      node_js: lts/*
 
 cache:
   directories:


davisjam:
> I don't think we need to run all tests against all node versions.

I think we should run tests against any node versions we support. For example, if `const/let` are put into the source, that will break backwards compatibility and we would like to know.

UziTech:
I agree, but I don't think we need to run EVERY test against EVERY node version we support.

In this PR I have spec and unit tests only running against node lts and integration tests running against node v0.10, v4, lts, and latest

UziTech:
@styfle `marked.min.js` should never be manually edited so there should never be merge conflicts and `marked.min.js` should always be in sync with `marked.js`

UziTech:
also npm is not the only package manager that marked is in. So having `marked.min.js`  only as part of the npm publish would not be a good idea

styfle:
> should never be manually edited so there should never be merge conflicts

Generated code can still have merge conflicts.
This is why outputs/binaries are rarely committed to source control.

> npm is not the only package manager that marked is in

Do we have the other package managers listed anywhere? Who is publishing to them?

UziTech:
I agree the min file should just be in the release but since it has been part of the repo for a long time i'm guessing it will break something if we remove it now. Maybe in v2.0 we should remove it.

> Do we have the other package managers listed anywhere?

I'm not sure if they are listed anywhere but here are the ones I know of:
  - [npm](https://www.npmjs.com/package/marked)
  - [cdnjs](https://cdnjs.com/libraries/marked)
  - [bower](https://bower.io/search/?q=marked)
  - [jsdelivr](https://www.jsdelivr.com/package/npm/marked)
  - [unpkg](http://unpkg.org/pkg/marked/)

styfle:
I know that `npm`, `jsdelivr`, `unpkg` all read from npm so we don't have to worry, its publish once and they all get updated.

Is `bower` still around? I was thinking they were encouraging users to [migrate away from bower](https://bower.io/blog/2017/how-to-migrate-away-from-bower/).

I have no idea how to publish to `cdnjs`, it seems its completely manual.

styfle:
> Maybe in v2.0 we should remove it.

Thats fine

UziTech:
> Is `bower` still around?

They are recommending people move to yarn (which also uses npm registry) but I'm sure there will still be people using them for a while.

joshbruce:
@UziTech: If this doesn't break anything I'm fine merging it now. Just let me know.

UziTech:
I removed the "spec tests" and renamed "integration tests" to "spec tests"

In order for the minify step to work travis will need a `GITHUB_TOKEN` environment variable

image

The token only needs the "public_repo" permissions

image

@joshbruce I'm not sure if you want to create a personal access token from your account otherwise I have seen organizations create a dummy account just for that.

UziTech:
Here is a test travis build showing the output for the minify job:
https://travis-ci.org/UziTech/marked/builds/349322194#L421

The GITHUB_TOKEN variable won't be leaked in the build output
image

styfle:
I'm still going to voice my opinion that travis should not be committing code.
This will very likely cause merge conflicts for every PR because the minified file is a single line.

UziTech:
The minify job is only run when code is pushed to markedjs/master so marked.min.js isn't changed in PRs (and should never be manually changed in PRs)

UziTech:
If this does cause problems, it would be trivial to remove the job and go back to the way we are updating `marked.min.js` now. So including the job now can only improve things.

joshbruce:
@UziTech: Can you hook me up with instructions for the keys you need (it's not something I do on the regular pretty sure it's not the SSH ones that we need here). I'll decide what to do about accounts after that.

UziTech:
1. goto https://github.com/settings/tokens
2. click <kbd>Generate new token</kbd>
3. Enter a "Token description"
4. Check "public_repo"
5. Click <kbd>Generate token</kbd>
6. Copy the token
image
7. Goto https://travis-ci.org/markedjs/marked/settings
8. Enter the token under "Environment Variables"
image
9. Click <kbd>Add</kbd>
10. Do what comes before Part-B 🎉

joshbruce:
@UziTech: Think I got it right. I created an account using the markedjs.org domain. It's just a bot with a forward to me at present; so, we'll figure stuff out as we progress.

UziTech:
Merge this and see what happens 😁

UziTech:
@joshbruce I think you have to add markedjsbot to the markedjs organization and git him push permissions

UziTech:
the error given is:

```
Permission to markedjs/marked.git denied to markedjsbot.
```

so it seems like the code worked, markedjsbot just doesn't have permission to push code to markedjs/marked

joshbruce:
Doh! And done.

UziTech:
Still getting the same error 😕 

Try making sure the permissions are correct then go to [travis](https://travis-ci.org/markedjs/marked) and click retry on the minify job

image

joshbruce:
Think `markdjsbot` had to be listed as an owner (not just a member of the org)??

styfle:
The minify step stopped working after I changed the settings for requiring approvers.

image


Which one do we think is more important:
Auto commit minify step or 2 required approvers?

joshbruce:
Is there no way to have both. My two cents is that we have ways around either one not being available; so, neither is a dyed in the wool must have. 

For releases having the minification happen on Travis cuts down work for me and minimizes the possibility of being out of sync.

UziTech:
I say the 2 approvers be just an unwritten rule

styfle:
I had to disable all the policies...leaving the `master` branch unprotected in order to allow travis to commit to master.

PR #1083, opened 2018-02-26
security: fix REDOS vulnerabilities

davisjam:
Demonstrate and fix catastrophic backtracking in two regexes.

For #1070.

PR #1135, opened 2018-03-08
[commonmark] make html comments, html inlines, html blocks and links compliant

### LINKED ISSUE #229: Escaped characters are ignored in url's
[Test](http://www.test.org/index.php/XSS_(Style\)_stuff)

See how this renders correctly here. The following results in **&lt;link&gt;Test&lt;/link&gt;_stuff** when parsed by marked.js

### LINKED ISSUE #257: Content of <script> tags is escaped when they are inside html tags and 'sanitize: false'
If for example we have this html in our markdown:

``` html
<div>
    <script>if (5 > 4) {}</script>
</div>
```

then content of `<script>` will be escaped. But content remains unescaped if `<script>` is not surrounded with html:

``` html
<script>if (5 > 4) {}</script>
```

If [here](https://github.com/chjj/marked/blob/master/lib/marked.js#L357) we just set `pre: true` then content of `<script>` within html will not be escaped as expected. Is it a bug or a feature, and how can I force marked.js to not escape content of `<script>` within html?

### LINKED ISSUE #303: a <textarea> parsing bug
Hi, I add a testcase

```
<textarea>
* abc
* 
</textarea>
```

I got

```
<textarea>
<em> abc
<em>
</textarea>
```

but

```
<textarea>
* abc
</textarea>
```

is ok.

### LINKED ISSUE #366: Parse complicated url problem
In github, `[Ajax](http://en.wikipedia.org/wiki/Ajax_(programming))` will be correctly parsed: [Ajax](http://en.wikipedia.org/wiki/Ajax_%28programming%29), but marked will add unnecessary `)`

### LINKED ISSUE #488: Add option to render Markdown inside <div> blocks
Hi,

First: thanks for writing this fantastic library!

My issue: I'm using Markdown in documentation for a project, and would like to use &lt;div> tags inside of the Markdown to specify that certain sections should be hidden (or otherwise stylized).

For example:

```
# Title

This is some text.

<div class="toggle-hide">
## Hidden section

A hidden paragraph.
</div>

This text is visible.
```

Unfortunately, the Markdown content inside the &lt;div> does not get rendered -- it's copied verbatim into the output.

Is there any way to configure marked.js to render Markdown inside of &lt;div> tags? If not: can you point me in the right direction for how to add such an option? Or: can you recommend some alternate strategy for hiding (or toggling visibility of) sections of content? 

Thanks!

### LINKED ISSUE #501: code block containing line break with new line starting with html fails
This fails

```
Additionally, arbitrary scripts can be executed by running `npm run-script
<pkg> <stage>`. Test.
```

giving

```
<p>Additionally, arbitrary scripts can be executed by running `npm run-script</p>
<p><pkg> <stage>`. Test.</p>
```

but this doesn't:

```
Additionally, arbitrary scripts can be executed by running `npm
run-script <pkg> <stage>`. Test.
```

giving

```
<p>Additionally, arbitrary scripts can be executed by running <code>npm
run-script &lt;pkg&gt; &lt;stage&gt;</code>. Test.</p>
```

### LINKED ISSUE #586: Trailing ) dropped in javascript links
Hello,
When a link targets javascript, the trailing ) is dropped.
Example:  `[hello](javascript:alert("Hello"))` links to `javascript:alert("Hello"`
Thanks for your job!
@chtixof

### LINKED ISSUE #619: URLs with parentheses are not converting properly.
Using the following markup:

```
[青 ("ao")](https://en.wikipedia.org/wiki/Ao_(color))
```

leads to a link that actually looks like `青 ("ao"))` with the link's own parenthesis getting cut off (marked seems to be treating the first `)` it sees as link terminator, instead of taking nesting into account)

### LINKED ISSUE #690: Marked does not handle links with parentheses
This is a carbon copy of [this bug](https://github.com/Microsoft/vscode/issues/380#issuecomment-158873568) for _VSCode_.

Here is an example where the preview engine in VSCode mis-renders links with parentheses:
image3

Looks like a Marked issue - any way to fix it?

### LINKED ISSUE #736: Conversion failed link, there are () when
Code:

``` js
var marked = require('marked');

var str = '[window.close()](javascript:window.close())';

console.log(str);
console.log(marked(str));
```

Output:

``` html
[window.close()](javascript:window.close())
<p><a href="javascript:window.close(">window.close()</a>)</p>
```

Right should be:   (My opinion)

``` html
<p><a href="javascript:window.close()">window.close()</a></p>
```

by npm marked@0.3.5, nodejs@4.4.0

---

Thanks ~

### LINKED ISSUE #829: Does not Handle Markdown Escapes In Link Href
From https://github.com/Microsoft/vscode/issues/14968

Marked does not seem to properly handle markdown escapes inside of links. Only the link text itself is properly handled, the `title` and `href` are not:

```js
var marked = require('marked');
marked('[text\\.](http://example\\.com "\\.title")')
```

produces:

```
<p><a href="http://example\.com" title="\.title">text.</a></p>
```

### LINKED ISSUE #1032: HTML following between two comments don't get recognized as HTML
```html
<!-- This is a comment -->
<details>
<summary>
    This is the summary.
</summary>
Hello!

hello!

hello!
</details>
<!-- This is a comment -->
``` 

## Expectation

```html
<!-- This is a comment -->
<details>
<summary>
    This is the summary.
</summary>
Hello!

hello!

hello!
</details>
<!-- This is a comment -->
``` 

## Result

```html
<!-- This is a comment -->
<p><details></p>
<p><summary>
    This is the summary.
</summary>
Hello!</p>
<p>hello!</p>
<p>hello!
</details>
<!-- This is a comment --></p>
```

## What was attempted

As shown above, HTML code between two comments are identified as `paragraph`s instead of `html`. Inserting blank lines between the comments and the HTML code in markdown temporarily solved the rendering issue. But the html regex in `marked.js` needs to be modified to recognize HTML both **before** and **after** a comment.

Alternatively, it could just mention in the documentation that blank lines are required between the comments and the HTML code.

https://github.com/chjj/marked/blob/84c330b62beee7dab83764f373eaf80414b2516e/lib/marked.js#L23

### LINKED ISSUE #1050: If HTML style comments contain headings they do not work
This is a Markdown sample which gets rendered unexpectedly. For this...

```
aaa
<!-- test
====
-->
bbb
```

...I would expect this HTML output:
```
aaa
bbb
```

But instead it looks like this:

grafik

### LINKED ISSUE #1071: Element names
## Expectation

The regex for detecting element names is incorrect. It detects some chardata sequences as tags and some tags as chardata. It may be that only the latter is problematic, however — I’ll explain what I mean after the examples.

Example 1: this is chardata — `<123>`
Example 2: these are element tags — `<foo.bar></foo.bar>`

## Result

Example 1: `<123>`
Example 2: `<p>&lt;foo.bar&gt;&lt;/foo.bar&gt;</p>\n`

## What was attempted

Reproducible with these minimal examples — `marked('<123>')`, `marked('<foo.bar></foo.bar>')`.

## Explanation

The current tag name matching portion of the `inline.tag` pattern looks like this:

`[a-zA-Z0-9\-]+`

However this isn’t the grammar of HTML tag names (and it also isn’t either a superset or a subset of the grammar). Taking into account that marked has already normalized newline sequences, the lexical production for a tag name can be represented with the regex pattern `[a-zA-Z][^\t\f\n \/>]*`. A tag name begins with an ascii alphabetic (rather than alphanumeric) char and consumes all characters following it until horizontal tab, form feed, line feed, space, slash or greater than is the next character (see [§ 12.2.5.6 ](https://html.spec.whatwg.org/multipage/parsing.html#tag-open-state) and [§ 12.2.5.8](https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state) for more detail).

Although the `<123>` sequence is detected as a tag, since it ends up being passed through as-is, it seems to be harmless; when the resulting HTML is parsed, the agent will understand that this is chardata rather than a tag. However in the reverse case, marked swaps in character references, so the original markup is lost.

### LINKED ISSUE #1090: Prevent from parsing inside custom elements
## Expectation

If I have a custom element in my Markdown, the contents should not be parsed:

```
<demo-el>
    <template>
        <script>
            let death = 'backticks';
            console.log(`backticks are ${death}`;
        </script>
    </template>
</demo-el>
```

Should produce `backticks are backticks` in the console and my custom element rendered as normal in the Markdown.

## Result

The following code actually produces `<code>backticks are ${death}</code>` in the rendered Markdown and an error in the console (because it cannot recognize `<`).

## What was attempted

I tried to find which elements are considered block elements in Marked.js to prevent this from happening, but it was not clear how this was decided.

### LINKED ISSUE #1092: Auto-linking - cm_autolinks test

### LINKED ISSUE #1134: script inside a div is escaped even if sanitize set to false
**Marked version:**
0.3.17+
gfm
 
214468b9b8f5a41d4feb8291af22d7178f60837a

## Expectation
```
<div>
  <script>
    alert('test');
  </script>
</div>
```

## Result
```
<div>
  <script>
    alert(&#39;test&#39;);
  </script>
</div> 
```

## What was attempted
```
<div>
  <script>
    alert('test');
  </script>
</div>
```

If I move the `script` tag outside the `div` tag, it will work.

### LINKED ISSUE #1140: <foo> is parsed as HTML element even if < is escaped with \
**Marked version:** 0.3.17

**Markdown flavor:** CommonMark or GitHub Flavored Markdown

 

## Expectation

Parsed as text and it should not be sanitized

## Result

Parsed as HTML tag and it is passed to a sanitizer

## What was attempted

Install marked with `npm install marked`.

Then put following code as `foo.js`

```javascript

const doc = '\\<div>';

console.log(
    marked(doc, {
        sanitize: true,
        sanitizer: tag => {
            console.log('tag:', tag);
            return tag;
        },
    }),
);
```

Finally run it with `node foo.js`.

It outputs:

```
tag: <div>
<p>\<div></p>
```

The `<div>` just after `\` is parsed as `<div>` HTML element even if `<` is escaped with `\`. However, as long as I confirmed with http://spec.commonmark.org/dingus/, `\<div>` should be parsed as text `'<div>'`. So it should not be passed to sanitizer and should be escaped as `\&lt;div&gt;` by marked.

I have trouble with this problem because sanitizer cannot know whether the input is just a text input (where `<` is escaped with `\`) or actual HTML tag.

### LINKED ISSUE #1141: HTML Attributes without quote is broken
**Marked version:** 0.3.17

**Markdown flavor:** Markdown.pl|CommonMark|GitHub Flavored Markdown

 

## Expectation

Marked can handle HTML attributes without quote like `<a href=https://example.com></a>` in markdown doc. It should be converted into `<p><a href="https://example.com"></a></p>`, the same as `<a href="https://example.com"></a>`.

## Result

Converted into broken HTML `<p>&lt;a href=<a href="https://example.com&gt;">https://example.com&gt;</a></a></p>`

## What was attempted

1. `npm install marked`
2. Put following code to `foo.js`
3. Run `node foo.js`

```javascript
console.log(marked('<a href=https://example.com></a>'));
```

## Call stack & console log

```
<p>&lt;a href=<a href="https://example.com&gt;">https://example.com&gt;</a></a></p>
```

though I expected

```
<p><a href="https://example.com"></a></p>
```

### LINKED ISSUE #1195: Some markdown elements not parsed for content with inline html
**Marked version:** 0.3.19

## Expectation

Rendering the markdown:

```md
<table>
</table>
<br>

## h1

1. a
1. **b**

<table>
</table>
```

## Result
Heading and list not parsed. Here's the html output:

```html
<table>
</table>
<br>

## h1

1. a
1. <strong>b</strong>

<table>
</table>
```

Feder1co5oave:
#### Parenthesis in links
- fixes #366 
- fixes #690 
- fixes #619 
- fixes #736
- fixes #586 

#### Backslash escapes in links
- fixes #829 
- fixes #229

#### HTML
- fixes a bunch of already closed issues (not gonna run through them all)
- fixes #488 
- fixes #1134
- fixes #1090
- does_not_fix #1071 because it is against commonmark
- fixes #1050
- fixes #1032
- closes #833
- fixes #501
- fixes #303 
- fixes #257 
- fixes #1092
- fixes #1140
- fixes #1141 and closes #1142
- fixes #1195 

### BREAKING-CHANGES:
See commit messages (look for "BREAKING CHANGE")
**This deserves a minor version bump.**

### TODOs:
- parse inline markdown inside HTML blocks if parameter `markdown=1`
- handle backslash escapes across the whole document
- consider #897 and re-enable test 478

### Non-compliance with commonmark:
See commit messages (look for "NON-COMPLIANT")

## Contributor

- [x] new feature has been documented in the appropriate places.

## Committer

- [ ] Merge PR


lib/marked.js
```diff
@@ -20,16 +20,25 @@ var block = {
...
   list: /^( *)(bull) [\s\S]+?(?:hr|def|\n{2,}(?! )(?!\1bull )\n*|\s*$)/,
-  html: /^ *(?:comment *(?:\n|\s*$)|closed *(?:\n{2,}|\s*$)|closing *(?:\n{2,}|\s*$))/,
+  html: '^ {0,3}(?:' // optional indentation
+    + '<(script|pre|style)[\\s>][\\s\\S]*?(?:</\\1>[^\\n]*\\n+|$)' // (1)
+    + '|comment[^\\n]*(\\n+|$)' // (2)
+    + '|<\\?[\\s\\S]*?\\?>\\n*' // (3)
+    + '|<![A-Z][\\s\\S]*?>\\n*' // (4)
+    + '|<!\\[CDATA\\[[\\s\\S]*?\\]\\]>\\n*' // (5)
+    + '|</?(tag)(?: +|\\n|/?>)[\\s\\S]*?(?:\\n{2,}|$)' // (6)
+    + '|<(?!script|pre|style)([a-z][\\w-]*)(?:attribute)*? */?>(?=\\h*\\n)[\\s\\S]*?(?:\\n{2,}|$)' // (7) open tag
```
Feder1co5oave: @davisjam it this safe?

lib/marked.js
```diff
@@ -20,16 +20,25 @@ var block = {
...
   lheading: /^([^\n]+)\n *(=|-){2,} *(?:\n+|$)/,
-  paragraph: /^([^\n]+(?:\n?(?!hr|heading|lheading| {0,3}>|tag)[^\n]+)+)/,
+  paragraph: /^([^\n]+(?:\n(?!hr|heading|lheading| {0,3}>|<\/?(?:tag)(?: +|\\n|\/?>)|<(?:script|pre|style|!--))[^\n]+)+)/,
   text: /^[^\n]+/
 };
 
-block._label = /(?:\\[\[\]]|[^\[\]])+/;
-block._title = /(?:"(?:\\"|[^"]|"[^"\n]*")*"|'\n?(?:[^'\n]+\n?)*'|\([^()]*\))/;
+block._label = /(?!\s*\])(?:\\[\[\]]|[^\[\]])+/;
+block._title = /(?:"(?:\\"?|[^"\\])*"|'[^'\n]*(?:\n[^'\n]+)*\n?'|\([^()]*\))/;
```
Feder1co5oave: @davisjam

lib/marked.js
```diff
@@ -47,23 +56,24 @@ block.list = edit(block.list)
...
+  + '|center|col|colgroup|dd|details|dialog|dir|div|dl|dt|fieldset|figcaption'
+  + '|figure|footer|form|frame|frameset|h[1-6]|head|header|hr|html|iframe'
+  + '|legend|li|link|main|menu|menuitem|meta|nav|noframes|ol|optgroup|option'
+  + '|p|param|section|source|summary|table|tbody|td|tfoot|th|thead|title|tr'
+  + '|track|ul';
+block._comment = /<!--(?!-?>)[\s\S]*?-->/;
+block.html = edit(block.html, 'i')
+  .replace('comment', block._comment)
+  .replace('tag', block._tag)
+  .replace('attribute', / +[a-zA-Z:_][\w.:-]*(?: *= *"[^"]*"| *= *'[^']*'| *= *[^\s"'=<>`]+)?/)
```
Feder1co5oave: @davisjam is this safe? it gets replaced into `html`.

lib/marked.js
```diff
@@ -101,6 +111,24 @@ block.tables = merge({}, block.gfm, {
...
 
+/**
+ * Pedantic grammar
+ */
+
+block.pedantic = merge({}, block.normal, {
+  html: edit(
+    '^ *(?:comment *(?:\\n|\\s*$)'
+    + '|<(tag)[\\s\\S]+?</\\1> *(?:\\n{2,}|\\s*$)' // closed tag
+    + '|<tag(?:"[^"]*"|\'[^\']*\'|\\s[^\'"/>\\s]*)*?/?> *(?:\\n{2,}|\\s*$))')
```
Feder1co5oave: @davisjam this should match your fix in #1083

lib/marked.js
```diff
@@ -458,13 +488,18 @@ Lexer.prototype.token = function(src, top) {
...
+  escape: /^\\([!"#$%&'()*+,\-./:;<=>?@\[\]\\^_`{|}~])/,
   autolink: /^<(scheme:[^\s\x00-\x1f<>]*|email)>/,
   url: noop,
-  tag: /^<!--[\s\S]*?-->|^<\/?[a-zA-Z0-9\-]+(?:"[^"]*"|'[^']*'|\s[^<'">\/\s]*)*?\/?>/,
-  link: /^!?\[(inside)\]\(href\)/,
-  reflink: /^!?\[(inside)\]\s*\[([^\]]*)\]/,
-  nolink: /^!?\[((?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]])*)\]/,
+  tag: '^comment'
+    + '|^</[a-zA-Z][\\w:-]*\\s*>' // self-closing tag
+    + '|^<[a-zA-Z][\\w-]*(?:attribute)*?\\s*/?>' // open tag
```
Feder1co5oave: @davisjam same as above

lib/marked.js
```diff
@@ -473,24 +508,34 @@ var inline = {
...
 inline._scheme = /[a-zA-Z][a-zA-Z0-9+.-]{1,31}/;
 inline._email = /[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+(@)[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+(?![-_])/;
-
 inline.autolink = edit(inline.autolink)
   .replace('scheme', inline._scheme)
   .replace('email', inline._email)
-  .getRegex()
+  .getRegex();
+
+inline._attribute = /\s+[a-zA-Z:_][\w.:-]*(?:\s*=\s*"[^"]*"|\s*=\s*'[^']*'|\s*=\s*[^\s"'=<>`]+)?/;
```
Feder1co5oave: @davisjam

lib/marked.js
```diff
@@ -473,24 +508,34 @@ var inline = {
...
-inline._inside = /(?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]]|\](?=[^\[]*\]))*/;
-inline._href = /\s*<?([\s\S]*?)>?(?:\s+['"]([\s\S]*?)['"])?\s*/;
+inline.tag = edit(inline.tag)
+  .replace('comment', block._comment)
+  .replace('attribute', inline._attribute)
+  .getRegex();
+
+inline._label = /(?:\[[^\[\]]*\]|\\[\[\]]?|`[^`]*`|[^\[\]\\])*?/;
+inline._href = /\s*(<(?:\\[<>]?|[^\s<>\\])*>|(?:\\[()]?|\([^\s\x00-\x1f()\\]*\)|[^\s\x00-\x1f()\\])*?)/;
+inline._title = /"(?:\\"?|[^"\\])*"|'(?:\\'?|[^'\\])*'|\((?:\\\)?|[^)\\])*\)/;
```
Feder1co5oave: @davisjam check these 3

USING_ADVANCED.md
```diff
@@ -43,7 +43,7 @@ console.log(myMarked('I am using __markdown__.'));
 |:----------|:---------|:----------------------------------------------------------------------------------------------------------------------------|
 |highlight  |`function`|A function to highlight code blocks. See also: <a href="#highlight">Asynchronous highlighting</a>.                           |
 |renderer   |`object`  |An object containing functions to render tokens to HTML. See [extensibility](https://github.com/markedjs/marked/blob/master/USAGE_EXTENSIBILITY.md) for more details. Default: `new Renderer()`|
-|pedantic   |`boolean` |Conform to obscure parts of `markdown.pl` as much as possible. Don't fix original markdown bugs or behavior. Default: `false`|
+|pedantic   |`boolean` |Conform to obscure parts of `markdown.pl` as much as possible. Don't fix original markdown bugs or behavior. Turns off and overrides `gfm`. Default: `false`|
```
styfle: FYI, one of us is going to get a conflict because I moved this file in PR #1138

Feder1co5oave:
I think Git is clever enough to automatically resolve this type of conflict
if you only renamed the file

joshbruce:
Outside of the doc conflict, what's stopping us from merging this?

UziTech:
It would be nice to get @davisjam 's review to check if the new regex's are safe

joshbruce:
@Feder1co5oave: I know most of us are waiting on a security check from @davisjam on this PR. Curious if you would be willing and able to update the PR to use the CommonMark test bed from #1160 .

davisjam:
Sorry, running it now!

lib/marked.js
```diff
@@ -505,7 +550,13 @@ inline.normal = merge({}, inline);
 
 inline.pedantic = merge({}, inline.normal, {
   strong: /^__(?=\S)([\s\S]*?\S)__(?!_)|^\*\*(?=\S)([\s\S]*?\S)\*\*(?!\*)/,
-  em: /^_(?=\S)([\s\S]*?\S)_(?!_)|^\*(?=\S)([\s\S]*?\S)\*(?!\*)/
+  em: /^_(?=\S)([\s\S]*?\S)_(?!_)|^\*(?=\S)([\s\S]*?\S)\*(?!\*)/,
+  link: edit(/^!?\[(label)\]\(\s*<?([\s\S]*?)>?(?:\s+(['"][\s\S]*?['"]))?\s*\)/)
```
davisjam: This regex is vulnerable.

davisjam:
Here's what my vuln-regex-detector tools said:

```
"vulnRegexes":["^!?\\[(label)\\]\\(\\s*<?([\\s\\S]*?)>?(?:\\s+(['\"][\\s\\S]*?['\"]))?\\s*\\)"],"file":"/tmp/marktex/lib/marked.js"
```
(JSON-encoded).

I flagged the appropriate line in the PR.

lib/marked.js
davisjam: A static scan of the tree on @Feder1co5oave's PR says:

```
"vulnRegexes":["^!?\\[(label)\\]\\(\\s*<?([\\s\\S]*?)>?(?:\\s+(['\"][\\s\\S]*?['\"]))?\\s*\\)"]
```
(JSON-encoded).

Checking the ones requested individually...

lib/marked.js
davisjam: Safe.

lib/marked.js
davisjam: Safe.

lib/marked.js
davisjam: Safe.

lib/marked.js
davisjam: Could you print out `html.source`?

lib/marked.js
davisjam: Could you print out `block.html` after this operation?

lib/marked.js
davisjam: Could you print out `inline.tag`?

davisjam: Nevermind, I did it. This regex is safe.

lib/marked.js
davisjam: Nevermind, I did it. Safe.

lib/marked.js
davisjam: Nevermind, I did it. The resulting regex is safe.

davisjam:
OK, @Feder1co5oave I went through all the regexes you flagged. When you said "This should match" I trusted you :-). I checked all statically declared regexes and the dynamically generated ones you flagged.

lib/marked.js
Feder1co5oave: I'm pretty sure this is the "old" link regex, I didn't change it.
Could you point out where the issue is? I'm thinking this part?

    (?:\s+(['"][\s\S]*?['"]))?

The double ?, maybe? I'm kinda lost.

davisjam: This is what `check-regex.pl` says:

```
{"pattern":"^!?\\[(label)\\]\\(\\s*<?([\\s\\S]*?)>?(?:\\s+(['\"][\\s\\S]*?['\"]))?\\s*\\)",
"evilInput": {"suffix":"e","pumpPairs":[
{"pump":"a","prefix":"![label](\t"},{"prefix":"a","pump":"\t"}]
},
"timeLimit":"5",
"nPumps":"250000",
"language":"javascript"}
```

The issue is the use of `[\s\S]*?` twice.
This results in a quadratic regex.

Feder1co5oave: Ok I see where we're getting at... I'm not sure what that output means, can you translate that into some example input that triggers the quadratic behavior?
I was thinking, maybe...?

    ![label](\t\t\t\t\t\t\t\t\t...

davisjam: `![label](\taaaaaaaaaaaaaaaa...aaaaaa\t\t\t\t\t\t\t\t....\t\t\te`

davisjam: Not sure if that string is overly complex or how many pumps you need.

Feder1co5oave:
@joshbruce I'm not sure about what would need to be done to do that...

Feder1co5oave:
I've merged but somehow I broke commonmark [example 60](http://spec.commonmark.org/0.28/#example-60).
I'm investigating this because I think it is caused by an ambiguity in the spec.

@joshbruce I guess there's no easy way to re-enable previously ignored tests that now pass?

UziTech:
@Feder1co5oave I was thinking about that. We could use [jasmine-should-fail](https://www.npmjs.com/package/jasmine-should-fail) for the failing tests. that way if the tests don't fail then jasmine will fail alerting us that one of the tests that should fail started working.

UziTech:
or we could just use a `expect(...).not.toBe(...)`

lib/marked.js
Feder1co5oave: I'm not sure how to fix this.

Feder1co5oave:
Why do I get different number of failures every time I run `npm test` ?!

`npm test:integration` gives me 0 failures.

UziTech:
jasmine runs tests in a random order so there must be some global state that is affecting some tests

lib/marked.js
davisjam: Can you try with `/^!?\[(label)\]\((\s*(<[^>]*>))?(?:\s+(['"][\s\S]*?['"]))?\s*\)/` ?

[existing](https://regexper.com/#%2F%5E!%3F%5C%5B%28label%29%5C%5D%5C%28%5Cs*%3C%3F%28%5B%5Cs%5CS%5D*%3F%29%3E%3F%28%3F%3A%5Cs%2B%28%5B'%22%5D%5B%5Cs%5CS%5D*%3F%5B'%22%5D%29%29%3F%5Cs*%5C%29%2F), [proposed](https://regexper.com/#%2F%5E!%3F%5C%5B%28label%29%5C%5D%5C%28%28%5Cs*%28%3C%5B%5E%3E%5D*%3E%29%29%3F%28%3F%3A%5Cs%2B%28%5B'%22%5D%5B%5Cs%5CS%5D*%3F%5B'%22%5D%29%29%3F%5Cs*%5C%29%2F)

Feder1co5oave: wrapping the URI in angular parenthesis is optional.
The issue is that in this original rule, spaces where allowed inside the URI, which _could_ then be followed by space and a title wrapped in single/double quotes. There is no easy way of finding where the uri stops and when the title begins, but for the presence of the opening quote (which in some cases could be part of the URI instead?). The rule itself is ambiguous.

davisjam: It doesn't sound like this can be done cleanly in a single regex. Can you provide a link to the spec for this rule?

UziTech:
We could set it to not be random but that won't really solve the problem, that would only hide the symptom.

davisjam:
If we set it to non-random *and the error manifests* then we can fix it and confirm the fix worked. If not then we won't learn much.

lib/marked.js
```diff
@@ -461,13 +491,18 @@ Lexer.prototype.token = function(src, top) {
...
-  link: /^!?\[(inside)\]\(href\)/,
-  reflink: /^!?\[(inside)\]\s*\[([^\]]*)\]/,
-  nolink: /^!?\[((?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]])*)\]/,
+  tag: '^comment'
+    + '|^</[a-zA-Z][\\w:-]*\\s*>' // self-closing tag
+    + '|^<[a-zA-Z][\\w-]*(?:attribute)*?\\s*/?>' // open tag
+    + '|^<\\?[\\s\\S]*?\\?>' // processing instruction, e.g. <?php ?>
+    + '|^<![a-zA-Z]+\\s[\\s\\S]*?>' // declaration, e.g. <!DOCTYPE html>
+    + '|^<!\\[CDATA\\[[\\s\\S]*?\\]\\]>', // CDATA section
+  link: /^!?\[(label)\]\(href(?:\s+(title))?\s*\)/,
```
Feder1co5oave: @davisjam the commonmark spec is covered by this rule (and subrules at lines 530 and following). This is safe because the href part can't contain spaces, so the URI stops at the closing ) or at the first space.

Feder1co5oave:
```
$ node_modules/.bin/jasmine --config=jasmine.json --random=false
Started
................................................................................................................................................................................................................................................................................................................................................................................................................................................................


448 specs, 0 failures
Finished in 1.426 seconds
```

lib/marked.js
Feder1co5oave: This is the "legacy" link rule of marked, it's been like this for a long time. I didn't change it. It doesn't follow any clear spec, it just tries its best. We should come up with a different rule that doesn't break compatibility.
However, keep in mind this is used only in pedantic mode, I think very few people every used it and it is planned to be deprecated soon.

davisjam: OK. Since you didn't change it, it shouldn't prevent the PR from being merged.

UziTech:
@davisjam The problem is that some tests are failing based on what order they run in. If the tests are not randomized, how are we going to know which tests are failing because of marked and which tests are failing because of an issue with the test runner?

joshbruce:
@UziTech: The failing due to the order is probably due to the singleton problem I mentioned before. We could add a beforeEach or an afterEach to reset Marked to default.

@Feder1co5oave: The baseline is set up so that all tests pass to prevent regression. From the Travis output, it appears something is messing up the previously passing block tests. The Travis output shows tests in the following suite are failing:

```js
describe('CommonMark 0.28 Indented code blocks', function() {
  var section = 'Indented code blocks';

  // var shouldPassButFails = [];
  var shouldPassButFails = [82];

  var willNotBeAttemptedByCoreTeam = [];

  var ignore = shouldPassButFails.concat(willNotBeAttemptedByCoreTeam);

  cmSpec.forEach(function(spec) {
    messenger.test(spec, section, ignore);
  });
});
```

Which should mean something other than singleton problem. Though it appears the tests running locally are passing.

davisjam:
@UziTech I was hoping the error would be recreatable without the randomness, in which case debugging would be simpler. Alas, apparently not.

Feder1co5oave:
@joshbruce I think this is due to the singleton problem.
When running all jasmine tests, test/index.js is required, and in that script, marked.setOptions is called, affecting the other integration tests.
Is there a way of randomizing tests only inside a single file?

UziTech:
Instead of removing randomization you could set the defaults back to the original defaults in a beforeEach

UziTech:
I created a PR that will reset the default options before each test #1212

styfle:
@UziTech Why did travis not run CI for this PR?

joshbruce:
Would also like to make sure they run against the known test suites from CommonMark and GFM.

UziTech:
I'm not sure why travis didn't run 😕

I did merge the latest changes  so this should be run against cm and gfm tests

here is the travis build from my merge branch: https://travis-ci.org/UziTech/marked/builds/364667018

styfle:
image

Alright, let's do this! 🎉 🍾 🎈

lib/marked.js
UziTech: @Feder1co5oave 	what is the `\\h*` for in `(?=\\h*\\n)`?

UziTech: it that supposed to be `\\n*`?

PR #1144, opened 2018-03-15
Start ordered lists using the initial numbers from markdown lists

### LINKED ISSUE #1143: Ordered lists always start with 1, regardless of starting number in markdown
**Marked version:** 0.3.17

## Expectation

Ordered lists should start with the number used to begin the list in markdown.  *e.g.* given:

```markdown
3. One
4. Two
```

We should see:

```html
<ol start="3">
<li>One</li>
<li>Two</li>
</ol>
```

## Result

We always start at 1.

```html
<ol>
<li>One</li>
<li>Two</li>
</ol>
```

## What was attempted

Appears to be the root cause of an issue logged for Atom's markdown-preview plugin: https://github.com/atom/markdown-preview/issues/502

## References:

CommonMark, [Lists](http://spec.commonmark.org/0.28/#lists):

> The start number of an ordered list is determined by the list number of its initial list item. The numbers of subsequent list items are disregarded.

(see also the examples in that section)

The GFM specs carry this requirement over.

paulroub:
Adds tests for list creation and continuation when starting with a
number other than 1.

Emits 'start' attribute only when necessary; simple cases behave as
always.

**Marked version:** 0.3.17

## Description

- Fixes #1143

## Contributor

- [X] Test(s) exist to ensure functionality and minimize regresstion (if no tests added, list tests covering this PR); or,

## Committer

- [x] Merge PR


joshbruce:
@paulroub: Thanks again for this contribution! 

We've started moving toward using Jasmine for our test harness. We are also looking to stick as close we can to objective examples from the specifications themselves. There are two entries in the CommonMark spec that seem be related to this issue. Would you be willing and able to create tests for these two examples using the proposed solution?

http://spec.commonmark.org/0.28/#example-230 - starting at 0 results in a forced starting number.

http://spec.commonmark.org/0.28/#example-234 - start at 10 and have a code block within a list item.

If you can add a test in `/test/integration/marked-spec.js` file it will help us determine where marked is objectively failing against the spec instead of any one of our interpretations of it. Tagging @UziTech to help with any test harness questions.

joshbruce:
ps. Please keep the tests you have (or copy and paste from the spec to update html and MD files).

joshbruce:
@paulroub: This looks good to me. Having said that, I would like to get a second and possibly a third pair of eyes. Thank you so much!

lib/marked.js
```diff
@@ -836,9 +839,10 @@ Renderer.prototype.hr = function() {
   return this.options.xhtml ? '<hr/>\n' : '<hr>\n';
 };
 
-Renderer.prototype.list = function(body, ordered) {
-  var type = ordered ? 'ol' : 'ul';
-  return '<' + type + '>\n' + body + '</' + type + '>\n';
+Renderer.prototype.list = function(body, ordered, start) {
+  var type = ordered ? 'ol' : 'ul',
+      startatt = (ordered && +start !== 1) ? (' start="' + start + '"') : '';
```
styfle: Why are you casting `+start` here? Isn't it already a number from `+bull` above?

PR #1151, opened 2018-03-21
Fix usage links in USING_ADVANCED.md ( 2nd )

wraith13:
Just fixes link path in document.


styfle:
Thanks for contributing! 👍

PR #64, opened 2012-07-14
Demo page - live preview, like showdown.im

fidian:
Works in Chrome on Linux.  Haven't tried with IE yet.

Please let me know if you'd think this sort of thing would be useful for others.

PR #1156, opened 2018-03-22
Docs navigation

UziTech:
**Marked version:** latest

## Description

This PR adds a ribbon that links to the github repo and adds a link to marked icon back to the README.md

image

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 <ul>
   <li><a href="#marked">About</a></li>
-  <li><a href="#install">Installation</a></li>
+  <li><a href="#installation">Installation</a></li>
   <li><a href="#usage">Usage</a></li>
   <li><a href="#specifications">Supported Markdown specifications</a></li>
   <li><a href="#security">Security</a></li>
@@ -11,7 +11,7 @@
 
 <h2 id="marked">Marked</h2>
 
-Marked is 
+Marked is
 
 1. built for speed.<sup>*</sup>
 2. a low-level markdown compiler that allows frequent parsing of large chunks of markdown without caching or blocking for long periods of time.<sup>**</sup>
@@ -96,4 +96,3 @@ For list of credited authors and contributors, please see our [authors page](AUT
 Copyright (c) 2011-2018, Christopher Jeffrey. (MIT License)
 
 See [license](https://github.com/markedjs/marked/blob/master/LICENSE.md) for more details.
-
--- a/docs/index.html
+++ b/docs/index.html
@@ -12,6 +12,7 @@
         }
 
         #container {
+            position: relative;
             max-width: 800px;
             margin: auto;
             padding: 10px;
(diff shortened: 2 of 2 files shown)


UziTech:
We should probably also make a way to be able to link to different parts of the site.

Because this is a SPA there is no way to view AUTHORS.md without clicking on a link in https://marked.js.org/

joshbruce:
@UziTech: I was wondering if a SPA would be the desired approach or not. Namely because of the double-anchor problem. /# (makes it a SPA) /#something#inline (becomes a different problem to solve).

One way to do it would be to use multiple index.html file within directories. Or, use something like VueJS or Angular.

joshbruce:
Would also be nice if the README still linked to different documents. Like having two links in the usage section - one to the advanced and another to the pro.

UziTech:
> Namely because of the double-anchor problem.

Ya I through together a `hashchange` router at https://github.com/UziTech/marked/commit/62b697afb09ced16e7dafbf222f4b61a5817f807 but it has some problems with the back and forward buttons.

I'm thinking the index files in directories would be a better solution

styfle:
I was thinking of doing something like using urls like `marked.js.org/#page=AUTHORS.md` which would render the AUTHORS.md file as html.

This would give us a way to render additional non-docs pages that are not part of the docs SPA such as `marked.js.org/demo.html`...see this: https://github.com/markedjs/marked/pull/64

joshbruce:
That alternative make my UX brain hiccup a bit...to computer, not enough human, if that makes sense. Same issues with `?p=5` from single-page server-side. We might be able to use just the router from something like Angular...that's all we're really looking for.

A router that doesn't break that back and forward buttons, while also allowing for inline anchors.

Still thinking the cheapest and easiest would be multiple index.html files:

```
/index.html <- readme with demo (that could be cool) see Haskell https://www.haskell.org
  /authors/index.html <- renders the md
  /code-of-conduct
  /contributing
  /publishing
  /using <- display readme again
    /advanced
    /pro
```

Just clean and simple human readable urls. Only drawback is that there would be some duplication of code...but I don't think it's that big of a deal for such a small static site.

??

joshbruce:
Just realized the demo thing is what #64 is about. :)

styfle:
Or we could avoid re-inventing the wheel and go with this: https://docsify.js.org

joshbruce:
Interesting. Could we use Jekyll (native GitHub Pages). Given our current focus - we could use the marked site to showcase some of our personal projects...touching on the professional growth itch.

I've been working on a design system based on the [USWDS](https://github.com/uswds/uswds) for 8fold and the practitioner sites: https://www.npmjs.com/package/8fold-web-assets

Definitely thinking starting a separate repo would be good as well. I don't mind reimagining the wheel so long as it helps us grow professionally in an environment that's not mission critical to us eating. :)

PR #1160, opened 2018-03-24
CommonMark Tests

### LINKED ISSUE #1159: Build CommonMark test bed
**Marked version:** 0.3.18

**Proposal type:** other

Our current test bed does not reflect the specifications in their entirety. Creating one test at a time is not a viable way forward (see #1124 & #1129).

**Other resources:**

http://spec.commonmark.org/0.28/spec.json
https://github.com/Feder1co5oave/marktex/commit/ec2e8731bfe2f1d69b6c228304ea29d7dff0c544
https://github.com/Feder1co5oave/marktex/blob/0e2b3eea3ec8013377e6fc4ea3d72624ad1c40c2/test/index.js

Use the spec JSON to generate test files. Thinking the following makes the most sense for readability and whatnot.

**Template:**

```js
// [flavor short name]_[section].js
it('should pass [flavor short name] example [example number]', function() {
  var result = marked([spec markdown]);
  var expected = [spec html];

  expected(result).toBe(expected);
}
```

**Live example:**
```js
// cm_list_items.js
it('should pass cm example 230', function () {
  var result = marked(0. ok\n);
  var expected = '<ol start=\"0\">\n<li>ok</li>\n</ol>\n';

  expected(result).toBe(expected);
}

it('should pass cm example 234', function () {
  var result = marked('  10.  foo\n\n           bar\n');
  var expected = '<ol start=\"10\">\n<li>\n<p>foo</p>\n<pre><code>bar\n</code></pre>\n</li>\n</ol>\n';

  expected(result).toBe(expected);
}
```

joshbruce:
**Marked version:** 0.3.18

## Description

- Fixes #1159 

Putting a script together in PHP (it's what I know and this should be a one time deal) to generate the test cases. However, the test harness does not seem to run all files in the `integration` folder. Given the number of test cases from the specs, think it would be better to divide the tests by flavor and section.

@UziTech: Thoughts??

## Contributor

## Committer

- [ ] Merge PR


joshbruce:
@UziTech:

1. @Feder1co5oave gave it to us back on #1124 - it's CommonMark 0.28 (not sure if an equivalent exists for the GFM extensions. http://spec.commonmark.org/0.28/spec.json

2. (Not sure where the extra step is coming up.) I'm only using the PHP to generate the tests that I then copy and paste into the JS file. Once all the tests are accounted for - no more PHP - I didn't have to do it the "long" way (what I and @davisjam were doing before). If the JSON converter was going to stick around I might figure out how to do it in JS...since I only intend to keep it around long enough to create the test bed I did it the easiest way I knew how - PHP's standard library (and PHP was built to spit out strings).

3. My question or request at the moment is to make the integration test run all files inside the folder (well all the tests running all the files within a given directory). Because there will most like be half a dozen sections: cm_028_tabs.js, cm_028_thematic_breaks.js, cm_028_headings.js, and so on. And, each section will have half a dozen tests or so.

UziTech:
The manual build step comes from creating the tests every time a new version of the spec comes out.

I say we should just create a script to load and run the json file instead of building the tests. That way when a new version of the spec comes out all we have to do is copy the new json file.

joshbruce:
@UziTech: Part of me wants to agree with that. For examples we aren't going to comply with are you thinking we would have something like an array of examples we aren't complying with?

```js

var noComply = [];
var shouldComplyButFail = [];

```

UziTech:
We could add something to the json like:

```
{
  markdown: ...,
  html: ...,
  noComply: true,
  ...
}
```

joshbruce:
We could. Though, if we have to modify the JSON I think we would be slipping back to a pre-build step, yeah? 0.29 gets released, we modify that JSON. I could take what @davisjam has put together.

1. Run the tests...by itself won't result in merge because Travis will fail.
2. Create a `shouldPass` array to denote broken examples...ignore them so Travis will pass.
3. Create a `willNotComply` to denote rules that are too complex for us to resolve (some of the `em` rules identified by @Feder1co5oave, for example).

Doesn't require modifying the JSON. Allows test bed to cover all the things while indicating broken things and examples we have no intention of covering. This would put a little bit more burden back on the developers to read the JSON to find out exactly what we're not in compliance with (something they get for free by doing it the longer way).

joshbruce:
@UziTech: Been looking at some of the failing tests. Looks like it could be an XHTML versus HTML5 issue. For self-closing elements XHTML says to add a forward slash, HTML5 doesn't have this requirement. It looks like Marked is spitting out the proper HTML5 but the HTML from the spec has the trailing slash.

```
407) CommonMark 0.28 should pass example 551
  Message:
    Expected '<p><img src="url" alt="foo"></p>
    ' to be '<p><img src="url" alt="foo" /></p>
    '.
```

Debating on have the script remove the trailing slash. Thoughts??

UziTech:
hmm ... I'm thinking this might be a better question for @Feder1co5oave.

Would it be easier to remove the slashes from the json file every time we update the spec? or would it be easier to add slashes in marked to copy the spec?

I don't think the spec requires them.

UziTech:
A third option could be coming up with some AST sort of approach for comparing them instead of comparing strings.

joshbruce:
1. Removing the slash from the JSON should be a matter of string replacement...can't imagine it being too difficult.
2. Might be able to set the XTHML option to true?? Right now it's marked out of the box, which is X?HTML false. Having said that, not sure if our XHTML implementation uses the same spacing.
3. Lost me on the acronym...AST??

Feder1co5oave:
Enabling XHTML mode solves the closing slash issue. I used this in my last
pr.

Space is ignored by the current comparing algorithm.

My old pr to use html-differ uses parse5 under the hood to compare html,
that's an AST.

Il mer 28 mar 2018, 01:35 Josh Bruce <notifications@github.com> ha scritto:

joshbruce:
1. Latest attempt (https://github.com/markedjs/marked/pull/1160/commits/f1445664e6acb05db0629450386dc3ac976b31d6) shows 624 with 379 failures.
2. Changing the configuration option to use XHTML gave mixed results. Same number of failing tests before but not the same tests.
3. Interesting. What does AST stand for in this context (abstract syntax tree??)? It appears the Jasmine tests include whitespace comparison. Some of the failing tests are because of new line characters.

joshbruce:
Add `html-differ` and `jasmine-custom-message`. Brings us to 284 failures. Next step will be the ignores.

Would like to get feedback, if any, on this as a viable solution for the future from @UziTech and @Feder1co5oave. Tagging in the rest of the core team as well.

styfle:
@joshbruce I think the purpose of doing an [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) comparison is to compare the semantics/meaning not the syntax. This would ignore whitespace and trailing slashes, probably even disregard the difference between quoted/unquoted attributes.

However, it does place some faith in the tool that is converting the HTML to an AST.

joshbruce:
@styfle: Agreed. I grabbed [Html Differ](https://github.com/bem/html-differ#ignorewhitespaces-boolean), which seems to have pretty rational defaults for what we are looking to accomplish. There wasn't something that explicitly said it would ignore the XHTML closure but that logic is there as well.

I've also added the skip logic for the 284 failing tests; so, this PR is considered green and will run all the passing tests. Given that we don't know how long it will take (or if we will) fix the failures. 

To the best of my knowledge this is the most honest and complete view of Marked according to the spec.

joshbruce:
Some of the failing tests seem to be related to the header id feature, which is under consideration for deprecation (#1043) and is not part of the specification. 

Is there a way to turn it off in the settings?

UziTech:
I added a PR (https://github.com/bem/html-differ/pull/154) to html-differ to ignore the self closing slash but it doesn't look like it is still in active development. The last release was Aug 2016.

test/integration/marked-spec.js
```diff
@@ -1,18 +1,51 @@
-var marked = require('../../marked.min.js');
+var sut = require('../../marked.min.js');
```
UziTech: why `sut`?

joshbruce: System Under Test. 

The hope is to distinguish it from the other modules which are there to aid in testing while not being "under test".

joshbruce: See also: https://en.wikipedia.org/wiki/System_under_test

joshbruce:
@UziTech: Is this a library we want to task ourselves with? or should we find a different differ (see what I did there?)? or is it as complete as it needs to be for our purposes?

UziTech:
I like the idea of comparing html vs comparing strings. We could just fork html-differ under markedjs for our own use.

styfle:
@UziTech It's not ideal but it would certainly work. Are you planning to publish to npm as `npm-differ2` or something like that?

joshbruce:
My inner PO wants to scale things back a bit, to be honest just to make sure we're staying focused.

The point of this exercise is to see where Marked stands regarding the CommonMark spec. I think we have done that for the most part. Beyond the few header ID issues, the rest are all honest failures with marked to deliver its promise...it's possible some of those will be resolved by #1135.

1. Re header IDs: I think modifying marked would be the more ideal way to go in the long term. Having the option on marked to not generate the header IDs seems more appropriate than telling the `diff` tool to ignore them.
2. Re stripping XHTML closings: Given the issues on `html-differ` related to the closing slash, the submitted PR is cool. Having said that, ideally, I would like to be able to disable/enable XHTML on marked and have it not dramatically change the test results (as it seems to do at present); otherwise, removing the trailing slash seems to work.
3. Re forking: I was kind of half-kidding when I asked if it's something we should take on. As long as the README is updated with language that says, "For use with the testing the Marked library only..." so people who decide to use it in their projects anyway (because they most likely will) know not to expect us to fix issues unrelated to testings Marked...and that we are very deliberate in the changes we make (ensure proper ownership of functionality so to speak).

joshbruce:
@UziTech and @styfle: Rebased to use the `headerIds` option. `setOptions` doesn't seem to work. 

When I did `marked.setOptions({ headerIds: false });` I was still receiving the failing header tests. When I did `marked.Renderer({ headerIds: false });` the header tests pass.

Thoughts??

test/integration/marked-spec.js
```diff
@@ -1,18 +1,61 @@
 var marked = require('../../marked.min.js');
```
UziTech: we should probably change this to `../../lib/marked.js`

UziTech: setOptions seems to work when this is changed

joshbruce:
Think this is as far as this PR needs to go (??).

- 624 examples from the spec get run.
- 371 pass, which gets us to about 60% spec-compliance. 
- Separated failures by section to perform further analysis.

Couple of issues:

1. It seems the header ID tests from the `new` directory are failing??
2. The `editorconfig` and lint aren't working for me right now - not a deal-breaker and #1185 might resolve.

joshbruce:
re setting options, see #1199

joshbruce:
Think I'm officially tired of staring at this. lol

1. Each Jasmine test suite (`describe`) represents a single section of the specification; therefore, 100% passing sections are there and should ensure against regressions.
2. Each test suite only identifies which examples are not currently passing or those that will not be attempted by the Marked core team. This makes checking in PRs pretty simple...the example is removed from the list of examples to ignore...and CI passes; if the solution didn't work, CI wouldn't pass.
3. Adding the GFM should become easier, as it will be new sections for the table, and other, extensions.
4. The front matter gives potential users and contributors a better understanding of where Marked is compared to the specs.
5. The first test suite gives instructions to potential contributors on how to read the follow-on test suites. (There was one other consolidation I considered but figured it would be overkill. Further, this may appear like there would be a performance hit from looping over the spec so much but I didn't feel it and the gains far outweighed that concern for me.)

joshbruce:
Would really like to have a separate file per spec. @UziTech what changes can we do to make that happen? Think it would be nice to have this set at least in place when we do the 0.4.x, which will most likely be the next release.

Instead of `marked-spec` would be something like `commonmark_0_00_28` then GFM would be something similar. Don't want to overrun people with test suites in a single file.

test/integration/marked-spec.js
```diff
@@ -1,17 +1,453 @@
...
+
+Messenger.prototype.message = function(spec, expected, actual) {
+  return 'CommonMark (' + spec.section + '):\n' + spec.markdown + '\n------\n\nExpected:\n' + expected + '\n------\n\nMarked:\n' + actual;
+}
+
+Messenger.prototype.test = function(spec, section, ignore) {
+  if (spec.section === section && ignore.indexOf(spec.example) < 0) {
+    var expected = spec.html;
+
+    var actual = marked(spec.markdown, { headerIds: false, xhtml: true });
```
UziTech: the `actual` and `expected` variables should be assigned inside the `it` function otherwise they will be calculated before any `beforeEach/All` functions

UziTech:
We could move these specs to a `test/integration/commonmark` folder and create a `test/integration/gfm` folder with a similar layout

joshbruce:
With this latest iteration I ran `npm test` instead of `npm run test:integration` and got the failures pointed out by CI. However, `npm run test:integration` didn't bring up those errors. Any thoughts on this @UziTech?? (That last commit seems like it had to be done for odd circumstances, I did it because I trust Travis.)

Think folder idea is a good way to go. You wanna do a separate PR for that; or, should we incorporate it here??

joshbruce:
@UziTech: Travis is failing things that aren't failing locally. Help a brother out??

test/integration/marked-spec.js
```diff
@@ -1,17 +1,451 @@
...
+  cmSpec.forEach(function(spec) {
+    messenger.test(spec, section, ignore);
+  });
+});
+
+describe('CommonMark 0.28 Emphasis and strong emphasis', function() {
+  var section = 'Emphasis and strong emphasis';
+
+  // var shouldPassButFails = [];
+  var shouldPassButFails = [447, 450, 445, 409, 448, 444, 395, 402, 443, 396, 398, 403, 372, 389, 456, 404, 371, 420, 455, 422, 387, 414, 452, 458, 421, 457, 424, 451, 426, 423, 416, 454, 349, 347, 453, 346, 353, 432, 378, 428, 369, 332, 381, 431, 363, 359, 380, 360, 355, 356, 342, 388, 364, 368, 433, 435, 434, 392, 348, 382, 436, 333, 449, 427, 377, 334];
```
UziTech: add numbers 352, 354, 393, 394, and 419 to skip the failing tests

joshbruce: If CI does something different from my local environment, I'm not gonna be able to trust my local, you know? Is there a reason this is happening?

joshbruce:
@styfle: You up for be merge master?? Thinking we finally got it!

PR #1161, opened 2018-03-25
Move demo to docs

styfle:
This PR adds a demo page to the docs.
Note: this is continued work from #64

## Contributor

## Committer

- [x] Merge PR


UziTech:
We should probably also link to the demo page in the docs/README.md

styfle:
@UziTech I almost forgot!

I made a bunch of changes including "Add link to demo from readme"

Thanks 👍

joshbruce:
Might be nice to make the two READMEs the same. Feels weird to me to have the README on the main repo - then see a more robust and complete one on the site; especially given the main README is the one posted to NPM. 

The links and everything should be able to remain the same.

styfle:
@joshbruce Yes I understand this might be a little confusing. But the idea is to not be publishing MB of data to npm because users who run `npm install --save marked` typically aren't referencing the README.md that they just downloaded.

However, it is a good starting point for people coming to the github page or npm page so they at least see one example usage and a link to the full docs.

Maybe the confusion can be avoided if we rename `/docs/README.md` to `/docs/INDEX.md`?

joshbruce:
@styfle: That's fair, and pretty understandable. I think I'm looking at it from the perspective of what is the minimum information a new developer would need or want to get started and optionally dive into higher level of complexity.

So, I think the README.md could be used without having a duplicated file in the `/docs`. The concept of the README.md being the locus of documentation seems to be the way GitHub and NPM operate; therefore, not having two files would make the most sense in my book.

PR #1164, opened 2018-03-26
Alphebatize block

joshbruce:
**Marked version:** 0.3.19

**Markdown flavor:** n/a

## Description

Pulls member declarations into block, standardizes naming convention, and alphabetizes the rules.  

All tests run during `node test` were checked.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -8,35 +8,54 @@
 'use strict';
 
 /**
- * Block-Level Grammar
+ * Markdown block-level element rules
+ *
+ * QUESTION: What is the different between heading the lheading?
+ *
+ * TODO: Make object with properties that can call functions.
  */
-
 var block = {
-  newline: /^\n+/,
-  code: /^( {4}[^\n]+\n*)+/,
-  fences: noop,
-  hr: /^ {0,3}((?:- *){3,}|(?:_ *){3,}|(?:\* *){3,})(?:\n+|$)/,
-  heading: /^ *(#{1,6}) *([^\n]+?) *#* *(?:\n+|$)/,
-  nptable: noop,
   blockquote: /^( {0,3}> ?(paragraph|[^\n]*)(?:\n|$))+/,
-  list: /^( *)(bull) [\s\S]+?(?:hr|def|\n{2,}(?! )(?!\1bull )\n*|\s*$)/,
-  html: /^ *(?:comment *(?:\n|\s*$)|closed *(?:\n{2,}|\s*$)|closing *(?:\n{2,}|\s*$))/,
+  bullet: /(?:[*+-]|\d+\.)/, 
+  code: /^( {4}[^\n]+\n*)+/,
   def: /^ {0,3}\[(label)\]: *\n? *<?([^\s>]+)>?(?:(?: +\n? *| *\n *)(title))? *(?:\n+|$)/,
-  table: noop,
+  fences: noop,
+  heading: /^ *(#{1,6}) *([^\n]+?) *#* *(?:\n+|$)/,
+  hr: /^ {0,3}((?:- *){3,}|(?:_ *){3,}|(?:\* *){3,})(?:\n+|$)/,
+  html: /^ *(?:comment *(?:\n|\s*$)|closed *(?:\n{2,}|\s*$)|closing *(?:\n{2,}|\s*$))/,
+  item: /^( *)(bull) [^\n]*(?:\n(?!\1bull )[^\n]*)*/,
   lheading: /^([^\n]+)\n *(=|-){2,} *(?:\n+|$)/,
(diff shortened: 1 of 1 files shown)


lib/marked.js
```diff
@@ -47,39 +66,32 @@ block.list = edit(block.list)
   .replace('def', '\\n+(?=' + block.def.source + ')')
   .getRegex();
 
-block._tag = '(?!(?:'
```
styfle: Why was this using and underscore before? Was it supposed to be private?

joshbruce: Fair question. There were four underscored variables all together. I could not discern a rationale behind it...even if they were meant to be private. It looked like misunderstood convention to be honest.

The only members that need to be set outside the block object are the ones making calls to `edit()` and the like. All the rest can be within the object; however, there were three or four outside of the block object. Each of them had the underscore prefix...none of which appear to be referenced anywhere else.

lib/marked.js
```diff
@@ -8,35 +8,54 @@
 'use strict';
 
 /**
- * Block-Level Grammar
+ * Markdown block-level element rules
+ *
+ * QUESTION: What is the different between heading the lheading?
```
styfle: - heading is `# Heading` and `lheading

styfle: Also this type of question should not be in the source code.

joshbruce: @styfle: Think that's the crux of the documentation issue. Where do we put it. Initially I though describing the code in the contributing made the most sense because an outsider could read it before diving into the code. It was recommended to put it in the source code.

When I put on my "new guy" hat, these are the questions I'm asking when I look at the code. The question itself shouldn't remain in the source code once it's answered.

styfle: > It was recommended to put it in the source code.

lib/marked.js
```diff
@@ -47,39 +67,32 @@ block.list = edit(block.list)
...
-  .replace('tag', '<' + block._tag)
+  .replace('tag', '<' + block.tag)
   .getRegex();
 
+/**
+ * QUESTION: Why do tests require this be last?
+ */
 block.blockquote = edit(block.blockquote)
   .replace('paragraph', block.paragraph)
   .getRegex();
```
UziTech: `block.blockquote` needs to use the `block.paragraph` regex so it must be constructed after `block.paragraph`

lib/marked.js
```diff
@@ -8,35 +8,55 @@
...
+  title: /(?:"(?:\\"|[^"]|"[^"\n]*")*"|'\n?(?:[^'\n]+\n?)*'|\([^()]*\))/
 };
 
-block._label = /(?:\\[\[\]]|[^\[\]])+/;
-block._title = /(?:"(?:\\"|[^"]|"[^"\n]*")*"|'\n?(?:[^'\n]+\n?)*'|\([^()]*\))/;
+/**
+ * QUESTION: This seems like a bad construction. We define block.def as the result
+ *           of a function call. The edit() function takes a parameter named the same
+ *           as what we are defining. Why?
+ */
```
UziTech: The `block.def` defined above is really just the beginning of the construction and the rest is below.

Any other way I can think of doing it would require temp variables or repeated code.

lib/marked.js
```diff
@@ -89,7 +81,7 @@ block.normal = merge({}, block);
 /**
  * GFM Block Grammar
  *
- * GFM is an extension of the CommonMark.
+ * GFM use the CommonMark spec as a base, which it then extends.
```
styfle: `use` => `uses`

Feder1co5oave:
Rules are listed that way because that's the order blocks are parsed in. It represents their "priority".
Rules with a trailing underscore are not used to parse a whole element but are often replaced inside "parent" rules, or used later on in the code.

styfle:
@Feder1co5oave Do you know the purpose of the underscore for `block._tag`?
Is this supposed to indicate a private property?

UziTech:
Ya the underscore variables are just used to construct other regexes

joshbruce:
> Needless to say I consider these type of changes pointless and even detrimental to the codebase.

I disagree. Things that make the code easier to understand and read are not pointless. Having said that, it might cause unwanted behavior, which is why the test bed is so important. Going to close for now to give me a chance to focus on the GFM extensions and test suite. When I resubmit, it will probably be more about commenting the code to help new contributors understand the why behind certain decisions. Do you know if there is a reason the members are not placed within the object itself??

Feder1co5oave:
> Do you know if there is a reason the members are not placed within the object itself??

I believe they were declared close to where they get substituted into the bigger rules: https://github.com/markedjs/marked/blob/c118bc43652b4b490a7b06ce9f4163ce2e712cb2/lib/marked.js#L31-L36

PR #1176, opened 2018-03-27
Document mutability of tokens argument in parser

vsemozhetbyt:
If this behavior is intended, it may need to be documented to reduce possible unexpectedness.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -119,3 +119,39 @@ $ node
   { type: 'blockquote_end' },
   links: {} ]
 ```
+
+The Lexers build an array of tokens, which will be passed to their respective
+Parsers. The Parsers process each token in the token arrays,
+which are removed from the array of tokens:
+
+``` js
+const marked = require('marked');
+
+const md = `
+  # heading
+
+  [link][1]
+
+  [1]: #heading "heading"
+`;
+
+const tokens = marked.lexer(md);
+console.log(tokens);
+
+const html = marked.parser(tokens);
+console.log(html);
+
+console.log(tokens);
+```
+
+``` bash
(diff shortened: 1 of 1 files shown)


docs/USING_PRO.md
```diff
@@ -133,4 +133,39 @@ $ node
     text: 'i am using marked.' },
   { type: 'blockquote_end' },
   links: {} ]
-```
\ No newline at end of file
+```
+
+Note that the array of tokens returned by the lexer
+will be exhausted by the parser call:
```
joshbruce: What is meant by "exhausted" here? 

From the text I can presume that the lexer eventually contains an array of elements (there are two levers block and inline). From the title of the PR I can presume that I have the ability to access and modify this array of tokens. Is that true?

vsemozhetbyt: I mean: if you have an array of tokens from the lexer and you give it to the parser, its content will be empty: the parser will consume all the elements (shift them one by one?). I.e. the parser will not work with the copy of tokens array, it will mutate the argument,

Can you suggest more clear wording?

joshbruce: Hmmmmm...good question. If I'm understanding your interpretation correctly, maybe something like:

> Lexer builds an array of tokens, which will be passed to the Parser. The parser will process each token in the array, which will be removed from the array of tokens.

Thoughts? 

Think this honors the call order Lexer -> Parser. Maintains the mutability of the array of tokens. Maybe there's a place for something about block and inline lexer and parser?

vsemozhetbyt: Thank you! I've replaced the note and slightly rephrased the commit title.

If there are any more additions, please, let me know.

styfle:
@UziTech Any objections?

PR #1177, opened 2018-03-27
Remove node v4 from travis

styfle:
## Description

Node v4.x is just about to reach [end-of-life](https://github.com/nodejs/Release#release-schedule) so we really don't need to be running tests against it.

This should speed up CI just a little bit.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ jobs:
   include:
     - stage: spec tests 👩🏽‍💻
       node_js: v0.10
-    - node_js: v4
     - node_js: lts/*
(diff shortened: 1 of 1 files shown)


UziTech:
I think this would be a breaking change and would be better to wait for the 1.0 release. At that time we will be able to remove v4 and v0.10 and maybe add v6

styfle:
Agreed, I’m not sure what I was thinking🤔

I guess I have been removing v4 from other projects but I forgot the marked still supports v0.10 🤯

Closing 😇

PR #1183, opened 2018-03-28
add permalink and clear button

UziTech:
**Marked version:** 0.3.19

## Description

Adds a way to link to the demo with given text

image

The URL is also updated automagically 🎉 whenever the input changes 😃 

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/docs/demo/demo.js
+++ b/docs/demo/demo.js
@@ -10,6 +10,8 @@ if (!window.fetch) {
 var $inputElem = document.querySelector('#input');
 var $outputTypeElem = document.querySelector('#outputType');
 var $previewElem = document.querySelector('#preview');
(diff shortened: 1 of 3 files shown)


styfle:
Should the url update every time the input changes?
I think it should only update when they click the permalink button, [like this](https://www.typescriptlang.org/play/).
Otherwise, clicking the Back would cycle through each keystroke, right?

UziTech:
No, it uses [`history.replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_replaceState()_method) to replace the **current** state in the browser so the last state (which is accessed by using the back button) stays the same.

PR #1190, opened 2018-03-31
Disable heading IDs

joshbruce:
**Marked version:** 0.3.19

**Markdown flavor:** n/a

## Description

PR adds functionality to disable header IDs. 

This is also related to work being performed on #1160.

For some reason testing options does not work in isolation.

## Contributor

## Committer

- [ ] Merge PR


styfle:
You should also add this to the documentation so users are aware of this feature.

UziTech:
You should always create new branches from master to avoid including changes from other PRs

joshbruce:
@styfle: Updated documentation. Maybe I should make that a separate PR given the review from @UziTech.

@UziTech: Believe it or not, I actually did branch off from `master`; I just did the same `--save-dev` from the other PR. Our test harness uses `min`; therefore, to get running tests I'm pretty sure there has to be updates to that file as well. Will put together some other PRs I think might cover the comments...sorry for the multiples and possible closings.

joshbruce:
See #1192 and #1193

joshbruce:
@UziTech: Think I follow what you're asking, though not sure if I could that easily. I'm adding this functionality to test the headers for CommonMark as none of the examples use IDs; so, the capability will be in the integration test suite at some point.

Will see what I can do on the more unit-test side of the house.

UziTech:
Ya that is fine if this option is used in the integration tests, but there is no need to have an integration test specifically for this feature.

The way I see integration tests is if a feature needs to test multiple parts of the software (and possibly the platform) than it should have an integration test (e.g. spec tests need to test the whole application; lexing, parsing, and rendering)

This feature only affects the rendering so it could be a unit test something like:

```js
it('should add header ids by default', function () {
  var rederer = new marked.Renderer(marked.defaults);
  var header = rederer.heading('test', 1, 'test');
  expect(header).toBe('<h1 id="test">test</h1>\n');
});

it('should ignore header ids', function () {
  var rederer = new marked.Renderer({ headerIds: false });
  var header = rederer.heading('test', 1, 'test');
  expect(header).toBe('<h1>test</h1>\n');
});
```

joshbruce:
I believe we've covered all the comments and a little beyond.

1. Uses unit tests instead of integration tests.
2. Documentation updated (including some things that were missing and ordering them alphabetically in the code and the docs).
3. Contributor checklist updated.

If we merge this, I will close #1192 and #1193. Then I should be able to go back to #1160 to handle the failing header things, which should have only things that are actually broken left.

package.json
```diff
@@ -31,7 +31,9 @@
     "eslint-plugin-standard": "^3.0.1",
     "front-matter": "^2.3.0",
     "glob-to-regexp": "0.3.0",
+    "html-differ": "^1.3.4",
     "jasmine": "^3.1.0",
+    "jasmine2-custom-message": "^0.9.0",
```
UziTech: these are not needed for this PR

test/unit/marked-spec.js
```diff
@@ -1,7 +1,21 @@
...
-  marked.parse('Hello World!');
-  expect(marked.parse).toHaveBeenCalled();
+  spyOn(sut, 'parse').and.callThrough();
+  sut.parse('Hello World!');
+  expect(sut.parse).toHaveBeenCalled();
+});
+
+describe('Test heading ID functionality', function() {
+	it('should add id attribute by default', function() {
+		var renderer = new sut.Renderer(sut.defaults);
```
UziTech: Could we change `sut` to something like `markedUnderTest` or something like that? It feels very unintuitive that these defaults are coming from marked.

I feel like the only reason to distinguish the "system under test" is if we would need the same system **not** under test. (e.g. Jasmine testing Jasmine with Jasmine) I can't think of a reason we would ever need that.

test/integration/marked-spec.js
```diff
@@ -1,4 +1,7 @@
 var marked = require('../../marked.min.js');
+var HtmlDiffer = require('html-differ').HtmlDiffer,
+    htmlDiffer = new HtmlDiffer();
+var since = require('jasmine2-custom-message');
```
UziTech: These are not needed

lib/marked.js
```diff
@@ -1342,22 +1346,23 @@ marked.setOptions = function(opt) {
 };
 
 marked.defaults = {
-  gfm: true,
-  tables: true,
+  baseUrl: null,
   breaks: false,
+  gfm: true,
+  headerPrefix: '',
+  headerIds: true,
```
UziTech: 'I' comes before 'P'

joshbruce: Good catch...funny enough, I got it right in the table. :)

package.json
```diff
@@ -23,19 +23,19 @@
     "html"
   ],
   "devDependencies": {
-    "eslint": "^4.15.0",
-    "eslint-config-standard": "^11.0.0-beta.0",
-    "eslint-plugin-import": "^2.8.0",
+    "eslint": "^4.19.1",
```
styfle: Why are you updating devDependencies?

styfle: This should be in a different PR in my opinion.

package.json
```diff
@@ -23,19 +23,19 @@
...
     "eslint-plugin-standard": "^3.0.1",
     "front-matter": "^2.3.0",
     "glob-to-regexp": "0.3.0",
     "jasmine": "^3.1.0",
     "markdown": "*",
-    "markdown-it": "*",
-    "showdown": "*",
-    "uglify-js": "^3.3.10"
+    "markdown-it": "^8.4.1",
+    "showdown": "^1.8.6",
```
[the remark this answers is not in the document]
joshbruce: Not intentional.

Will try to reset manually.

screen shot 2018-04-03 at 12 02 41 pm

package-lock.json
```diff
@@ -11,9 +11,9 @@
       "dev": true
     },
     "acorn": {
-      "version": "5.4.1",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.4.1.tgz",
-      "integrity": "sha512-XLmq3H/BVvW6/GbxKryGxWORz1ebilSsUDlyC27bXhWGWAZWkGwS6FLHjOlwFXNFoWFQEO/Df4u0YYd0K3BQgQ==",
+      "version": "5.5.3",
```
styfle: Can you revert this file too?

PR #1200, opened 2018-04-04
Fix: TextRenerer.paragraph is not a function

FlorianKoerner:
## Description

The method `paragraph` is missing in the TextRenderer.

## Result

```
TypeError: this.renderer.paragraph is not a function
```

## What was attempted

```
marked('Test', {
    renderer: new marked.TextRenderer()
})
```

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -967,6 +967,10 @@ TextRenderer.prototype.br = function() {
   return '';
 }
 
+TextRenderer.prototype.paragraph = function(text) {
+  return text;
+}
+
 /**
  * Parsing & Compiling
  */


UziTech:
@FlorianKoerner what was the purpose of setting the TextRenderer as the Renderer in marked?

FlorianKoerner:
Originally my intention was to parse only inline blocks. Meanwhile I know that the TextRenderer is the wrong tool for this.

Is there an easy way to render only inline blocks?

joshbruce:
See lines 856 to 858:

```js
Renderer.prototype.paragraph = function(text) {
  return '<p>' + text + '</p>\n';
};
```

Is this what you were trying to access @FlorianKoerner and received a bad function call reference??

joshbruce:
Not that I can tell given the way the renderer is described at the moment. Having said that, it might make sense in the future. We have inline and block flavors of the lexer and parser...and almost the renderer.

joshbruce:
Tagging with #746 something to potentially introduce in 1.x or 2.x

PR #1203, opened 2018-04-04
use default options in Renderer by default

UziTech:
**Marked version:** 0.3.19

## Description

- Sets the Renderer options to the default options by default

I'm not sure if there is a reason the Renderer options were set to an empty object instead of the default options. The Parser and Lexers are set to the default options already

```js
function Renderer(options) {
  this.options = options || {};
  ...
```

```js
function Lexer(options) {
  this.options = options || marked.defaults;
  ...
```

```js
function InlineLexer(links, options) {
  this.options = options || marked.defaults;
  ...
```

```js
function Parser(options) {
  this.options = options || marked.defaults;
  ...
```

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -788,7 +788,7 @@ InlineLexer.prototype.mangle = function(text) {
  */
 
 function Renderer(options) {
-  this.options = options || {};
+  this.options = options || marked.defaults;
 }
 
 Renderer.prototype.code = function(code, lang, escaped) {


UziTech:
@Feder1co5oave is there any reason Renderer options would be set to an empty object instead of the `marked.defaults`?

UziTech:
Yes it would be a breaking change but I feel like it would be a fix. Right now that would be the same as the following:

```js
var r = new Renderer({
  baseUrl: undefined,
  breaks: undefined,
  gfm: undefined,
  headerIds: undefined,
  headerPrefix: undefined,
  highlight: undefined,
  langPrefix: undefined,
  mangle: undefined,
  pedantic: undefined,
  renderer: undefined,
  sanitize: undefined,
  sanitizer: undefined,
  silent: undefined,
  smartLists: undefined,
  smartypants: undefined,
  tables: undefined,
  xhtml: undefined
});
```

which I don't think anyone would do willingly. So it would be more likely that anyone using the renderer would pass some options so this wouldn't break anything.

joshbruce:
ps. Just because this is a breaking change doesn't stop me from merging this. If neither @UziTech or @styfle are concerned committing to a 0.4.x as the next release, I would say we are good to go.

UziTech:
Ya I would say we should merge it as a bug fix and release when 0.4.x milestone is ready. I was looking through the blame and it looks like it was just missed when the renderer was created.

styfle:
Done

Feder1co5oave:
Because of lines [549](https://github.com/UziTech/marked/blob/a19e2b631a06466d45b6b80f20ed7f105ec84907/lib/marked.js#L549) and [980](https://github.com/UziTech/marked/blob/a19e2b631a06466d45b6b80f20ed7f105ec84907/lib/marked.js#L980). It gets overwritten anyways,

PR #1210, opened 2018-04-09
GitHub Flavored Markdown test bed

joshbruce:
**Marked version:** 0.3.19

## Description

Adds GitHub Flavored Markdown to test bed.

It is our understanding that GFM is basically just CommonMark with some extensions. Therefore, this only adds the extensions to the overarching test suite.

## Contributor

## Committer

- [ ] Merge PR


joshbruce:
@UziTech: Would also like to start shifting to the separate folders concept but don't know where to begin; think it's just easier to ask and learn or if you would be willing to take it on.

With these additional tests we should have a more trustworthy test bed overall...specifically one that protects itself against regression. Would like to get #1135 using it before merging that in, unless you think there's a good reason not to.

UziTech:
where are you getting the examples for github_flavored_markdown.json?

joshbruce:
@UziTech: That one I had to build, unfortunately. 

https://github.github.com/gfm/ - I went to each of the TOC links that said "extension" - and copied and pasted.

styfle:
Here's the plain text version: https://github.github.com/gfm/spec.txt

Feder1co5oave:
You can run [this](https://github.com/commonmark/CommonMark/blob/master/test/spec_tests.py) on [this](https://github.github.com/gfm/spec.txt), like [this](https://github.com/commonmark/CommonMark/blob/master/Makefile#L20), and then filter it in some way.

UziTech:
1. https://stackoverflow.com/a/12299717/806777
2. No, the sanitize set in beforeEach in #1213 is just to reset the defaults after the original tests. once we merge #1212 it will be removed. I just added it in so the tests would pass.

PR #1212, opened 2018-04-10
Reset defaults before tests

UziTech:
**Marked version:** 0.3.19

## Description

- add function `marked.getDefaults()` that gets assigned to `marked.defaults`
- resets the defaults before each test

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1346,7 +1346,7 @@ marked.setOptions = function(opt) {
 };
 
 marked.getDefaults = function () {
-  var defaults = {
+  return {
     baseUrl: null,
     breaks: false,
     gfm: true,
@@ -1365,13 +1365,6 @@ marked.getDefaults = function () {
     tables: true,
     xhtml: false
   };
-
-  var clone = {};
-  for (var opt in defaults) {
-    clone[opt] = defaults[opt];
-  }
-
-  return clone;
 }
 
 marked.defaults = marked.getDefaults();
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1345,30 +1345,36 @@ marked.setOptions = function(opt) {
   return marked;
 };
 
-marked.defaults = {
-  baseUrl: null,
-  breaks: false,
-  gfm: true,
-  headerIds: true,
-  headerPrefix: '',
-  highlight: null,
-  langPrefix: 'lang-',
-  mangle: true,
-  pedantic: false,
-  renderer: new Renderer(),
-  sanitize: false,
-  sanitizer: null,
-  silent: false,
-  smartLists: false,
-  smartypants: false,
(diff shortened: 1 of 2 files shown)


UziTech:
good call

lib/marked.js
```diff
@@ -1345,25 +1345,36 @@ marked.setOptions = function(opt) {
...
+    tables: true,
+    xhtml: false
+  };
+
+  var clone = {};
+  for (var opt in defaults) {
+    clone[opt] = defaults[opt];
+  }
+
+  return clone;
```
styfle: Is the clone still necessary? Can't you `return defaults;` here instead?

joshbruce:
I ran into problems using `setOptions`, that's why we switched to making them part of the call. I was thinking we would want to do something like:

```js
  beforeEach(function() {
    marked.defaults = {
      headerIds: false,
      xhtml: true
    }
  });

  afterEach(function() {
    marked.defaults = {
      headerIds: true,
      xhtml: false
    }
  });
```

The CommonMark spec doesn't recognize header IDs and uses XHTML for some of the tests.

lib/marked.js
```diff
@@ -1345,25 +1345,36 @@ marked.setOptions = function(opt) {
...
+    sanitize: false,
+    sanitizer: null,
+    silent: false,
+    smartLists: false,
+    smartypants: false,
+    tables: true,
+    xhtml: false
+  };
+
+  var clone = {};
```
joshbruce: Is it possible to do this as part of the test itself instead of the lib? Believe @styfle mentioned something like that before. Thinking it would be better to have the test use a blank object over the singleton pattern that seems preferred by Node (understandably in many use cases).

lib/marked.js
UziTech: I see what you are saying

UziTech:
@joshbruce We could do something like that in the tests but then next time we add a new test file we will have to remember to do the same thing or we will fall into the same trap. This way we never have to think about it again.

joshbruce:
@UziTech: Agreed. Think that's on us as reviewers. Not sure how often we'll be adding new test suites at this juncture, just seems the more logical place and, if we can do it in a single place, then we should be good.

joshbruce:
@UziTech: PS. good catch on the not returning thing. I mean to ask about that.

UziTech:
Yes, `marked.getDefaults()` will always return *our* default options. `marked.defaults` is still overwritten by `marked.setOptions` (which I think we should change in 1.x)

We are resetting `marked.defaults` to `marked.getDefaults()` in a `beforeEach` that will run before any other `beforeEach` so we can still set options for tests in a `beforeEach` and it will work as expected.

Issue #1218, opened 2018-04-15
Code blocks with email-like text are no longer parsed as code


kornelski:
**Marked version:** 0.3.12 (OK) 0.3.14 (buggy)

```markdown
`user@host`
```

used to render as `<code>` block:

`user@host`

but now it's rendered as an e-mail link with literal `` ` `` in it:

[\`user@host\`](mailto:user@host)

UziTech:
seems to work in the latest 0.3.19: [demo](https://marked.js.org/demo/?outputType=html&text=%60user%40host%60)

kornelski:
Try

```

 `/users?email=joe@smith.com`.

```

UziTech:
Yup ` `user@host.com` ` does it

joshbruce:
Looking at the specs to try and figure out a test to create and they don't seem to call out this specific use case.

For CommonMark:

> Autolinks are absolute URIs and email addresses inside < and >.

So, this would be something related to GitHub Flavored Markdown:

> An extended email autolink will be recognised when an email address is recognised within any text node. 

Looking at code spans (examples 322 to 327) I would suspect that the proper behavior would be something like:

```
Markdown:

`user@example.com`

Html:

<code>user@example.com</code>
```

@kornelski: Is that what you're looking for?

joshbruce:
`someone@example.com`

kornelski:
In my case `<code>` is more important than `<a>`, because I'm using a made-up e-mail address in documentation and don't want it linked. If you prefer links, then both (`<code><a>email</a></code>`) would be a reasonable rendering IMHO.

UziTech:
IMP creating a link without `<>` surrounding it should only happen when the gfm option is true since this would not be compliant with commonmark

joshbruce:
GFM complies with CommonMark; so, autolinking should be invoked by the `<>` combination. Given what I can discern from the code span area of CM and GFM, I would say the code span leaves everything inside alone. The linked PR runs a test that would result in that ignore.

UziTech:
in order to pass [CommonMark Example #583](https://spec.commonmark.org/0.28/#example-583) This autolinking without `<>` will have to be moved to `gfm: true`

joshbruce:
Think this conversation might go beyond this issue...deeper issues with Marked at the moment...making a new issue:

joshbruce:
[...]

PR #1219, opened 2018-04-15
Test for #1218 & "marked" spec

joshbruce:
**Marked version:** 0.3.19

## Description

Adds test case defining the behavior mentioned in #1218 to ensure we have a test case for it. After looking at the specs, the question does stand, I think, on what the desired output should be when using GFM.

Note: This PR does not submit a solution; only adds to the growing test bed.

## Expectation

```
<p><code>someone@example.com</code></p>
```

## Result

```
<p><a href="mailto:%60someone@example.com">`someone@example.com</a>`</p>
```

## What was attempted

```
`someone@example.com`
```

## Contributor

## Committer

- [ ] Merge PR


UziTech:
Could we move all of the original tests here?

joshbruce:
@UziTech: Great minds. Once we get this merged in I was going to go ahead and start putting that PR together. One that takes into account that we are doing the CM and GFM comparison separately while trying to maintain backward compatibility awareness.

joshbruce:
To @UziTech's point, I will see what I can do to put messenger somewhere else...maybe set the options in the spec and have messenger just use whatever is there. Thinking I'll put it in the root of the spec directory - gives me a chance to rename it.

Issue #1201, opened 2018-04-04
Thoughts about using an eslint plugin for REDOS detection?


davisjam:
This is a question for discussion, not an issue.

- I am working on a [vuln-regex-detector](https://www.npmjs.com/package/vuln-regex-detector) module.
- I will be creating an eslint plugin based on this module in the next day or two. It will make one synchronous HTTP query for each regex it finds in the source code. Queries go to a server I host at Virginia Tech which has a database mapping regexes to pre-computed "is vulnerable?" answers. The server is running the server side of the code in [this project](https://github.com/davisjam/vuln-regex-detector).

Once it's ready, does anyone have concerns with adding my REDOS plugin to this project's eslint config?

joshbruce:
Tagging @UziTech

UziTech:
It would be really nice to get some sort of ReDoS test on travis. Could we make it so the server side code runs on travis instead of making http requests?

davisjam:
@UziTech 
>  Could we make it so the server side code runs on travis instead of making http requests?

Testing a regex is expensive (potentially 60 seconds+ per regex) so memoizing them is helpful. Then that memoized result has to live somewhere.

My impression (could be wrong) is that Travis isn't a good place for permanent storage.
Is there something wrong with HTTP requests?

UziTech:
> Is there something wrong with HTTP requests?

In my experience http requests from travis seem to be the least reliable part of travis and "one synchronous HTTP query for each regex" could be a lot of http requests for every pr.

Could we download the memoized results and check the regexes ourselves or send a bunch of regexes at once to limit the number of http requests?

davisjam:
> Could we download the memoized results and check the regexes ourselves?

100MB+, not a good option.

> or send a bunch of regexes at once to limit the number of http requests?

I don't think you can batch nicely with an eslint plugin. The "is vulnerable?" function gets called for each regex node in the AST and doesn't have a bigger sense of the context.

If you want to set up a post-eslint "security scanner" that runs as part of `npm run test` or `npm run lint`, you could do batching that way.

> Could be a lot of http requests for every PR

About 100 requests.

```bash
grep '/.*/' lib/marked.js | grep -v '\/\/' | wc -l
98
```

UziTech:
I would rather it be separate from eslint so we could run tests, run redos tests, then run lint. I feel like that would make it easier to see what failed in travis.

davisjam:
[...]

PR #1220, opened 2018-04-15
test: security scan

davisjam:
- `npm run test:redos` now scans for REDOS issues
- added a Travis stage for 'security scan', which at the moment just checks for REDOS

Fixes: #1201.


davisjam:
Not sure whether the CI stage is configured properly. I don't think CI changes take effect until after it's merged though, so someone should check it over.

davisjam:
Actually it seems the Travis run is picking up the security scan, [here](https://travis-ci.org/markedjs/marked/jobs/366906356). Results are as expected.

The security scan took 60 seconds, of which running the eslint plugin took 13 seconds. So I don't think querying my REDOS DB will be problematic.

UziTech:
I still don't think the eslint plugin is all that helpful in our situation because of the way we construct the regexes.

davisjam:
As you've pointed out earlier, a static analysis won't guarantee protection. I disagree, however, that it "is [not] all that helpful". It has already identified several REDOS issues, and it is certainly much better than nothing. I therefore suggest we adopt it.

We can open an issue to revisit this with a more thorough approach. Exposing the module's regexes and invoking tests on them from the test suite would be a nice extension.

styfle:
@davisjam CI is showing a vulnerability.

```
557:14  error  Regex pattern vulnerable to catastrophic backtracking: ^!?\[(label)\]\(\s*<?([\s\S]*?)>?(?:\s+(['"][\s\S]*?['"]))?\s*\)  vuln-regex-detector/no-vuln-regex
```

Is this a false positive? If not, can you fix it?

davisjam:
> Is this a false positive?

This scan will not produce false positives. This tool only flags a regex if it has demonstrated super-linear behavior against it. It will have some false negatives, e.g. it did not catch #1224 .

joshbruce:
I like the idea of CI covering us on security. Not going to require this be there for the 0.4.0 release as it seems like there might be some more conversation desired.

styfle:
@davisjam Want to rebase on master?

.travis.yml
```diff
@@ -13,6 +13,10 @@ jobs:
     - node_js: lts/*
     - node_js: node
 
+    - stage: security scan
```
UziTech: security scan 🔐

davisjam: Where do you guys find these cool symbols?

styfle: @davisjam If you type `:` you can start typing words and find emoji 🔍 

Otherwise, checkout https://emojipedia.org 😄

Issue #1222, opened 2018-04-15
security: vulnerable regex


davisjam:
```
(18:57:29) jamie@woody ~/Desktop/floss/marked $ npm run test:redos

> marked@0.3.19 test:redos /home/jamie/Desktop/floss/marked
> eslint --plugin vuln-regex-detector --rule '"vuln-regex-detector/no-vuln-regex": 2' lib/marked.js


/home/jamie/Desktop/floss/marked/lib/marked.js
  557:14  error  Regex pattern vulnerable to catastrophic backtracking: ^!?\[(label)\]\(\s*<?([\s\S]*?)>?(?:\s+(['"][\s\S]*?['"]))?\s*\)  vuln-regex-detector/no-vuln-regex
```

UziTech:
out of curiosity what is the part of this regex that is vulnerable?

davisjam:
The issue is the quantified adjacent whitespaces here: `\s*<?[\s\S]*>?\s+`. The angle brackets are optional so input including a ton of whitespace can be problematic.

UziTech:
wouldn't some thing like this solve it?

```js
...\s*(<[\s\S]*?>|[^\s]+)(?:\s+...
```

no more quantified adjacent whitespaces

UziTech:
and the angle brackets are already being taken out of the url by [this line](https://github.com/markedjs/marked/blob/master/lib/marked.js#L765)

davisjam:
`...\s*(<[\s\S]*?>|[^\s]+)(?:\s+...`

This would not solve it, just change the form it took. On malicious input of the form `<<...<` it would have to decide between the left or right node in the disjunction. Backtracking manifests around 10K `<`s and gets worse as you increase.

PR #1223, opened 2018-04-15
security: replace vulnerable regex with parser

davisjam:
Problem: link regex was vulnerable
Solution: dedicated parser

Fixes: #1222

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -554,9 +554,48 @@ inline.normal = merge({}, inline);
 inline.pedantic = merge({}, inline.normal, {
   strong: /^__(?=\S)([\s\S]*?\S)__(?!_)|^\*\*(?=\S)([\s\S]*?\S)\*\*(?!\*)/,
   em: /^_(?=\S)([\s\S]*?\S)_(?!_)|^\*(?=\S)([\s\S]*?\S)\*(?!\*)/,
-  link: edit(/^!?\[(label)\]\(\s*<?([\s\S]*?)>?(?:\s+(['"][\s\S]*?['"]))?\s*\)/)
-    .replace('label', inline._label)
-    .getRegex(),
+  link: {
+    exec: (s) => {
+      // [TEXT](DESTINATION)
+      var generalLinkRe = edit(/^!?\[(label)\]\((.*?)\)/)
+        .replace('label', inline._label)
+        .getRegex();
+
+      function unwrapCarats (str) {
+        if (str.match(/^<.*>$/)) {
+          str = str.substr(1, str.length - 1);
+        }
+        return str;
+      }
+
+      var fullMatch = generalLinkRe.exec(s);
+      if (fullMatch) {
+        var text = fullMatch[1];
+        var destination = fullMatch[2];
+
+        var m;
+
+        var destinationAndTitleRe = /^([^'"(]*[^\s])\s+(['"(].*['")])/;
+        if (m = destinationAndTitleRe.exec(destination)) {
+          // <destination> -> destination
+          var dest1 = m[1].trim();
+          dest1 = unwrapCarats(dest1);
+          var title1 = m[2];
+          return [fullMatch[0], text, dest1, title1];
+        }
+
+        var destinationRe = /^(<?[\s\S]*>?)/;
+        if (m = destinationRe.exec(destination)) {
+          // <destination> -> destination
+          var dest2 = m[1].trim();
+          destination = unwrapCarats(dest2);
+          var title2 = '';
+          return [fullMatch[0], text, dest2, title2];
+        }
+      }
+      return null;
+    }
+  },
   reflink: edit(/^!?\[(label)\]\s*\[([^\]]*)\]/)
     .replace('label', inline._label)
     .getRegex()


davisjam:
If this is merged, a new CI on #1220 should come back clean.

UziTech:
I feel like this complicates the code a lot. Is there no way we can do this with a constructed regex?

lib/marked.js
```diff
@@ -554,9 +554,48 @@ inline.normal = merge({}, inline);
...
-    .replace('label', inline._label)
-    .getRegex(),
+  link: {
+    exec: function (s) {
+      // [TEXT](DESTINATION)
+      var generalLinkRe = edit(/^!?\[(label)\]\((.*?)\)/)
+        .replace('label', inline._label)
+        .getRegex();
+
+      function unwrapCarats (str) {
```
joshbruce: Carets?? Carat and carrot are different. :)

Angle brackets might be most appropriate if I'm reading the regex correctly.

davisjam: Haha I was wondering if I was spelling that right. I'll switch to AngleBrackets anyway.

lib/marked.js
```diff
@@ -554,9 +554,48 @@ inline.normal = merge({}, inline);
...
+          dest1 = unwrapCarats(dest1);
+          var title1 = m[2];
+          return [fullMatch[0], text, dest1, title1];
+        }
+
+        var destinationRe = /^(<?[\s\S]*>?)/;
+        if (m = destinationRe.exec(destination)) {
+          // <destination> -> destination
+          var dest2 = m[1].trim();
+          destination = unwrapCarats(dest2);
```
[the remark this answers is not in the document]
davisjam: Nope. Will fix.

lib/marked.js
```diff
@@ -554,9 +554,48 @@ inline.normal = merge({}, inline);
 inline.pedantic = merge({}, inline.normal, {
   strong: /^__(?=\S)([\s\S]*?\S)__(?!_)|^\*\*(?=\S)([\s\S]*?\S)\*\*(?!\*)/,
   em: /^_(?=\S)([\s\S]*?\S)_(?!_)|^\*(?=\S)([\s\S]*?\S)\*(?!\*)/,
-  link: edit(/^!?\[(label)\]\(\s*<?([\s\S]*?)>?(?:\s+(['"][\s\S]*?['"]))?\s*\)/)
-    .replace('label', inline._label)
-    .getRegex(),
+  link: {
+    exec: function (s) {
```
joshbruce: Might worth adding some doc blocks to introduce the why behind some of this...nothing too major, just to help those new to the code.

davisjam:
@styfle @joshbruce fbf93a8 addresses your comments.

lib/marked.js
```diff
@@ -554,9 +554,51 @@ inline.normal = merge({}, inline);
...
+        if (m = destinationAndTitleRe.exec(destination)) {
+          // <destination> -> destination
+          var dest1 = m[1].trim();
+          dest1 = unwrapAngleBrackets(dest1);
+          var title1 = m[2];
+          return [fullMatch[0], text, dest1, title1];
+        }
+
+        var destinationRe = /^(<?[\s\S]*>?)/;
+        if (m = destinationRe.exec(destination)) {
```
styfle: The two `if blocks` are nearly identical. Can you make a functions for those?
```

Comment (joshbruce) [reply lib/marked.js]: Agree, but not a deal breaker for my review.

Comment (davisjam) [reply lib/marked.js]: @styfle See ba2fc13

File: lib/marked.js [hunk 5]
```diff
@@ -554,9 +554,72 @@ inline.normal = merge({}, inline);
...
+        for (var i = 0; i < parsingRegexes.length; i++) {
+          match = parsingRegexes[i].exec(destination);
+          if (match) {
+            dest = match[1];
+            title = match[2];
+            break;
+          }
+        }
+
+        if (match) {
```
styfle: Can you flip this so that the if statement is smaller:

```

File: lib/marked.js [hunk 6]
```diff
@@ -554,9 +554,72 @@ inline.normal = merge({}, inline);
...
+          dest = dest.trim();
+          dest = unwrapAngleBrackets(dest);
+
+          return [dest, title];
+        }
+        return null;
+      }
+
+      var fullMatch = generalLinkRe.exec(s);
+      if (fullMatch) {
```
styfle: Can you flip this so that the if statement is smaller:

```
if (!fullMatch) {
  return null;
}
// ... split and such here
return [fullMatch[0], text, destinationAndTitle[0], destinationAndTitle[1]];
```

lib/marked.js
```diff
@@ -554,9 +554,72 @@ inline.normal = merge({}, inline);
...
+        var parsingRegexes = [destinationAndTitleRe, destinationRe];
+
+        var match = false;
+        var dest = undefined;
+        var title = undefined;
+        for (var i = 0; i < parsingRegexes.length; i++) {
+          match = parsingRegexes[i].exec(destination);
+          if (match) {
+            dest = match[1];
+            title = match[2];
```
styfle: You don't need to assign dest and title here. Simply use `match` below.

davisjam:
@styfle Addressed your comments in 47f4388, thank you.

styfle:
@UziTech Can you take a look at the code now? Any other reservations?

UziTech:
Not sure. That seems to be happening a lot lately. Maybe a travis bug?

joshbruce:
@UziTech: Good point. I think parsers like this should be done as a last resort unless we can demonstrate through benchmarking that there isn't a big performance difference...I don't think we should get in the position of asserting something is more performant in all cases.

The CI thing is weird. Is it possible that the key thing is causing issues?

styfle:
Looks like this failed lint: https://travis-ci.org/markedjs/marked/builds/367190715

@davisjam Can you try running lint and tests on your machine and fix any issues?

davisjam:
Lint passes now, that CI was on a previous version.

styfle:
Excellent! ✨ 
Merged and CI looks good on `master` ✅

https://travis-ci.org/markedjs/marked/builds/367431216

UziTech:
looks like the `unwrapAngleBrackets` is done twice:
https://github.com/markedjs/marked/blob/master/lib/marked.js#L765

shouldn't this type of processing be done in the output function?

PR #1224, opened 2018-04-16
security: fix unsafe heading regex

davisjam:
Problem:
REDOS could be triggered through exploitation of the 'heading' regex.

Solution:
Refactor regex. It matches the same language as before but
is less vulnerable to REDOS.
It is now safe using the bounds suggested by those disclosing it.
It remains super-linear but a successful exploit requires
a much longer attack string.

Fixes:
Issue disclosed privately.

Credit:
This issue was pointed out by
Nick Starke (@nstarke) and Adam Cazzolla (@atcazzual) of Sonatype Security research.


styfle:
Thanks! 👍

Issue #1123, opened 2018-03-04
User analysis


joshbruce:
**Marked version:** 0.3.17

**Markdown flavor:** n/a

**Proposal type:** other

## What pain point are you perceiving?

Who are they? Seriously. Historically, based on the PRs, I think this is what I'm seeing from my UX world.

1. Polyglot: This could cause a lot of communication difficulties; therefore, patience is critical.
  - The header id issue mainly seems to affect Asian character localizations.
  - GitHub user handles and names point toward users in Russia and Italy.
  - Primarily the traffic on GitHub is in English.
2. Didn't ask for this (just example personas):
  - My name's Jim, I'm a retired software engineer from way back. A friend of mine asked me to take over their site. I've been out of the game for a while now and wanted something to do on the weekends to keep my mind fresh and it seemed easy enough.
  - My name's Adeline, I recently graduated from a developer bootcamp and am at my first job. I'm trying to maintain a legacy site. I didn't even know marked existed.
  - My name is Jhumpa, I'm a recent college graduate. One of our projects was to create an NPM package and make it open source. I created a markdown editor that uses marked. Now I have a full-time job and can't really maintain it and its dependencies.

Experiment:

1. Close 60% of issues and PRs.
2. See who responds.
3. What they're response was.

Results:

(I didn't spend the time to keep track, this is one of those after the experiment to try a different experiment deals.) 

- ~80% never responded. 
- ~15% were pretty fine with it, they had walked away from marked long ago. 
- ~4% were a little disappointed but understanding of where marked was and was trying to go.
- ~1% were very upset (only one actually took action to submit a proposal PR but ran into some trouble with understanding the test harness)

## What solution are you suggesting?

- [ ] Create form letter to dependents
- [ ] Committers go to their dependents who depend on marked and post an issue containing the form letter (see who responds).
- [ ] Once the dust settles, determine if further experiments are required.

UziTech:
We could automate the process. Get a list of dependents from npm, use GitHub API to add an issue with the form, see who replies.

joshbruce:
I like. Maybe we should scale back the initial requirement??

Just see what the version declarations are in `package.json`. Then target the ones with latest (*) - hopefully it's the smallest sample returned...thinking they're the ones who need to know the most.

joshbruce:
Maybe a sub-issue of this epic assigned to the "Improved product scope" project...??

joshbruce:
[...]

PR #587, opened 2015-05-06
Github Task Lists

jhollingworth:
Add support for Github Task Lists under the gfm flag.

Changes to API
- list(_string_ body, _boolean_ ordered, _boolean_ taskList)
- listitem(_string_ text, [_boolean_ checked]).

`checked` is defined when you have a list item which starts with `[ ]` or
`[x]`.If defined its a boolean depending on whether the `x` is
present. When checked is defined we add a input type type `checkbox` to
the list item and add the class `task-list-item-checkbox`.

`taskList` is true if a list has any list items where `checked` is
defined. When true we add the class `task-list` to the list.

Resolves chjj/marked#107

PR #1226, opened 2018-04-16
security: finish fixing unsafe heading regex

davisjam:
Apply similar patch to a similar heading regex.
Follow-on to f052a2c04 (#1224).

Test: Add a test case to demonstrate the slower blow-up.


UziTech:
Looks like the tests are failing because the new test takes more than a second to pass. Is the test supposed to be that slow?

styfle:
> Test: Add a test case to demonstrate the slower blow-up.

You successfully demonstrated the blow-up 😃 
What was the time to run the test before the change?

UziTech:
We can do some parsing in Lexer.prototype.token

https://github.com/markedjs/marked/blob/4711f6bb51c538b3693e35e8091cd55bdf23a934/lib/marked.js#L235-L244

davisjam:
@UziTech Acknowledged, in the style of #1227?

UziTech:
Ya similar to #1227.

changing `' *([^\n]+?) *'` to `'(.+?)'` and trimming at `text: cap[2].trim()` seems to speed it up considerably

joshbruce:
@styfle: I'm starting to question the futility of making the 2 approver thing governed by the machine. It doesn't give users - nor us - a green light until the approvals are in place. I liked seeing the all green (or red) and knowing pretty much out of hand whether it was merge conflicts, Travis, and so on.

Maybe we can keep it a "gentleman's agreement" - we've been kinda doing the two approver thing anyway...TOS it in a "Committing" document...what it is to have the mantle of committer on the project.

styfle:
>  It doesn't give users - nor us - a green light until the approvals are in place

Correct, meaning it can be merged until all the policies are satisfied.
You can see which ones are blocking by expanding the checks.
All green means you can click the Merge button.
You can approve at any time but typically it would make sense to approve after CI is passing.

joshbruce:
@styfle: Would it show merge conflicts before having all the other prerequisites??

styfle:
Yes, see #587

And note that the checks looks a little different on mobile than on the desktop.

lib/marked.js
```diff
@@ -16,7 +16,8 @@ var block = {
   code: /^( {4}[^\n]+\n*)+/,
   fences: noop,
   hr: /^ {0,3}((?:- *){3,}|(?:_ *){3,}|(?:\* *){3,})(?:\n+|$)/,
-  heading: /^ *(#{1,6}) *([^\n]+?) *(?:#+ *)?(?:\n+|$)/,
+  // cap[2] might be ' HEADING # ' and must be trimmed appropriately.
+  heading: /^ *(#{1,6})([^\n]*)(?:\n+|$)/,
```
UziTech: could we replace `[^\n]` with `.`?

davisjam: Yes haha

davisjam: Worth noting that this regex is substituted into `block.paragraph` and the follow-up parsing will not be performed there. I think we're OK spec-wise but might want to double-check.

davisjam:
FWIW Here's a Node-only way to deal with all REDOS issues:

```
const vm = require('vm');
const util = require('util');

const marked = require('./lib/marked.js');

const myContext = {
  marked: (text) => {
    return marked(text);
  }
};

function safeMarked(md) {
  let result;
  try {
    result = vm.runInNewContext(`marked("${md}")`, myContext, {
      timeout: 20
    });
  } catch (e) {
    util.log(e);
    result = 'TIMEOUT';
  }

  return result;
}
```

styfle:
@davisjam I would say a new module. In fact, this is small enough to just be in the docs as BEST_PRACTICES.md

UziTech:
@davisjam I submitted a PR to this PR that fixes all of the CM ATX header examples except one: https://github.com/davisjam/marked/pull/1

the example that still fails is because of whitespace so I'm not sure exactly how to fix that one yet:

```
CommonMark (ATX headings):
foo
    # bar

------

Expected:
<p>foo
# bar</p>

------

Marked:
<p>foo
    # bar</p>
```

davisjam:
I'll take a look soon.

davisjam:
@UziTech How much do you hate the `rtrim` I added in 0ceca90?

lib/marked.js
```diff
@@ -4,6 +4,38 @@
  * https://github.com/markedjs/marked
  */
 
+// Return str with all trailing {c | all but c} removed
+// allButC: Default false
+function rtrim(str, c, allButC) {
```
styfle: Should there be unit tests for this function?

UziTech: We would need to export the function. I think that would constitute [testing the implementation](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html).

davisjam: I agree with @UziTech.

lib/marked.js
```diff
@@ -552,6 +561,14 @@ var inline = {
   text: /^[\s\S]+?(?=[\\<!\[`*]|\b_| {2,}\n|$)/
 };
 
+if (NEW_TEXT) {
+  // TODO: If we replace ' {2,}\n' with '  \n' and address trailing whitespace,
+  //       we break the definition of GFM inline.breaks further down (affects the gfm_break test).
+  // Furthermore, we still have trouble with the email pattern substituted in: /|[...]+@/, which
+  // is vulnerable to REDOS just like /| {2,}\n/ was
+  inline.text = /[\s\S](?:[\\<!\[`*]|\b_|  \n|$)/;
```
davisjam: Looking for other thoughts here.

1. I think we can work with the definition of line breaks (` {2,}\n` vs. `  \n` with possibly-preceding spaces).
2. I'm not sure how to deal with the email problem. The full regex is this:

```[\s\S](?:[\\<!\[`*~]|https?:\/\/|ftp:\/\/|www\.|[a-zA-Z0-9.!#$%&'*+\/=?_`{\|}~-]+@|\b_|  \n|$)```

and the `/|[a-zA-Z0-9.!#$%&'*+\/=?_`{\|}~-]+@` is problematic without deliberately seeking out emails char-by-char.

We could do a linear pass for each of these options, but @UziTech likes regexes. This particular case seems pretty friendly to a linear pass IMO.

**Comment by UziTech** [comment]
> How much do you hate the `rtrim`

Not at all. I would move it to the bottom maybe just above `function marked(src, opt, callback) {`

> In your PR you said one case was still failing

I was talking about example 40 in the common mark spec:

```js
var shouldPassButFails = [40];
```

lib/marked.js
UziTech: Where is that email regex in `marked.js`? I don't see it

davisjam: [Here](https://github.com/markedjs/marked/blob/master/lib/marked.js#L577)

davisjam:
@UziTech Failing [40] is actually due to failing the CommonMark paragraph spec, it doesn't seem to have to do with headers per se.

```
  var shouldPassButFails = [185, 186];
```

For example:

```
3) CommonMark 0.28 Paragraphs should pass example 185
  Message:
    CommonMark (Paragraphs):
      aaa
     bbb
    
    ------
    
    Expected:
    <p>aaa
    bbb</p>
    
    ------
    
    Marked:
    <p>  aaa
     bbb</p>
```

davisjam:
bed4c13 makes heading case 40 pass along with a variety of others. CommonMark 318 fails now though.

 @UziTech @styfle Did I cheat somehow, or does applying the whitespace thing during rendering seem like a good idea?

joshbruce:
@davisjam: Take a look at #1229 marked settings at the top of `commonmark-spec`. I don't know if the one being used here has the same settings but think we've settled in and should look at our logical paths for different at some point. If we presume CommonMark is the no setting default, that means both `pedantic` and `gfm` should be false.

I don't know if this test bed had that yet. I noticed a lot of inconsistencies on what passes and what fails by changing those two setting.

davisjam:
@joshbruce I tried with those testOptions and there was no change in the success/failure of the tests affected by bed4c13. I guess orthogonal changes?

UziTech:
> CommonMark 318 fails now though

It looks like 318 should be failing without the changes. It is a false positive on html differ.

> does applying the whitespace thing during rendering seem like a good idea?

ya that works

davisjam:
> It is a false positive on html differ

Good eye. I'll add it to the "currently failing" list then.

davisjam:
This PR is making progress and has fixed a few REDOS vectors (rtrim) and spec errors (paragraph leading spaces). Hooray!

lib/marked.js
UziTech: I'm fine with doing more work in the lexer if it is going to speed things up.

Would the fix be to look for an `@` symbol before looking for those other chars, or checking for the rest of the regex if an `@` isn't found?

davisjam: I sketched an approach in 24d4a5e. I used my typical tactic of "mimic `RegExp.exec`" to reduce the visible impact. The email is handled [here](https://github.com/markedjs/marked/commit/24d4a5efb3653a676646750a67c1195959e27f93#diff-81ab0a5aa39b7a91951fc28d2d151496R587) amidst all the other offsetOfX routines. The rulesets are loaded with functions like [this](https://github.com/markedjs/marked/commit/24d4a5efb3653a676646750a67c1195959e27f93#diff-81ab0a5aa39b7a91951fc28d2d151496R656) and [this](https://github.com/markedjs/marked/commit/24d4a5efb3653a676646750a67c1195959e27f93#diff-81ab0a5aa39b7a91951fc28d2d151496R662).

Before I debug it all the way and deal with TODOs like [this](https://github.com/markedjs/marked/commit/24d4a5efb3653a676646750a67c1195959e27f93#diff-81ab0a5aa39b7a91951fc28d2d151496R733), I'd like to hear opinions about this approach strategy.

As a proof of concept, the PoC from the Sonatype folks is fine after 24d4a5e.

```
$ ./poc-sonatype.js  > /dev/null
String Length = 100005 | Execution time (seconds) 0.013
String Length = 1000005 | Execution time (seconds) 0.075
String Length = 10000005 | Execution time (seconds) 0.764
```

It shows linear growth as string becomes very large, to be expected with any O(n) algorithm. Without 24d4a53 it takes more than 10 seconds to evaluate just the first case.

davisjam:
Looking for feedback on the approach in 24d4a5e. It's a WIP (an off-by-one error somewhere) but I want to know whether you guys like the direction.

UziTech:
One of the benefits of having the regexes do most of the work is that it makes marked easy to [extend](https://github.com/markedjs/marked/blob/master/docs/USING_PRO.md).

If I want to use marked but change the way text is parsed I can with:

```js
marked.InlineLexer.rules.text = /some other regex/;
```

If we are going to start moving away from just regexes we need to figure out how to keep marked easily extendable.

UziTech:
On a security note. I also think we need to distinguish ReDoS attacks between "Can be triggered by a malicious actor" and "Can be triggered accidentally".

I don't think it is our job to prevent a malicious actor from causing marked from taking a long time to parse the input. We should educate our dependents on the safe way to deal with parsing user input. (i.e. web worker/vm.runInNewContext)

- If one of our dependents allows users to upload a 1 Terabyte markdown file there is nothing we can do to prevent marked from going slow.
- If a user must add 1000 spaces after character in order for marked to take 1 second to parse, is it really necessary to prevent that? (unless there might be some reason a normal person would add 1000 spaces after that character)

UziTech:
> (perhaps with an unsafe faster version)

See, I think marked **is** that unsafe faster version.

There are many other (safer) markdown parsing libraries. marked's "claim to fame" is it's speed and compliance with the original (pedantic) test suite

davisjam:
> marked's "claim to fame" is it's speed

Do we know how true this is anymore?

There are many markdown libraries, but *this* one has 3K dependents on npm. I feel reasonable security guarantees trump performance for any software, especially software with a large userbase.

joshbruce:
Leaning toward @davisjam on this one. Back to the priorities:

1. Security
2. Spec-compliance
3. Speed
4. Low-level
5. Lightweight

If I had to add extensibility it would 6.

> But personally I think saying "use module at your own risk, see threat model" is just another way of saying "don't use this module".

Agreed. 

(Maybe creating a threat model would be beneficial anyway. Give white hats a block of wood to widdle on - new challenges + new badges.) 

It seems there are two major topics of concern keeping us from merging this: speed and extensibility, neither of which supersedes the priority of security. As long as it's not a "serious" performance hit I say security wins and it seems like @davisjam is doing his best to strike a balance; especially if the performance hit doesn't show up until someone puts in 500+ blank spaces.

Marked isn't the fastest kid on the block anymore (if it ever was) from the stats we've seen: https://github.com/markedjs/marked/issues/963. The primary reason @UziTech and I showed up was security, seems like we might be wavering a bit. (Marked appears to be twice as slow as Showdown in #963.)

Maybe we should put in a more proper and consistent benchmarking methodology and add it to Travis. Then actually capture the benchmark somewhere (right now we have the 1s, which doesn't seem accurate given the size of the tests). Give us the ability to say "we won't go slower than" - unless X, Y, Z. With the JSON we can even put a target `ms` per test. 

Marked is going to change as we progress. I even mentioned this to @chjj at the beginning of the transition just before getting the org...it may not resemble the code he wrote...it's claim to fame when we showed up was that it was dead but no one had the heart to put it down humanely. Now it's maybe 60% spec compliant with CommonMark and GFM. Uses GFM by default (not sure how many people actually use Pedantic - with or without Marked - most of us didn't even know how to get our hands on the original Perl...think a couple of spike contributors didn't know about Gruber and Daring Fireball). It's trying to come back to life. 

What if Marked's new claim to fame is that it is the safest, most spec-compliant, *and* fastest. (I've definitely appreciated our security solutions since @davisjam showed up.)

joshbruce:
@davisjam: We are working on user analysis slowly but surely, see https://github.com/markedjs/marked/issues/1123

UziTech:
@davisjam `reasonable security` is what I am getting at. Trying to prevent **all** potential malicious DoS vectors in my mind is way past `reasonable security`. Dependents must understand that, regardless of using regexes, there will always be potential for malicious actors to make parsing slow if it is on the main thread, and if that is a security issue for them, the best course of action is to move the parsing off the main thread not slow down parsing for normal use cases.

@joshbruce I agree with that list of priorities and security should come first. I'm just saying there is such a thing as too much security and we need to draw that line somewhere. It would be really nice to get some sort of benchmarking so we can tell if these types of changes do in fact affect performance rather than just guessing.

P.S. I'm not against the changes in this PR or in 24d4a5efb3653a676646750a67c1195959e27f93 I just think we need to watch where that line is.

joshbruce:
@UziTech: I think it will be easier to draw the lines once we have a way to measure impacts on performance until then it's mainly conjecture. Therefore, I think drawing the line toward security is the better way to go. 

To @davisjam's point, I think we can also build the publicly safe API during 0.x then scale back a bit through options once we have a way to better measure performance. We can do that work in #963 for something after 0.4.

We're starting to rub up against the 2-week target for security things. What, if anything, can I do to help?

davisjam:
1. I agree that some kind of performance benchmark would go a long way to addressing these disputes.
2. @joshbruce We could pursue a general safe API as an alternative to all of the changes in this PR (except the ones that help with spec compliance). In particular 24d4a5e is somewhat character-changing, since it adds 100 LoC to a 1KLoC project for just one regex and there are surely others that would be similarly problematic. Though a safe API wouldn't resolve the issue unless we:
a. Told the reporters that we thought the exploit was out of scope, or
b. Replaced the existing API with the safe one and exposed the unsafe one as a separate API (thus protecting anyone who upgraded), or
c. Told everyone they had to refactor their marked code to use the safe API, or
d. ?

UziTech:
@davisjam is this PR still being worked on?

UziTech:
I'm going to close this PR. The changes will need to be moved to the `/src/` folder if you are still working on this.

PR #1227, opened 2018-04-17
move processing to output function

UziTech:
**Marked version:** 4711f6bb51c538b3693e35e8091cd55bdf23a934

## Description

- Moves processing in #1223 away from regex initialization

## Contributor

## Committer

- [x] Merge PR


lib/marked.js
```diff
@@ -762,8 +703,18 @@ InlineLexer.prototype.output = function(src) {
...
       this.inLink = true;
       href = cap[2];
-      href = href[0] === '<' ? href.substring(1, href.length - 1) : href;
-      title = cap[3] ? cap[3].substring(1, cap[3].length - 1) : cap[3];
+      if (this.options.pedantic) {
+        link = /^([^'"(]*[^\s])\s+(['"(].*['")])/.exec(href);
+
+        if (link) {
+          href = link[1];
+          title = link[2].trim().slice(1, -1);
```
davisjam: If you change the regex to `...\s+['"(](.*)['")]` then you wouldn't need this slice.

davisjam: I added `(` and `)` to the list of valid ways to denote a title based on the CommonMark [spec](https://spec.commonmark.org/0.28/#link-title). They were not in the original regex. Do we have a test for this?

lib/marked.js
```diff
@@ -762,8 +703,18 @@ InlineLexer.prototype.output = function(src) {
       src = src.substring(cap[0].length);
       this.inLink = true;
       href = cap[2];
-      href = href[0] === '<' ? href.substring(1, href.length - 1) : href;
-      title = cap[3] ? cap[3].substring(1, cap[3].length - 1) : cap[3];
+      if (this.options.pedantic) {
+        link = /^([^'"(]*[^\s])\s+(['"(].*['")])/.exec(href);
```
davisjam: I added `(` and `)` to the list of valid ways to denote a title based on the CommonMark [spec](https://spec.commonmark.org/0.28/#link-title). They were not in the original regex. Do we have a test for this?

lib/marked.js
```diff
@@ -762,8 +703,18 @@ InlineLexer.prototype.output = function(src) {
       src = src.substring(cap[0].length);
       this.inLink = true;
       href = cap[2];
-      href = href[0] === '<' ? href.substring(1, href.length - 1) : href;
-      title = cap[3] ? cap[3].substring(1, cap[3].length - 1) : cap[3];
+      if (this.options.pedantic) {
```
davisjam: We don't test for pedantic much, and I don't see it well documented anywhere. The only docs I could find are in the man page.

Can you explain when I should test for pedantic in the future?

davisjam: If I understand this code right, without `pedantic` we will set `href` but not `title`. If my understanding is correct, why is this behavior desirable?

UziTech: 1.  The regexes are grouped based on the options. The link regex that you fixed was in the pedantic group

    ```js
    ```

2.  The title is set in the else clause. The regex for a non-pedantic link is `/^!?\[(label)\]\(href(?:\s+(title))?\s*\)/` which is extremely complex when constructed and probably also error prone.

UziTech: here is the actual non-pedantic link regex:

```js
/^!?\[((?:\[[^\[\]]*\]|\\[\[\]]?|`[^`]*`|[^\[\]\\])*?)\]\(\s*(<(?:\\[<>]?|[^\s<>\\])*>|(?:\\[()]?|\([^\s\x00-\x1f()\\]*\)|[^\s\x00-\x1f()\\])*?)(?:\s+("(?:\\"?|[^"\\])*"|'(?:\\'?|[^'\\])*'|\((?:\\\)?|[^)\\])*\)))?\s*\)/
```

UziTech: I don't see why this shouldn't be part of the 0.4.0 release

UziTech:
I removed the `()` for title since the original markdown spec didn't allow them

> (['"])	# quote char = $5
> (.*?)		# title = $6
> \5		# matching quote

lib/marked.js
joshbruce: Sounds good. Just wasn't sure where the pedantic situation was.

davisjam: > The title is set in the else clause

Sorry I misspoke earlier.

On the path `!pedantic` it gets set.

I am thinking of the path:

`pedantic && !link`

and it seems like we then have `href` defined but not `title`?

UziTech: The regex that you used before when there was no title (`/^(<?[\s\S]*>?)/`) literally matches any string and since `href` is already set to the whole string there is no need to change it.

davisjam: Further down we then set the `title` field in the returned object based on the (undefined?) value of title. Do we want to set a default value for `title`, e.g. `''`?

joshbruce: My two cents would be an empty value of expected type. Avoid null and undefined checks and possibilities.

PR #1243, opened 2018-04-27
Base rewrite of rendering outputs to generate DOM elements instead of HTML strings

tbaeg:
Base rewrite of rendering outputs to generate DOM elements instead of HTML strings


tbaeg:
AH! Wrong branch. Sorry about that!

Issue #1208, opened 2018-04-09
the option : headerIds is not valid.


li-xiaojun:
**markd: 0.3.19:**
when i set headerIds: false，but it still generated the ids for per element.

UziTech:
I don't believe the headerIds option was added to a release yet

/cc @joshbruce

joshbruce:
Believe @li-xiaojun is referring to the feature that is already in place. The only change we made was to make it so you could disable it (not part of a release yet). The default behavior is to add header ids. We've always had issues with UTF-8 characters; see #981.

styfle:
Hi @li-xiaojun sorry for the confusion 😄 

Our docs are based off of `master` which includes this new `headerIds` option however it is not released yet. Our next release, 0.4.0 will include it (as the other team members mentioned above).

If you're feeling adventurous, you could try out the code in the master branch before the release.

We'll close this issue when 0.4.0 gets released. Thanks! 👍

PR #1244, opened 2018-04-30
Update advanced usage docs

styfle:
## Description

This adds a couple missing option types to the docs and also shows the version introduced for `headerIds`.

- Related #1208

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/docs/USING_ADVANCED.md
+++ b/docs/USING_ADVANCED.md
@@ -47,7 +47,7 @@ console.log(myMarked('I am using __markdown__.'));
 |headerIds   |`boolean` |Whether to add an `id` attribute to headers. Default: `true`. Added in 0.4.0           |
 |headerPrefix|`string`  |A short string to add as a prefix to the `id` attributes added to headers by default. Default: `empty string`                |
 |highlight   |`function`|A function to highlight code blocks. See also: <a href="#highlight">Asynchronous highlighting</a>.                           |
-|langPrefix  |`??`      |Default is `lang-`
+|langPrefix  |`string`  |A string to prefix the className in a `<code>` block. Useful for syntax highlighting. Default is `lang-`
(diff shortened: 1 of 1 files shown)


styfle:
@joshbruce @UziTech @davisjam I'm thinking of adding 2 additional columns to the options table:

- default value
- since version

Do you think that's a good idea?

joshbruce:
tl;dr: Not a fall of the sword for me. Just some things to consider.

I definitely think having `default` following `type` would be good. My only reservation on `since` would be in keeping up if they're out of sync. Though it does open the door for a `deprecated` column as well when we get there.

docs/USING_ADVANCED.md
```diff
@@ -41,18 +41,18 @@ console.log(myMarked('I am using __markdown__.'));
 
 |Member      |Type      |Notes                                                                                                                        |
 |:-----------|:---------|:----------------------------------------------------------------------------------------------------------------------------|
-|baseUrl     |`??`      |Default is `null`                                                                                                            |  
+|baseUrl     |`string`  |A prefix url for any relative link.  Default is `null`               |
```
joshbruce: If we add the columns, think the order would be:

|Member |Type |Default |Added in |Notes|

styfle: Thanks! I went with `Since` and put `v0.4.0` so its clear that its clear that this is a version number. I couldn't easily find when these other features were introduced so I used `???`

styfle:
> My only reservation on since would be in keeping up if they're out of sync

I don't understand this reservation. What would get out of sync?

> Though it does open the door for a deprecated column as well when we get there.

I like the idea of deprecating too if we ever reach that point.

styfle:
@joshbruce Any opposition here?

PR #1245, opened 2018-05-01
Gfm tables

tomtheisen:
**Marked version:** 0.3.19

**Markdown flavor:**  GitHub Flavored Markdown

## Description

This PR introduces full compliance with GFM tables. (I think)  I'm new here, and this is a somewhat bigger code change.  Despite my best efforts, I may have messed something up, so feedback is welcome.

## Contributor

The gfm_tables tests cover this functionality.

## Committer

- [x] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -266,13 +266,13 @@ Lexer.prototype.token = function(src, top) {
             item.align[i] = null;
           }
         }
-        
+
         for (i = 0; i < item.cells.length; i++) {
           item.cells[i] = splitCells(item.cells[i], item.header.length);
         }
-        
+
         this.tokens.push(item);
-        
+
         continue;
       }
     }
@@ -932,7 +932,7 @@ Renderer.prototype.paragraph = function(text) {
 };
 
 Renderer.prototype.table = function(header, body) {
-  if (body) body = '<tbody>' + body + '</tbody>'; 
+  if (body) body = '<tbody>' + body + '</tbody>';
 
   return '<table>\n'
     + '<thead>\n'
@@ -1322,7 +1322,7 @@ function splitCells(tableRow, count) {
   if (cells.length > count) {
     cells.splice(count);
   } else {
-    while (cells.length < count) cells.push("");
+    while (cells.length < count) cells.push('');
   }
 
   for (; i < cells.length; i++) {
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -107,8 +107,8 @@ block.gfm.paragraph = edit(block.paragraph)
  */
 
 block.tables = merge({}, block.gfm, {
-  nptable: /^ *(\S.*\|.*)\n *([-:]+ *\|[-| :]*)\n((?:.*\|.*(?:\n|$))*)\n*/,
-  table: /^ *\|(.+)\n *\|( *[-:]+[-| :]*)\n((?: *\|.*(?:\n|$))*)\n*/
+  nptable: /^ *([^|\n ].*\|.*)\n *([-:]+ *\|[-| :]*)(?:\n((?:.*[^>\n ].*(?:\n|$))*)\n*|$)/,
+  table: /^ *\|(.+)\n *\|?( *[-:]+[-| :]*)(?:\n((?: *[^>\n ].*(?:\n|$))*)\n*|$)/
 });
 
 /**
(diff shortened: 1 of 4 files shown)


tomtheisen:
It seems there's some catastrophic backtracking in some of the regexes.  For some reason, my local tests did not find these, but I can see a list in the travis-ci link.

UziTech:
That is strange. No errors show up on my computer either when I run `npm run test:redos` on this PR.

/cc @davisjam 

It looks like the failing tests aren't ones you changed in this PR anyway.

lib/marked.js
```diff
@@ -944,7 +949,7 @@ Renderer.prototype.tablerow = function(content) {
 Renderer.prototype.tablecell = function(content, flags) {
   var type = flags.header ? 'th' : 'td';
   var tag = flags.align
-    ? '<' + type + ' style="text-align:' + flags.align + '">'
+    ? '<' + type + ' align="' + flags.align + '">'
```
Martii: Re: 
* https://www.w3schools.com/tags/att_table_align.asp ... Not HTML5 compliant according to them. Also in a few other places in this PR. Is that compliance being discarded?

Martii:
Be interesting to compare this PR on Android as my GFM tables are way messed up there with Android 8 and latest Chrome too. Desktops have been acceptable for years but portables really look weird. [Example rendered page](https://openuserjs.org/scripts/Marti/oujs_-_Meta_View#-compatibility-matrix) where a GFM table is made and generated with this issue.

lib/marked.js
tomtheisen: This is GFM compliant, not HTML  compliant.  Compliance with GFM requires that we break HTML compliance.  

https://github.github.com/gfm/#example-192

davisjam:
Here are the explanations.

1. As @styfle noted in the associated [issue](https://github.com/markedjs/marked/issues/1201), the REDOS detector has a lag. On new regexes it returns UNKNOWN (translated to SAFE by the eslint plugin) and analyzes them in the background; a subsequent query may return VULNERABLE. This is probably what affected @tomtheisen.
2. I recently updated the REDOS detector logic to identify new classes of vulnerable regexes. See [this commit](https://github.com/davisjam/vuln-regex-detector/pull/49). For example, `/\n+$/` would not have been flagged before but was vulnerable. That particular regex is fixed in the [rtrim commit](https://github.com/markedjs/marked/pull/1226/commits/0cfe39ec2db1e4362f5d004d4e955648d95b501b) of #1226, though some of the others may still be relevant.
3. The eslint plugin used in `npm run test:redos` relies on the [vuln-regex-detector](https://www.npmjs.com/package/vuln-regex-detector) module, which maintains a local [cache](https://github.com/davisjam/vuln-regex-detector/pull/28). I have not yet published a version that [expires](https://github.com/davisjam/vuln-regex-detector/pull/46) older entries in the event of updates to the REDOS detector (like the one described in the previous bullet). This may be what's affecting @UziTech's runs. The REDOS server previously concluded that these regexes were SAFE and that result would have been cached on developer machines.

Travis is run repeatedly so it overcomes limitation 1 eventually. Travis does not have a cache so it does not have a problem with the combination of limitations 2 and 3.

If @tomtheisen did not introduce these regexes, the PR is fine from a security standpoint, although a separate issue should be opened to track them. This is relevant to the discussion about a safe API in #1226.

styfle:
I still stand by my solution for a [versioned API](https://github.com/markedjs/marked/issues/1201#issuecomment-378992300) to avoid this problem in the future. Then we could disable the caching.

styfle:
@davisjam In the meantime, can we merge this PR?

davisjam:
If @tomtheisen did not introduce these regexes, the PR is fine from a security standpoint. I didn't check.

UziTech:
@styfle yes, the travis log lists the line numbers of the vulnerable regexes and none of them were touched by this PR

image

@davisjam Just curious, how is the regex `/ *\| */` vulnerable to catastrophic backtracking?

davisjam:
@UziTech `/ *\| */` lacks a leading anchor, and a mismatch will be triggered on whitespace followed by no pipe. Each mismatch costs `O(n)`. Meanwhile the V8 regex engine has an outer `for all starting offsets` loop, thus `O(n) * O(n) = O(n^2)` complexity.

```
(11:27:14) jamie@woody ~ $ cat /tmp/dos.js
/ *\| */.exec(' '.repeat(100000))
(11:27:16) jamie@woody ~ $ time node /tmp/dos.js 

real	0m5.742s
user	0m5.734s
sys	0m0.009s
```

UziTech:
I see, so any regex that doesn't have a character limit or lead anchor will be subject to catastrophic backtracking?

lib/marked.js
Martii: @tomtheisen 
> This is GFM compliant, not HTML compliant. Compliance with GFM requires that we break HTML compliance.

That's why it's in the form of a question. Good to know of a major breaking change as @styfle put it in the issue. If we continue to use this package we'll need to undo this for our sanitizer *(which we do for HTML5 anyhow)* as we are always striving for HTML5 compliance. Anyone using this change that has a HTML5 compliant badge will need to do this as well since it's non-standard or they'll lose their logo/rating.

Since I've been on GH just about as long as they've been around I can make an *educated guess* that they did this spec for older browser compatibility and also their sanitizer. So when the HTML4.x specifications drop out *(presumably around 2020ish when everything hits the fan)* the backward compatibility reintroduced here will not render tables correctly and it may need to be reversed. Time will tell.

Issue #1108, opened 2018-03-02
Devel: Require updates to 'News' as part of PRs


davisjam:
**Proposal type:** project operations

## What pain point are you perceiving?

Hard to write changelogs for new releases.

## What solution are you suggesting?

Require PRs to update a "News" file with a changelog-friendly description of the change.
This is what cpython does.
Then generating a changelog is a copy/paste job or could be automated.

davisjam:
@UziTech

UziTech:
The only problem I see with making every PR add their own change to a file is whenever a PR is merged then all other unmerged PRs have a merge conflict.

I don't think the average person/dependent cares what changed and the people who do can easily [compare the commits](https://github.com/markedjs/marked/compare/0.3.15...v0.3.17) between releases

joshbruce:
To be fair, I'm more interested in the PRs that have gone into `master` between releases. I might be able to do the language thing pretty quickly as long as I can see the PRs easily somewhere; so, now I need the education on how to get wherever you got that page from.

I'm not sure I agree that they don't care. We've had a couple of folks talk about semver not letting them know and, from a marketing perspective, it's more about bringing on new users.

styfle:
Changes since our last release:
https://github.com/markedjs/marked/compare/v0.3.17...master

joshbruce:
@styfle: Thanks. I think my ignorance is greater than that. lol 😂 How do I actually get to the point of seeing that page in GitHub? UI sequence?

UziTech:
![image](https://user-images.githubusercontent.com/97994/36917936-d49b4b56-1e1d-11e8-85c6-f99c0baace21.png)

davisjam:
> whenever a PR is merged then all other unmerged PRs have a merge conflict.

Sorry, put them in *distinct* news files. e.g. `2018-03-02-davisjam-PRX-ChangeLog.md`.

UziTech:
https://github.com/motemen/git-pr-release

styfle:
I don't really see the value in this since we are already maintaining the [release notes](https://github.com/markedjs/marked/releases) which can be fetched via GitHub API and any git user can [generate a changelog](https://github.com/markedjs/marked/issues/1108#issuecomment-370025973) from the tags.

Should we close this ticket?

joshbruce:
[...]

PR #1251, opened 2018-05-03
Update issue templates

styfle:
This adds templates so users are prompted when they create a new issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
+ - Device: [e.g. iPhone6]
+ - OS: [e.g. iOS8.1]
+ - Browser [e.g. stock browser, safari]
+ - Version [e.g. 22]
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -26,19 +26,5 @@ Steps to reproduce the behavior:
 **Expected behavior**
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
-
-**Desktop (please complete the following information):**
- - OS: [e.g. iOS]
- - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
-
-**Smartphone (please complete the following information):**
- - Device: [e.g. iPhone6]
- - OS: [e.g. iOS8.1]
- - Browser [e.g. stock browser, safari]
- - Version [e.g. 22]
-
 **Additional context**
 Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -9,10 +9,19 @@ A clear and concise description of what the bug is.
 
 **To Reproduce**
 Steps to reproduce the behavior:
-1. Go to '...'
-2. Click on '....'
-3. Scroll down to '....'
-4. See error
+
+<!-- If possible, use the Marked Demo permalink and compare to the CommonMark Dingus permalink to demonstrate the bug -->
+<!--
+1. [Marked Demo](https://marked.js.org/demo/)
+2. [CommonMark Demo](https://spec.commonmark.org/dingus/)
+-->
+
+<!-- If you need a specific version and options to reproduce the bug, use the following template -->
+<!--
+1. Install marked `npm install --save marked@0.3.19` with the version you are using
+2. Run marked with input string and options such as `marked('hello *world*', {gfm: true})`
+3. Actual output (or error) is...
+-->
 
 **Expected behavior**
 A clear and concise description of what you expected to happen.
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Marked says it does this thing but does not
 
 ---
 
@@ -25,6 +25,3 @@ Steps to reproduce the behavior:
 
 **Expected behavior**
 A clear and concise description of what you expected to happen.
-
-**Additional context**
-Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/Feature_request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_request.md
@@ -1,0 +1,17 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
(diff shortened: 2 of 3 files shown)


joshbruce:
I like the idea but not sure about these as options. Wondering if we could reformat the issue template we have into two or three types from GitHub.

The quality of issues in my opinion since creating the template we're using now.

Bug report, feature request, and proposal??

UziTech:
could we also add the demo links to a bug report.

styfle:
@UziTech I added demo links

@joshbruce What is the difference between Feature Request and Proposal? And can we add that in a later PR?

joshbruce:
@styfle: I think

- Issue: Marked says it does this thing but doesn't.
- Feature Request: Marked doesn't do this thing and I think it should
- Proposals: Kind of an "other" bucket with direction could be and might have nothing to do with the code, "I think badge X is awesome, you should add badge X to the README" or "I think we should operate things a bit differently"..."have you seen this tool"...that sorta thing. We as the core have other channels, this is the public channel.

The most vibrant communities I've seen have a way to hold public discourse on all sort of topics. I've seen giving folks a way to report a defect, requesting a feature, and "other" (propose something) as beneficial...some of the tickets I closed fell into this category and I personally tend to use that channel myself.

joshbruce:
ps. I definitely think our current issue template has proven itself worthy. These don't seem like us and maybe overly generic.

styfle:
@joshbruce I updated the descriptions like you suggested and made the template shorter.

Do you have an idea for the contents of the Proposal template?
Can the Proposal template be added in a later PR so we can get an idea of how this works?

joshbruce:
@styfle:

> Do you have an idea for the contents of the Proposal template?

Sure. See current issue template. ;)

```
**Proposal type:** project operations | other

```

> Can the Proposal template be added in a later PR so we can get an idea of how this works?

I suppose - it's already there though and has been used by people other than me: https://github.com/markedjs/marked/issues/1108. 

Again, my concern is that we seem to be heading into new feature world and thinking we have to replace the content we already have to get there. Again, our current template seems to work for our users and we can modify the content and put in the front matter which seems to be how GitHub knows what's going - no need to reinvent the content. Adding the feature request, which is currently part of the proposal type options in our current template.

joshbruce:
Are we concerned that we don’t fully know how it works? We can always update later...just curious about any urgency for testing purposes. Sorry, again, just got back up with my laptop; so, I’m behind on Marked conversations.

styfle:
@joshbruce Ok I added the proposal template.

The first two templates "Bug" and "Feature" were generated by GitHub and I like the simplicity of them. Our current template has too much going on in my opinion...43 lines is kind of off-putting if you're a first time contributor.

Here's what it was before:

```
**Marked version:**

<!-- The NPM version or commit hash having the issue --> 

<!-- 

	If submitting something other than a defect with Marked itself, please use the following:

**Proposal type:** new feature | project operations | other

-->

## Expectation

**CommonMark Demo:** [demo](https://spec.commonmark.org/dingus/)
<!-- You can replace the link above with a permalink from: https://spec.commonmark.org/dingus/ -->

<!-- Describe the output you are expecting from marked -->

## Result

**Marked Demo:** [demo](https://marked.js.org/demo/)
<!-- You can replace the link above with a permalink from: https://marked.js.org/demo/ -->

<!-- Describe the output you received from marked -->

## What was attempted

<!-- Describe what code combination got you there -->

<!-- 
	If error is thrown add the following:

## Call stack & console log

-->

```

styfle:
> Are we concerned that we don’t fully know how it works?

Yes, I'm not sure how it works and it would be good to see it in action. I don't know of a way to try it out this without merging to master.

.github/ISSUE_TEMPLATE/Feature_request.md
```diff
@@ -0,0 +1,14 @@
+---
+name: Feature request
+about: Marked doesn't do this thing and I think it should
+
+---
+
+**Is your feature request related to a problem? Please describe.**
```
joshbruce: Not feeling this set...it has some potential overlap with a bug report. If it's cool with you, once we merge and verify I might submit another PR as follow up. (To the line count comment combined it's the same count. :) )

Don't think the marked version makes sense here.

.github/ISSUE_TEMPLATE/Proposal.md
```diff
@@ -0,0 +1,11 @@
+---
+name: Proposal
+about: Marked doesn't do this thing and I think it should
+ 
+---
+ 
+**What pain point are you perceiving?.**
```
joshbruce: This one is where having the marked version made the least sense - but that was the limitation of having to put all three in one.

joshbruce:
To the reference of the current one, thinking we could divide it into three without having to change verbiage or formatting (because it does serve three disparate purposes because it used to be the only way.

## Bug

```
**Marked version:**

<!-- The NPM version or commit hash having the issue --> 

## Expectation

**CommonMark Demo:** [demo](https://spec.commonmark.org/dingus/)
<!-- You can replace the link above with a permalink from: https://spec.commonmark.org/dingus/ -->

<!-- Describe the output you are expecting from marked -->

## Result

**Marked Demo:** [demo](https://marked.js.org/demo/)
<!-- You can replace the link above with a permalink from: https://marked.js.org/demo/ -->

<!-- Describe the output you received from marked -->

## What was attempted

<!-- Describe what code combination got you there -->

<!-- 
	If error is thrown add the following:

## Call stack & console log

-->
```

## Proposal

```
**Proposal type:** project operations | other

```

## Feature request

```
<!-- describe the desired functionality and any alternatives you have attempted -->
```

Dividing them this way gives us clarity into what might be the problem child of complexity - the defect one...but people seem to be doing okay even with that complexity.

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Bug report
+about: Marked says it does this thing but does not
+
+---
+
```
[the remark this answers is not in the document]
styfle: The marked version is in step 1 below but I can make it a separate question so it's more clear.

.github/ISSUE_TEMPLATE/Proposal.md
styfle: I'm confused. There is no version requested for the Proposal template.

.github/ISSUE_TEMPLATE/Bug_report.md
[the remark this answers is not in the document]
styfle: Now I remember why, I decided to tell them to use the demo to see if they can reproduce, that way we can easily see the bug with one click. I then suggest more specific steps to repro if the bug doesn't show up in the demo (that might mean they're not using the latest version).

I'm open to suggestions with how to reword this one.

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Bug report
+about: Marked says it does this thing but does not
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
```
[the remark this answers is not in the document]
styfle: Yeah you're right, it was making the user delete a bunch of text before. Now they'll get to choose with buttons

.github/ISSUE_TEMPLATE/Feature_request.md
styfle: Yeah there's room for improvement. But after reading through it a couple times, I think it makes sense. A feature request should be rooted in a problem that needs to be solved.

Maybe we can change this question to say something like "Why is this feature valuable?"

.github/ISSUE_TEMPLATE/Bug_report.md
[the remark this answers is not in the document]
joshbruce: Funny enough I got thinking we could have two bug reports - one using the demo and one without...still debating that one. If this PR is mainly for testing the concept of buttons and not focused on the templates then we can probably do the merge and I might come in behind with a PR using UX-based considerations.

One of the things we've been doing with the previous template was to make the desired template easy to access and let people opt in to alternative templates and complexity the less users have to do themselves the better is the general rule. Then, from journalism, using the inverted pyramid to present more general information to more detailed: What version? What options? What did you do? What did you expect? What did you get? (Believe these are out of order in our current template actually.) Then fewer words.

.github/ISSUE_TEMPLATE/Feature_request.md
styfle: Yes, but in your example about iTunes offering a way to purchase rented movies, you do have an implicit problem that has not been stated. Maybe the problem is that you can't watch rented movies more than once, so buying it solves that problem.

I'll change this to "why is this feature valuable" so we can at least get a Why from the person requesting it.

.github/ISSUE_TEMPLATE/Feature_request.md
```diff
@@ -0,0 +1,14 @@
+---
+name: Feature request
+about: Marked doesn't do this thing and I think it should
+
+---
+
```
joshbruce: Appreciating the updates to the feature request. Gonna be catching up on some things elsewhere but want to drop by as I'm going through emails.

styfle:
@joshbruce Any other changes requested?

PR #1258, opened 2018-05-09
GFM strikethrough compatibility

tomtheisen:
**Marked version:** 0.3.19

## Description

This pull request brings GFM strikethrough compatibility.

## Contributor

## Committer

- [ ] Merge PR


tomtheisen:
I don't *think* this is catastrophically backtracking regex.

lib/marked.js
```diff
@@ -588,7 +588,7 @@ inline.gfm = merge({}, inline.normal, {
     .replace('email', inline._email)
     .getRegex(),
   _backpedal: /(?:[^?!.,:;*_~()&]+|\([^)]*\)|&(?![a-zA-Z0-9]+;$)|[?!.,:;*_~)]+(?!$))+/,
-  del: /^~~(?=\S)([\s\S]*?\S)~~/,
+  del: /^~+(?=\S)([\s\S]*?\S)~+/,
```
davisjam: 1. Could this be simplified (no look-ahead) to `/^~+([^\s~][\s\S]*?\S)~+/`?
2. Why not `~~ leading/trailing spaces ~~`? Seems like the [spec](https://github.github.com/gfm/#strikethrough-extension-) permits this.

UziTech: 2. The spec might allow it but github doesn't

without space:
~asdf~

with space:
~ asdf ~

davisjam: Interesting. Would this work? `/^~+([^~]+)~+/`.

The spec says "No nesting" which I think means that the first `~` will terminate the strike-through.

UziTech: I think we should act like GitHub when the spec is ambiguous

davisjam: To answer my own question, `/^~+([^\s~][\s\S]*?\S)~+/` won't work -- it requires a minimum of two characters between the `~`s.

davisjam: I'm a little concerned that this regex (old and new versions) will match `~~~`. How about 

```
/^~+(?=[^\s~])([\s\S]*?[^\s~])~+/
```

which will ensure that the "text" neither begins nor ends with whitespace or a `~`.

tomtheisen: Is there any particular reason to avoid the non-greedy modifier?

davisjam: I'm nervous about the structure `/~+...[\s\S]*?...~+/`. I do not believe that *as written* the regex is vulnerable, but a slight modification would make it so. If the middle group excludes a `~` character then the risk is removed.

styfle: @davisjam If this is not vulnerable as written then I think we should merge it because it brings us into compliance. Adding @joshbruce to take a look.

styfle:
Thanks Tom! 🎉

PR #1266, opened 2018-05-11
Remove extra inner linebreak from code blocks

remyrylan:
## Description

- Fixes
Currently, `marked` adds a line break before inserting `</code></pre>`, some highlighting and parsing tools I've ran into actually interpret this line break (since it's wrapped in a `<pre>`) as an extra line break that should be displayed in the output.

Essentially the following HTML from marked:
```html
<pre><code class="language-js">const foo = 'bar'
</code></pre>
```

Will produce rendered output with some highlighters as having an extra trailing inner line break:
```js
const foo = 'bar'


```

## Expectation
(no trailing inner line break)
```html
<pre><code class="language-js">const foo = 'bar'</code></pre>
```
Then highlighters will render correctly as:
```js
const foo = 'bar'
```

## Contributor

## Committer

- [ ] Merge PR


remyrylan:
Forced a new commit that removed `marked.min.js` from my changes.

styfle:
@hail2u Which example number in the spec are you referencing?

hail2u:
Example 109 is most clear. This 1 line fenced code block has line break. Also [commonmark.js dingus](https://spec.commonmark.org/dingus/?text=%60%60%60js%0Aconst%20foo%20%3D%20'bar'%0A%60%60%60%0A) can reproduce this behavior.

styfle:
This is a bit strange since the spec doesn't seem to call it out directly. Only the example html output (and consequently the dingus) show this "feature".

[Example 99](https://spec.commonmark.org/0.28/#example-99) seems to avoid adding a newline.

[Example 109](https://spec.commonmark.org/0.28/#example-109) seems to add a newline.

Our tests show both 99 and 109 as passing but I would expect one to pass and one to fail.

@UziTech Is this due to the way `html-differ` works?

hail2u:
To be honest, there is no big problem here. Some syntax highlighter shows `↲` character just before line break like GUI text editors. With an HTML rendered by marked v0.5.0, they failed to show last `↲` because there is no last line break. But this is a little problem.

I think the problem mentioned in this issue is also a little problem. The original markdown, CommonMark, GFM, and marked prior to v0.4.0 leave the last line break as it is. It would be a reasonable and a safest behavior, isn’t it? Markdown has been used for a long time, there might be a tool that depends on this behavior.

---

I find a clear rule about this line break from the CommonMark spec - [4.4 Indented code blocks](https://spec.commonmark.org/0.28/#indented-code-blocks):

> ... The contents of the code block are the literal contents of the lines, including trailing line endings, ...

Feder1co5oave:
The passing tests are because of [this bug](https://github.com/bem/html-differ/issues/150) in html-differ.
[My fork](https://github.com/Feder1co5oave/html-differ) should fix this. Try it out and see what goes wrong.

Issue #302, opened 2013-12-04
Removing existing options in favor of the renderer


chjj:
We have the opportunity to remove:
- highlight
- langPrefix
- headerPrefix
- heading IDs

In favor of the new renderer.

The downside to this (at least for 3 of them): lack of backward compatibility, and less convenience for users used to marked. While I did merge the renderer, I'm **tentatively** reimplementing a few of these options. I think the removal of these features should be discussed in another issue because it is a big decision.

cc @lepture

birendra-ideas2it:
how to give blockquote regular expression like slack?

PR #1275, opened 2018-05-24
Support math latex.

ahuigo:
**Marked version:**
Release 0.4.0

## Contributor
I don't know how to test this project exactly. Any advice?

The demo for this feature:
> https://github.com/ahuigo/marked/blob/master/docs/demo/latex.html

## Committer

- [ ] Merge PR


ahuigo:
@UziTech  
It's hard to add a wrapper for marked. I have to add a math lexer to marked.( If user want to parse math, he should pass option `latexParser`).  

1. I wish marked itself would support math lexer, because many people need math latex.
2. If I create a  separate npm, it was hard to pull code back to upstream , as I have changed marked's lexer directly.

UziTech:
We are trying to [slim marked down](https://github.com/markedjs/marked/issues/302). By the 1.0 release we would like to deprecate most of the options.

We are trying to focus on making marked spec compliant with CommonMark and GFM. For everything else we would like to create separate modules that extend marked.

I can leave this PR open so we can revisit it in the future and possibly create a module ourselves but it will be a while before that happens.

BenjaminHoegh:
Is there a API for marked or is it code it from the ground if you wanna make extensions to it?

BenjaminHoegh:
@UziTech Posible without node.js? Some web hotels don't support nodejs and would like to keep it working on so many server types as possible

UziTech:
@BenjaminHoegh Is what possible without node? marked is written in js so it needs a js engine in order to run at all.

If you use marked on the client side it will create a global variable `marked` so you just replace `var myMarked = require('marked');` with `var myMarked = marked;`

BenjaminHoegh:
@UziTech Is there documentation about how to add new rules to inline and block

UziTech:
I don't believe there is a way to add **new** rules easily yet. All of our documentation is as [marked.js.org ](https://marked.js.org)

yhcharles:
@ahuigo 
It seems the blocklatex is rendering in display mode.

ahuigo:
@yhcharles What you need is `\displaystyle` , for example`$\displaystyle \sum_i^j$`.  You can try it  here: https://ahuigo.github.io/a/md.html

BrianHung:
@yhcharles 

This is a bit late, but another solution would be to change the source code beginning at line 935 from
```javascript
      var html = marked.defaults.latexRender(text);
      if (block) {
        out = '<div style="text-align:center">' + html + '</div>';
      }else{
        out = html;
      }
```

to
```javascript
      if (block) {
        out = marked.defaults.latexRender(text, {displayMode: true});
      }else{
        out = marked.defaults.latexRender(text);
      }
```

This assumes you're using KaTeX to render LaTeX; documentation [here](https://katex.org/docs/options.html). I'm sure that MathJax has the same setting.

Note I haven't tested this implementation yet.

Issue #1215, opened 2018-04-12
Dependents active??


joshbruce:
**Marked version:** 0.3.19

**Markdown flavor:** n/a

**Proposal type:** other

## What pain point are you perceiving?

3,000+ dependents...really??

## What solution are you suggesting?

An automated way to iterate over the list of dependents. See what version they are using and when they last published to NPM. Hypothesis: Many of those packages are not active, reference Marked as prior art, or have a small audience base. 

Tagging @UziTech, @styfle, and @Martii

See also #1123

styfle:
You can [page through all 3000 dependendents](https://www.npmjs.com/browse/depended/marked).

But it doesn't look like there is any particular order.

@joshbruce What would we consider inactive? No downloads in the last 30 days?

joshbruce:
@UziTech: ::sad trombone:: 

```
error getting dependents for [marked] { Error: getaddrinfo ENOTFOUND registry.npmjs.org registry.npmjs.org:443

error getting dependents for [marked] Error: Invalid JSON (Unexpected "h" at position 1 in state STOP)
```

@styfle: I don't know what constitutes active, to be honest, and it's a fair question. Think I'm mainly looking for ways to even start gathering the data easily on the dependents. In some cases, paging through, the publish date would be a good first start (some were published in the last week). npm/npm is easy because they specify a version number and "give credit" when they update; so, we get notified that they're still active and actively using (aware of) Marked.

UziTech:
@joshbruce the errors are from sending too many requests to the npm rest api. There has to be two requests per dependent. I am working on a way to fix that ... stay tuned

UziTech:
Here is a distribution of the version numbers:

```
[
  [
    594,
    "^0.3.6"
  ],
  [
    450,
    "^0.3.5"
  ],
  [
    218,
    "^0.3.2"
  ],
  [
    182,
    "^0.3.3"
  ],
  [
    103,
    "~0.3.2"
  ],
  [
    77,
    "*"
  ],
  [
    76,
    "0.3.6"
  ],
  [
    59,
    "0.3.5"
  ],
  [
    54,
    "^0.3.9"
  ],
  [
    51,
    "^0.3.12"
  ],
  [
    49,
    "~0.2.9"
  ],
  [
    40,
    "0.3.2"
  ],
  [
    37,
    "~0.3.1"
  ],
  [
    35,
    "^0.3.7"
  ],
  [
    35,
    "0.2.x"
  ],
  [
    34,
[...]

PR #1299, opened 2018-06-28
fix issues link references and prototypes

Trott:
**Marked version:** 0.4.0

**Markdown flavor:** Markdown.pl

## Description

Link with names that clashed with properties inherited from the
Object prototype (such as "constructor") were not expanding. This fixes
this issue.

Before this change, markdown of this form...:

    Link: [constructor][].

    [constructor]: https://example.org/

...resulted in HTML output of this form:

    <p>Link: [constructor][].</p>

With this change, it now renders as expected:

    <p>Link: <a href="https://example.org/">constructor</a>.</p>

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/test/specs/marked/marked.json
+++ b/test/specs/marked/marked.json
@@ -52,5 +52,11 @@
     "markdown": "|1|2|\n|-|-|\n| |2|",
     "html": "<table><thead><tr><th>1</th><th>2</th></tr></thead><tbody><tr><td></td><td>2</td></tr></tbody></table>",
     "example": 9
+  },
+  {
+    "section": "Links",
+    "markdown": "Link: [constructor][].\n\n[constructor]: https://example.org/",
+    "html": "<p>Link: <a href=\"https://example.org/\">constructor</a>.</p>",
+    "example": 10
   }
 ]
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -135,7 +135,7 @@ block.pedantic = merge({}, block.normal, {
 
 function Lexer(options) {
   this.tokens = [];
-  this.tokens.links = {};
+  this.tokens.links = Object.create(null);
   this.options = options || marked.defaults;
   this.rules = block.normal;
 
--- a/test/specs/marked/marked-spec.js
+++ b/test/specs/marked/marked-spec.js
@@ -47,6 +47,20 @@ describe('Marked Code spans', function() {
   });
 });
 
+describe('Marked Links', function() {
+  var section = 'Links';
+
+  var shouldPassButFails = [];
+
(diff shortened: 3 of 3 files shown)


UziTech:
@joshbruce @styfle 
this may need to be polyfilled if we intend on supporting ie8

[Object.create Browser Compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#Browser_compatibility)

test/specs/marked/marked.json
```diff
@@ -5,6 +5,12 @@
     "html": "<p><code>someone@example.com</code></p>",
     "example": 1
   },
+  {
+    "section": "Links",
+    "markdown": "Link: [constructor][].\n\n[constructor]: https://example.org/",
+    "html": "<p>Link: <a href=\"https://example.org/\">constructor</a>.</p>",
+    "example": 10
+  },
```
UziTech: Can you keep the example numbers in order?

Trott: Sure, done. (I thought they were alphabetically ordered by section.)

Trott:
> this may need to be polyfilled if we intend on supporting ie8

The polyfill at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create does not support `null` as an argument. I suspect other polyfills may have the same shortcoming.

Would this be tolerable?

```
// IE8 does not have Object.create()
this.tokens.links = Object.create ? Object.create(null) || {};
```

The tiny edge-case bug this fixes will still exist in IE8, but it also means not having to ship a polyfill. An issue can be opened for adding the polyfill and it can happen later (or the issue can be closed whenever IE8 support is dropped).

Trott:
> I'm ok with the check for `Object.create`, I'm also ok with not supporting IE8

I'll assume "not support IE8" is the most likely path until someone else weighs in to the contrary. :-D

joshbruce:
@styfle, @UziTech, et al: Given the data we have from #1215 and the fact that IE11 isn't supported by Microsoft anymore I'm pretty okay with not supporting IE8 as well. 

Having said that, I do wonder what criteria we should set for supporting IE in general (more specifically JS engines)...feel like we almost had this conversation not too long ago.

styfle:
@joshbruce There is some discussion of when to being using newer JS features (namely ES6) in the following tickets:

- #1089 - Coding standards
- #746 - Modularization of library

I am fine with dropping support for IE 8 but we have to be careful about IE 11 because that still has sizable market share in enterprise environments (unfortunately).

PR #1304, opened 2018-07-03
loose lists

### LINKED ISSUE #1298: Marked adds extra/unnecessary <p> tags to list elements
**Describe the bug**
When using lists with linebreak separation, Marked adds unnecessary &lt;p&gt; tags to list items.

**To Reproduce**
Using CommonMark, the following adds erroneous &lt;p&gt; tags around the BBB, DDD, and EEE elements.
Using MarkedDemo, the following adds &lt;p&gt; tags around AAA (correct) and BBB (incorrect).

```
## Testing

* AAA

* BBB
    * CCC
* DDD
* EEE
```

**Expected behavior**
With the above code, only AAA should have &lt;p&gt; element wrappers.

** Additional Example **
```
## Testing

* AAA
* BBB
    * CCC
* DDD

* EEE
```
Above, CommonMark adds around AAA, BBB, DDD, EEE.
Above, MarkedDemo adds around DDD, and EEE

UziTech:
**Marked version:** 0.4.0

## Description

Makes "loose" lists conform to the [standard](https://spec.commonmark.org/0.28/#loose), requiring all items in a loose list to be loose.

This conflicts with #1 in which @chjj takes his own approach to loose lists, which I don't think comes from any standard. My proposed solution to the failing non-standard tests would be to just remove the tests.

- Fixes #1298 

## Contributor

## Committer

- [ ] Merge PR


lib/marked.js
```diff
@@ -365,6 +372,10 @@ Lexer.prototype.token = function(src, top) {
           if (!loose) loose = next;
         }
 
+        if (loose) {
+          listStart.loose = true;
```
styfle: Can this `if` statement be removed in favor of `listStart.loose = loose`

UziTech: no because if `loose` is false  and `listStart.loose` is true I don't want it to change `listStart.loose` to false.

`listStart.loose` is true if **any** of `listStart` items are loose

UziTech:
Since I am removing the token `loose_item_start` in favor of a .loose property on a `list_item_start` token this will probably be a breaking change

lib/marked.js
```diff
@@ -1217,28 +1238,20 @@ Parser.prototype.tok = function() {
     }
     case 'list_item_start': {
       body = '';
+      var loose = this.token.loose;
```
styfle: Any reason why you create a new var here instead of using `this.token.loose` below?

UziTech: because `this.token` is changed when `this.next()` is called in the while loop below.

`body += !this.token.loose && this.token.type === 'text'` would be checking if the text token is loose not the list start token

styfle:
@joshbruce @davisjam Can you review this PR?

@Martii Can you check to see if this solves [your issue](https://github.com/markedjs/marked/issues/1308#issuecomment-404964542)?

Martii:
@styfle 

Maybe in a little bit I'll have some time... but the [Marked demo](https://marked.js.org/demo/?outputType=html&text=*%20bullet%0A*%20bullet%0A*%20bullet%0A%0A*%20bullet) seems to not be giving me a preview in a clean browser profile. The HTML source that is selectable looks different than I recall.

Martii:
And now it appeared for preview in a new tab... love quirks like this.

Martii:
> That being said, this PR is not released yet and therefore not available in the demo right now.

I know... that's why I [made the suggestion of having a current HEAD option](https://github.com/markedjs/marked/issues/1308#issuecomment-404966661). Was doing a bazillion things yesterday and never got around to testing the current HEAD myself.

> chrome extension that was causing an issue with the demo

...

> preview in a clean browser profile.

;) So it was something else... perhaps the CDN wasn't delivering? Anyhow... probably something for a different issue if it continues to crop up. :) Thanks.

PR #1331, opened 2018-09-06
use @markedjs/html-differ

### LINKED ISSUE #1330: Update dependencies

Security vulnerability due to `devDependencies` in `package.json`. It's on `lodash`, which is most likely the dependency of a dependency. While this doesn't negatively effect our users it might cause problems with automatic tracking databases that don't distinguish between dependencies (which Marked has none) and devDependencies (which Marked has a few).
 
Some type of dependency check protocol we can perform. Maybe on a somewhat rotating basis so one of us doesn't get stuck doing it all the time. Something captured in the contributing documentation; who knows, maybe someone from the community would be willing to step up to make sure we're up to date and whatnot...just spitballing.

UziTech:
**Marked version:** v0.5.0

## Description

Update dependencies to use `@markedjs/html-differ` instead of `html-differ`

- Fixes #1330 

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,13 +4,78 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@markedjs/html-differ": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/@markedjs/html-differ/-/html-differ-2.0.0.tgz",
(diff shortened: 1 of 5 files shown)


UziTech:
Looks like this should be good to go 🚢 👍

PR #1333, opened 2018-09-10
Add dependency badges to readme

### LINKED ISSUE #1330: Update dependencies

styfle:
## Description

This PR adds a couple badges to show out of date dependencies (none) and dev dependencies.

I also updated the gzip size badge to use badgen.net for consistency.

Fixes #1330

image

## Contributor

## Committer

- [ ] Merge PR


styfle:
@joshbruce How about this one?

PR #1339, opened 2018-09-22
Add CDN Support

fransallen:
I will skip the contribution template in order to just add CDN support for this project so people can easly use what kind of CDN provider they like. If you think this is useful, please accept this request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ Also read about:
 
 **In-browser:** `npm install marked --save`
 
+## CDN Support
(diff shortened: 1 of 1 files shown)


styfle:
JSDeliver already uses multiple CDNs behind the scenes and claims production support. 

Do the others claim you can use for production? 

Also, what about unpkg.com which is gaining popularity?

styfle:
Closing since there was no response.

PR #1346, opened 2018-09-27
Make Steve publisher

joshbruce:
@styfle 

## Recommendation to:

- [ ] Change user group
- [x] Add a badge
- [ ] Remove a badge

1. The one separated the docs from the code.
2. Always able to read when the community is ready for something and willing to make requests and discuss with the "right" people.
3. Always adding little things to make a better experience for all.
4. The selfish bit: Given where I am with my latest "growth spurt", I might end up having to hustle a bit harder; therefore, having three of us seems like a good idea - in order of fallback - me, @styfle, @UziTech, @chjj (pretty sure he's still watching lol)

We've done great as a crew and I cannot adequately express my gratitude to all that you do and have done. When I came in to do the release two or three back I was amazed at how smooth and complete it was. Well done.

## As the one mentioned, I would like to:

- [ ] accept the recommendation; or,
- [ ] graciously decline; or,
- [ ] dispute the recommendation

within 30 days, if you have not indicated which option you are taking one of the following will happen:

1. If adding a badge, we will assume you are graciously declining.
2. If removing a badge, we will assume you do not want to dispute the recommendation; therefore, the badge will be removed.

Note: All committers must approve via review before merging, the disapproving committer can simply close the PR.

### DIFF
--- a/docs/AUTHORS.md
+++ b/docs/AUTHORS.md
@@ -29,8 +29,8 @@ Marked takes an encompassing approach to its community. As such, you can think o
         </a>
         <br>
         <a href="https://www.ceriously.com">Steven</a>
-        <div>Admin</div>
-        <small>Open source, of course; GitHub Guru; Humaning Helper</small>
+        <div>Publisher</div>
+        <small>Release Wrangler; Dr. Docs; Open source, of course; GitHub Guru; Humaning Helper</small>
       </td>
     </tr>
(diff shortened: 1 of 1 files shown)


joshbruce:
@davisjam: Just a friendly poke. Would like to add @UziTech and @styfle at the same time if this gets approved.

Issue #1106, opened 2018-03-02
You're friend here is only mostly dead...


joshbruce:
**Marked version:** 0.3.17

See next: #1522

[![Your friend here is only mostly dead. And, mostly dead, is slightly alive](http://i.imgur.com/IYiVgOm.jpg)](https://youtu.be/xbE8E1ez97M)

^^ There's a YouTube video behind there

On December 1, 2017 we started working to analyze and bring marked back to life. Feedback in #956 was pretty positive; so, we kept pressing forward.

Around February 16, 2018 @chjj moved marked to the MarkedJS organizational account, giving us the opportunity to manage the project more effectively in a self-organizing and self-managing way.

Over the course of that time we have managed to:

1. Add continuous integration.
2. Analyzed, triaged, and closed hundreds of pull requests and issues.
3. Fix some of the defects (including another discovered security vulnerability).
4. And establish a basic roadmap for the future.

## Where we are and where we're going

We realize that, if this revival is going to work it cannot be one person (or even four people) doing most of the work; _it's simply not sustainable_. 

Therefore, we would like to let you know where we are, what the plan is (at this moment), and ask you to help any way you can (even if it's just fixing a typo). There are only four of us on the core team; therefore, we prefer that if you see something that needs fixed, please submit a PR that fixes it (this will let us focus on the bigger picture and problems).

1. The only thing that is time-sensitive on this revival effort is a security vulnerability. See [README](https://github.com/markedjs/marked/blob/master/README.md#security).
2. Our current test bed does not seem to adequately reflect or cover the [supported specifications](https://github.com/markedjs/marked/blob/master/README.md#specifications). If it were able to cover it, we would not have the number of tickets we do concerning not being compliant with those specs. Further, there may be areas of the specifications we will not be able to comply with for whatever reason (see #1036). Finally, we will be focusing our efforts on getting marked to be as spec-compliant as we can before focusing on user issues (as we believe spec-compliance will cover a lot of those issues). **If you would like to help, please comment below and we'll reach out when we're ready.**
3. Once we have performed the necessary analysis to get a more objective understanding of where we are regarding spec compliance, we may or may not turn our attention to the other user submitted issues.

## Until then

[...]

PR #1349, opened 2018-10-02
move to ESM and what not

### LINKED ISSUE #1288: ESM (ES6 module) distribution
**Describe the feature**

43081j:
👋 

I had this lying around and decided to finish it off recently in my fork as it was needed for trying out in some web component stuff I had setup.

Fixes #1288

Is this something you'd be interested in? Happy to keep it in my fork if not and close this PR.

---

Here's a summary of what happened in this branch:

* Sources separated into modules
* All sources are valid ESM (`src/*`)
* Moved from prototypical concepts to classes
* Introduced babel to transpile to CJS modules (`lib/*`) for when node users require this
* Introduced rollup to bundle to a browser-compatible library (`marked.min.js`) which behaves the same as the original
* Introduced `files` to package.json so `npm pack` and publish will package only bin/lib/man (CJS sources)
* Introduced a `module` entry to package.json so bundlers can resolve the ESM sources
* Replaced all unscoped vars and what not to `let`/`const`

All tests pass, build succeeds.

Have not been using it in a browser myself so it seems that would need further testing.

Whats missing:

* The browser bundle isn't minified (something like `babel-preset-minify` could achieve that)
* Performance comparison
* Build will fail on CI as `npm run build` needs adding to travis.yml

---

Again, if this is of any use to you just let me know and ill get back on it.

cc @joshbruce

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,0 @@
-.git*
-test/
-docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ jobs:
 
   include:
     - stage: unit tests 👩🏽‍💻
-      script: npm run test:unit
+      script: npm run build && npm run test:unit
       node_js: lts/*
 
     - stage: spec tests 👩🏽‍💻
-      script: npm run test:specs
+      script: npm run build && npm run test:specs
       node_js: v0.10
     - node_js: v4
     - node_js: lts/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ jobs:
 
   include:
     - stage: unit tests 👩🏽‍💻
-      script: npm run build && npm run test:unit
+      script: npm run test:unit
       node_js: lts/*
 
     - stage: spec tests 👩🏽‍💻
-      script: npm run build && npm run test:specs
+      script: npm run test:specs
       node_js: v6
     - node_js: lts/*
     - node_js: node
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,7 @@ jobs:
 
     - stage: spec tests 👩🏽‍💻
       script: npm run build && npm run test:specs
-      node_js: v0.10
-    - node_js: v4
+      node_js: v6
     - node_js: lts/*
     - node_js: node
 
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "lint": "eslint --fix bin/marked \"src/**/*.js\"",
     "build:module": "npm run lint && babel src -d lib",
     "build": "npm run build:module && rollup -c",
-    "test": "jasmine --config=jasmine.json",
+    "test": "npm run build && jasmine --config=jasmine.json",
     "test:unit": "npm test -- test/unit/**/*-spec.js",
     "test:specs": "npm test -- test/specs/**/*-spec.js",
     "test:cm": "npm test -- test/specs/commonmark/**/*-spec.js",
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "author": "Christopher Jeffrey",
   "version": "0.5.1",
   "main": "./lib/marked.js",
+  "module": "./src/marked.js",
   "bin": "./bin/marked",
   "man": "./man/marked.1",
   "repository": "git://github.com/markedjs/marked.git",
@@ -23,24 +24,32 @@
     "html"
   ],
   "devDependencies": {
+    "@babel/cli": "^7.1.2",
+    "@babel/core": "^7.1.2",
+    "@babel/preset-env": "^7.1.0",
+    "@markedjs/html-differ": "^2.0.0",
+    "babel-eslint": "^10.0.1",
+    "babel-plugin-add-module-exports": "^1.0.0",
     "commonmark": "0.x",
-    "eslint": "^4.19.1",
-    "eslint-config-standard": "^11.0.0",
+    "eslint": "^5.6.1",
+    "eslint-config-standard": "^12.0.0",
     "eslint-plugin-import": "^2.14.0",
-    "eslint-plugin-node": "^5.2.1",
-    "eslint-plugin-promise": "^3.8.0",
-    "eslint-plugin-standard": "^3.1.0",
+    "eslint-plugin-node": "^7.0.1",
+    "eslint-plugin-promise": "^4.0.1",
+    "eslint-plugin-standard": "^4.0.0",
     "eslint-plugin-vuln-regex-detector": "^1.0.4",
     "front-matter": "^2.3.0",
     "glob-to-regexp": "0.3.0",
-    "@markedjs/html-differ": "^2.0.0",
     "jasmine": "^3.2.0",
     "jasmine2-custom-message": "^0.9.3",
     "markdown": "0.x",
     "markdown-it": "8.x",
-    "uglify-js": "^3.4.8"
+    "rollup": "^0.66.2"
   },
   "scripts": {
+    "lint": "eslint --fix bin/marked \"src/**/*.js\"",
+    "build:module": "npm run lint && babel src -d lib",
+    "build": "npm run build:module && rollup -c",
     "test": "jasmine --config=jasmine.json",
     "test:unit": "npm test -- test/unit/**/*-spec.js",
     "test:specs": "npm test -- test/specs/**/*-spec.js",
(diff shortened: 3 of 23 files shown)


.npmignore
```diff
@@ -1,3 +0,0 @@
-.git*
-test/
-docs
```
styfle: Why was .npmignore removed?

package.json
```diff
@@ -51,11 +60,15 @@
...
   },
   "engines": {
     "node": ">=0.10.0"
-  }
+  },
+  "files": [
+    "bin/**",
+    "lib/**",
+    "man/**",
+    "README.md"
```
styfle: I see now, this is a better solution than npmignore

.npmignore
43081j: Yup the recommended method of defining what gets packaged is the `files` array now and thngs like `node_modules` are automatically excluded.

package.json
43081j: Note to self: this needs `src/**` adding for compatibility with browsers importing the ESM source, though.

package.json
```diff
@@ -51,11 +60,15 @@
     "test:lint": "eslint bin/marked .",
     "test:redos": "eslint --plugin vuln-regex-detector --rule '\"vuln-regex-detector/no-vuln-regex\": 2' lib/marked.js",
     "bench": "node test --bench",
-    "lint": "eslint --fix bin/marked .",
-    "build": "uglifyjs lib/marked.js -cm  --comments /Copyright/ -o marked.min.js",
```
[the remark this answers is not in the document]
43081j: the rollup call produces it, unminified at the min but introducing `babel-preset-minify` should sort that.

styfle:
The [tests](https://travis-ci.org/markedjs/marked/builds/436376656) are failing with Error: Cannot find module './lexer'

43081j:
Yup this is because the travis config needs updating to `npm run build` beforehand.

styfle:
@43081j You should be able to modify `.travis.yml` and add a build step

43081j:
Yup its part of the list in my original post of whats left to do. simple enough, just didn't get around to it

styfle:
Excellent! Let's wait to see what the rest of the team thinks.

.travis.yml
```diff
@@ -7,13 +7,12 @@ jobs:
 
   include:
     - stage: unit tests 👩🏽‍💻
-      script: npm run test:unit
+      script: npm run build && npm run test:unit
```
UziTech: since `npm test` already includes `npm run build` it is redundant here and below

package.json
```diff
@@ -23,39 +24,53 @@
...
-    "test:redos": "eslint --plugin vuln-regex-detector --rule '\"vuln-regex-detector/no-vuln-regex\": 2' lib/marked.js",
+    "test:lint": "eslint bin/marked \"test/**/*.js\"",
+    "test:redos": "eslint --plugin vuln-regex-detector --rule '\"vuln-regex-detector/no-vuln-regex\": 2' src/marked.js",
     "bench": "node test --bench",
-    "lint": "eslint --fix bin/marked .",
-    "build": "uglifyjs lib/marked.js -cm  --comments /Copyright/ -o marked.min.js",
     "preversion": "npm run build && (git diff --quiet || git commit -am 'minify')"
   },
   "engines": {
     "node": ">=0.10.0"
```
UziTech: should this be updated to `">=6.0.0"`?

UziTech:
Do we need to add [browserslist](https://babeljs.io/docs/en/babel-preset-env#browserslist-integration) to `package.json` for `babel-preset-env`?

43081j:
We probably should but i wasn't too sure what the correct targets should be.

The ESM sources account for any browsers supporting ES modules.

The CJS sources should only be used by node, so our babel-preset config should probably be `node: true`.

The bundle should be consumed by non-ESM browsers so its actually a different target to `lib/` I guess (overridable via rollup.config.js if we must). We would have to find the minimum browser version we could support.

package.json
```diff
@@ -23,39 +24,53 @@
...
     "test:specs": "npm test -- test/specs/**/*-spec.js",
     "test:cm": "npm test -- test/specs/commonmark/**/*-spec.js",
     "test:gfm": "npm test -- test/specs/gfm/**/*-spec.js",
     "test:marked": "npm test -- test/specs/marked/**/*-spec.js",
     "test:old": "node test",
-    "test:lint": "eslint bin/marked .",
-    "test:redos": "eslint --plugin vuln-regex-detector --rule '\"vuln-regex-detector/no-vuln-regex\": 2' lib/marked.js",
+    "test:lint": "eslint bin/marked \"test/**/*.js\"",
+    "test:redos": "eslint --plugin vuln-regex-detector --rule '\"vuln-regex-detector/no-vuln-regex\": 2' src/marked.js",
     "bench": "node test --bench",
```
UziTech: `"test:old"` and `"bench"` will need to be prepended with `npm run build &&`

rollup.config.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default {
+  input: 'src/marked.js',
+  output: {
+    file: 'marked.min.js',
+    format: 'umd',
+    name: 'marked'
+  },
+  plugins: [babel()]
+}
```
UziTech: Missing semicolon (sorry but my IDE nags me 🙃)

UziTech:
benchmarks seem to be consistent 👍:

```
// master
marked completed in 5809ms.
marked (gfm) completed in 7043ms.
marked (pedantic) completed in 5590ms.
commonmark completed in 8877ms.
markdown-it completed in 6774ms.
markdown.js completed in 22230ms.

// this pr
marked completed in 6027ms.
marked (gfm) completed in 6981ms.
marked (pedantic) completed in 5792ms.
commonmark completed in 9057ms.
markdown-it completed in 8349ms.
markdown.js completed in 20882ms.
```

43081j:
What sort of time frame would you be looking at for 1.x?

To be honest, ES modules are still a bit flaky with node which is why we're forced to have 3 builds (ESM, CJS, browser).

I think if we try wait for node to have a better implementation, we'll be here _a while_.

It may be worth us looking into using this:
https://github.com/standard-things/esm

instead of shipping two copies of the sources 🤔  but its a fairly big dependency to be shipping i guess

styfle:
I looked this over again and I am becoming more reserved the more I look at it.

> ES modules are still a bit flaky with node which is why we're forced to have 3 builds (ESM, CJS, browser).

This PR is adding more complexity to a relatively small project and I'm not sure the value added here is worth it at this time.

Let's do a pros/cons list

## Pros

1. Developers get to use multiple files - maybe avoid merge conflicts?
2. Developers get to use ES6 syntax - less writing and more readable in most cases
3. (maybe) Users can use ESM sources with bundlers to maybe treeshake output

## Cons

1. The ESM source will not run in the browser as-is (due to bare specifiers)
2. The ESM source will not run in Node.js as-is (due to .mjs and experimental)
3. The bundle size is actually larger by a few KB (was 42 KB and now 47 KB)
4. We'll be publishing more sources files so removing those sources is a breaking change (probably can do this in 1.x)
5. More complexity due to more build tools requires maintenance when upgrading versions: babel, rollup, and several more plugins
6. (maybe) Some js user's don't know ES6 syntax and won't be able to contribute
7. (maybe) Build time will increase slightly due to more build targets

joshbruce:
@styfle brings up some very good points.

Tagging #746 as it describes a method for modularizing the codebase with straight ES5 or 6 (??) (doing this off the top of my head).

To @43081j's question, the core team is really trying to get the community to step up more. One of the issues of the past (before the current core team) was an expectation became set that the core team would do everything...so I cannot tell you how awesome it is to have you post this PR, whether it's accepted or not...anyhoo, the core team primarily sets direction, creates tools, and ensures consistency of the codebase and approach; so, all that to say, timeframe is up to the broader community as the core team will focus more on helping and coordination than doing (though we do some of that when we can). A refrain that became popular for a minute was "1.0 could happen tomorrow or may never happen, it's up to the community" (or something like that).

See https://github.com/markedjs/marked/issues/1106 for the details and drama...in fact, probably time to write another one.

**Throwing copper:** I think, based on @styfle's thoughts, we should close this for now - reference to what I think is the main modularization issue - and maybe resurrect at that time. Spec-compliance and keeping the infrastructure clean per the roadmap and milestones (https://github.com/markedjs/marked/milestones) unless we find something that can minimize impact on legacy users or out right fake backward compatibility. Not to say the new stuff isn't nice, just may not be ready to be part of main..

43081j:
This is meant to introduce the ability to use the ES modules in browsers _as is_. So if that means I forgot to (and need to) add the file extension to each import, I should.

The bundle size is _of course_ larger, it isn't even minified yet (and rollup will tree-shake, which it hasn't done yet either).

The rest of @styfle's points are fairly small IMHO, tooling being introduced isn't a problem, nor is build time (it won't be noticeable most likely). Git history shouldn't be an issue either as the files should be detected as being moved.

Martii:
> Developers get to use ES6 syntax - less writing and more readable in most cases

This is actually a con in our book. ES6 may be supported however it's *very subjective on readability*. Examples written in ES5 are a good base for projects that don't use ES6+ and extending it to the more quirky ES6+ syntaxes.

PR #1350, opened 2018-10-04
Make URL case insensitive.

roygold7:
Now URL's starting with hTTp or Https will be captured by the regular expression.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -647,6 +647,7 @@ function InlineLexer(links, options) {
   } else if (this.options.gfm) {
     if (this.options.breaks) {
       this.rules = inline.breaks;
+      this.rules.url = new RegExp(this.rules.url, "i");
     } else {
       this.rules = inline.gfm;
     }
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -647,7 +647,7 @@ function InlineLexer(links, options) {
   } else if (this.options.gfm) {
     if (this.options.breaks) {
       this.rules = inline.breaks;
-      this.rules.url = new RegExp(this.rules.url, "i");
+      this.rules.url = new RegExp(this.rules.url.source, "i");
     } else {
       this.rules = inline.gfm;
     }
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -647,7 +647,7 @@ function InlineLexer(links, options) {
   } else if (this.options.gfm) {
     if (this.options.breaks) {
       this.rules = inline.breaks;
-      this.rules.url = new RegExp(this.rules.url.source, "i");
+      this.rules.url = new RegExp(this.rules.url.source, 'i');
     } else {
       this.rules = inline.gfm;
     }
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -606,7 +606,7 @@ inline.pedantic = merge({}, inline.normal, {
 inline.gfm = merge({}, inline.normal, {
   escape: edit(inline.escape).replace('])', '~|])').getRegex(),
   _extended_email: /[A-Za-z0-9._+-]+(@)[a-zA-Z0-9-_]+(?:\.[a-zA-Z0-9-_]*[a-zA-Z0-9])+(?![-_])/,
-  url: /^((?:ftp|https?):\/\/|www\.)(?:[a-zA-Z0-9\-]+\.?)+[^\s<]*|^email/,
+  url: /^((?:ftp|https?):\/\/|www\.)(?:[a-zA-Z0-9\-]+\.?)+[^\s<]*|^email/i,
   _backpedal: /(?:[^?!.,:;*_~()&]+|\([^)]*\)|&(?![a-zA-Z0-9]+;$)|[?!.,:;*_~)]+(?!$))+/,
   del: /^~+(?=\S)([\s\S]*?\S)~+/,
   text: edit(inline.text)
--- a/test/unit/marked-spec.js
+++ b/test/unit/marked-spec.js
@@ -25,3 +25,13 @@ describe('Test paragraph token type', function () {
     expect(tokens[7].type).toBe('text');
   });
 });
+
+describe('Test url token type', function () {
+  it('should use the "url" type', function () {
+    const md = 'HtTps://www.testsite.com';
+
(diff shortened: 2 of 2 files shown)


styfle:
@roygold7 Thanks for the PR! Can you add a unit test for an uppercase url?

@UziTech Do you think this will affect the `email` replacement?

@davisjam I don't think this will introduce ReDOS but please review just in case.

lib/marked.js
```diff
@@ -647,6 +647,7 @@ function InlineLexer(links, options) {
   } else if (this.options.gfm) {
     if (this.options.breaks) {
       this.rules = inline.breaks;
+      this.rules.url = new RegExp(this.rules.url.source, "i");
```
styfle: The lint step is failing because we expect single quotes: `new RegExp(this.rules.url.source, 'i');`

roygold7: Thanks!

davisjam:
Case-insensitivity does not affect this regex's behavior w.r.t. ReDoS.

lib/marked.js
```diff
@@ -606,7 +606,7 @@ inline.pedantic = merge({}, inline.normal, {
 inline.gfm = merge({}, inline.normal, {
   escape: edit(inline.escape).replace('])', '~|])').getRegex(),
   _extended_email: /[A-Za-z0-9._+-]+(@)[a-zA-Z0-9-_]+(?:\.[a-zA-Z0-9-_]*[a-zA-Z0-9])+(?![-_])/,
-  url: /^((?:ftp|https?):\/\/|www\.)(?:[a-zA-Z0-9\-]+\.?)+[^\s<]*|^email/,
+  url: /^((?:ftp|https?):\/\/|www\.)(?:[a-zA-Z0-9\-]+\.?)+[^\s<]*|^email/i,
```
styfle: Now that this is case-insensitive, could you also change `a-zA-Z` to `a-z`?

styfle:
@UziTech I think it already is invoked as case-insensitive

UziTech:
looks like this could be solved easier by changing line 618 from 

```js
inline.gfm.url = edit(inline.gfm.url)
```

to 

```js
inline.gfm.url = edit(inline.gfm.url, 'i')
```

Martii:
Careful on this PR... See https://url.spec.whatwg.org/#url-writing and the keyword "not" floating around in there for certain schemes.

Martii:
Here's at least one where I am thinking it should be double checked and possibly added to the tests:

> a URL-scheme string that is _not_ an ASCII case-*in*sensitive match for a special scheme, followed by U+003A (:) and a relative-URL string

Removing the double negative *(emphasized font in immediate previous quote)* ... it seems like this use case may need case sensitivity:

> a URL-scheme string that is an ASCII case-sensitive match for a special scheme, followed by U+003A (:) and a relative-URL string

In this browser the Address bar does take a mix of upper case absolute URLs but then copying it out it always lower cases it... so that's how this browser handles it at least.

UziTech:
The spec says it "must be **one** of the following" and ftp/http(s) matches

> a URL-scheme string that is an ASCII case-insensitive match for a special scheme and not an ASCII case-insensitive match for "file", followed by U+003A (:) and a scheme-relative-special-URL string

UziTech:
also looks like github doesn't mind if it is not lowercase.

hTtP://example.com

Martii:
> also looks like github doesn't mind if it is not lowercase.

Doesn't mean that GFM is following output guidelines from the W3 for well written HTML code conformance.

The decisions are these:

1. Fix a nuisance and change the incomplete specification to a potentially larger breaking change *(this PR and a move to 0.6.0 of marked with all use case scenarios)*. This will trickle up the chain to all sanitizers and top-level code *(which is why I'm commenting here)*. Quite frankly in my circles writing `hTTp`, or whatever variant, is poor coding and communication practice but the W3 says it's okay in certain circumstances. \*shrugs\*. Wonder how *node* handles this with plucking individual pieces out in `URL` API?
2. Check the input to see if there is a relative url and conditionalize the insensitive vs. sensitive operation to spec... potentially smaller breaking *(only a use case scenario)* but additive towards the compliance of the spec instead of subtraction. e.g. standard I/O... check the Input for potentially bad input and output the correct syntax specification. The text can always include the mult-cased `http` but the url itself should be normalized in this scenario.
3. Keep it as is ... still incomplete specification however unbreaking atm.

We'll accommodate regardless if this goes through but filtering will be affected with items 1 and 2. This is probably why item 3 exists is it is a balance perhaps. Hence why I said be careful and haven't bothered to vote prematurely either way.

UziTech:
> Doesn't mean that GFM is following output guidelines from the W3 for well written HTML code conformance.

This only affects inline url when `gfm` is turned on (which means they want it to act like GitHub).

Martii:
> act like GitHub

One place this could be addressable is in github/cmark however there's no real unit test, that I know of, since the W3 spec mentions bases and relative urls which is the use case exception. I'm not against this PR just noting what effects will happen with it. Filtering is where we will need to adjust to this forcing it to lowercase most likely. Still need to test this *(when I get back to dev station next week)* in *node* with the `URL` API and see if it breaks it natively with "parting it out"... it may not affect it at all or could affect it adversely. Quite honestly I've never seen anyone do a mix of upper and lower case schemes in another language including JavaScript so it caught my primary attention. GFM will most likely be normalized if it isn't. Should probably test in Commonmark too but again don't know if they have a base url floating around to test against from my understanding. :)

styfle:
This works just fine:

```js
const str = 'hTtP://example.com';
const url = new URL('hTtP://example.com');
console.log('node\t', process.version);
console.log('input\t', str);
console.log('url\t', url.href);
```

https://glot.io/snippets/f5if7pn2lf

Martii:
A little more on the test range:

``` sh-session
$ node -v && node -e 'var x = new URL("/doc/this.html", "hTtP://example.com"); console.log(x.protocol)'
v10.11.0
http:
```
... this passes. :+1: 

Haven't had a chance to test the legacy `URL` API yet and probably older versions of *node* that are still active... little later perhaps if I have time this evening.

styfle:
Closing in favor of #1384

PR #1343, opened 2018-09-25
add LICENSE file

danmaas:
Hi - I noticed that the GitHub API, and other open-source indexes, are not recognizing the proper license on this repo because there is no LICENSE file. This PR adds the LICENSE.

PR #1355, opened 2018-10-17
Add a LICENSE file so that lawyers can be happy

chrisjshull:
## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,20 @@
+The MIT License
+
+Copyright (C) 2011-2018, Christopher Jeffrey
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


styfle:
Hi Chris, this looks like a duplicate of #1343

chrisjshull:
Hmm. docs/ is excluded from npm by .npmignore, so it is not being distributed.
Also, automated tools expect LICENSE to be in a consistent place.

joshbruce:
Maybe we should figure something out about this license thing. The fact that two people have submitted PRs to make the structure more like that commonly found on GitHub makes me think there's a UX goal we're missing.

styfle:
Agreed. Let's move the file back instead of creating a new file. That way git log will follow the history.

@chrisjshull Would you like to create a new PR?

chrisjshull:
Don't actually have this checked out, just did in the GH UI, but not seeing the option to move.
Can someone else?

Martii:
@joshbruce 

> Maybe we should figure something out about this license thing.

LICENSE can be [specific to GH](https://help.github.com/articles/adding-a-license-to-a-repository/) to show in a web browser immediately what the licensing is and the file itself is used for checkouts as common practice with OSS. It's quite picky with dual/double licensing but single licensing works well. 

The type of license should be same as by it's [SPDX short name](https://spdx.org/licenses/MIT.html) at [/markedjs/marked/blob/`4dce02d`/package.json#L14](https://github.com/markedjs/marked/blob/4dce02d12e19074cfc6fe89df536e600770fde63/package.json#L14) which is [*npm* specific](https://docs.npmjs.com/files/package.json#license).

These are usually the first two places we look when adopting a package. :)

Martii:
@chrisjshull 

Clone your fork of this repo and then checkout the branch at https://github.com/chrisjshull/marked/tree/patch-1 ...  then you can undo and move using `git mv source target`.

PR #1357, opened 2018-10-17
Add the ability to specify options on the demo page as JSON

aprotim:
## Description

Create a details toggle that hides a textarea into which a user can type a JSON-formatted options object.

(For some reason, I can't seem to get the textarea inside a details tag to match the width of the other textareas. After staring at the CSS and banging my head against it, I figured I'd let other people look at it, decide if it was important, and if it was, how to fix it.)

 
screenshot from 2018-10-17 01-27-32

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/docs/demo/demo.css
+++ b/docs/demo/demo.css
@@ -49,7 +49,7 @@ header h1 {
 	resize: both;
 }
 
-.pane, #input {
+.pane, .inputPane {
 	margin-top: 5px;
 	padding: 0.6em;
 	border: 1px solid #ccc;
--- a/docs/demo/demo.css
+++ b/docs/demo/demo.css
@@ -45,6 +45,10 @@ header h1 {
 	box-sizing: border-box;
 }
 
+#options {
+	resize: both;
+}
+
 .pane, #input {
 	margin-top: 5px;
 	padding: 0.6em;
--- a/docs/demo/demo.js
+++ b/docs/demo/demo.js
@@ -49,7 +49,14 @@ if ('text' in search) {
 if ('options' in search) {
   $optionsElem.value = search.options;
 } else {
-  $optionsElem.value = JSON.stringify(marked.getDefaults(), null, ' ');
+  $optionsElem.value = JSON.stringify(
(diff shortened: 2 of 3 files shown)


UziTech:
I like being able to add the options but could we move it to the right list

image

aprotim:
I can do that, although right now all the things on the left are output/read-only, while the right side is input. What if instead of the details toggle, I added a similar dropdown on the input side? (Example of what it looks like forthcoming.)

aprotim:
How about this: now all input is on the left, all output is on the right, and both have parallel type selectors.

aprotim:
(FWIW, you can look at the live version at https://aprotim.github.io/marked/demo/)

styfle:
This is awesome!

When the page loads the first time and there are no options in the query string, can you do `$optionsElem.value = JSON.stringify(marked.getDefaults(), null, ' ');`

This will give people an idea of what options are available 👍

UziTech:
@aprotim looks like your demo page has an error

```
this.renderer.text is not a function
```

seems to be because the options are duplicated in the "renderer" option

```
...
"renderer": {
  "options": {
   "baseUrl": null,
   "breaks": false,
   "gfm": true,
   "headerIds": true,
   "headerPrefix": "",
   "highlight": null,
   "langPrefix": "language-",
   "mangle": true,
   "pedantic": false,
   "renderer": {},
   "sanitize": false,
   "sanitizer": null,
   "silent": false,
   "smartLists": false,
   "smartypants": false,
   "tables": true,
   "xhtml": false
  }
 },
...
```

aprotim:
Ah, yeah. This is a problem with the roundtrip through JSON, when the default value for "renderer" is a complex object, not a dumb data object. I can either remove renderer from the object before stringifying, or I can set the default options text to be some manually-curated subset... let me tinker with it for a sec.

aprotim:
Ok, now we filter any complex objects (or at least any objects whose prototype is different than Object's) out when we stringify. Seems to have cleared up the problem.

Issue #1232, opened 2018-04-19
Sanitize and sanitizer


joshbruce:
**Marked version:** 0.3.19

**Markdown flavor:** n/a

**Proposal type:** other

## What pain point are you perceiving?

Marked is taking on too much; or, at least more than it needs to.

## What solution are you suggesting?

It appears at some point someone had the idea of adding sanitization to marked. Later, someone else had the idea of letting people plug in their own sanitizer.

I suggest we go for the latter and deprecate the former. Turning on sanitize doesn't really seem to give us much in all honesty.

UziTech:
I say we remove sanitizing all together and recommend `DOMPurify.sanitize(marked(...))`

joshbruce:
@Martii: Thanks for chiming in as well. Right now we don't have a direct way to contact users. I was hoping more people would have added themselves to the "users" portion of the Authors page, to be honest. So, we're counting on the tickets, milestones, and projects to give people the warning.

We'll probably spend most of 1.x preparing people to lose things.

@UziTech: Agreed.

Tagging #1114

alystair:
If you're discussing escaping HTML... here's a solution that utilizes a browsers own filtering. If it breaks it means there's a browser vulnerability and you made yourself some nice pocket change via their  vulnerability rewards program.
```js
var escape = document.createElement('textarea');
function escapeHTML(html) { escape.textContent=html; return escape.innerHTML; }
function unescapeHTML(html) { escape.innerHTML=html; return escape.textContent; }
```
This doesn't solve the XSS vuln in the comment above but does for plenty of other things, assuming one doesn't want HTML code passthrough :)

styfle:
Here's another 3rd party sanitizer I just found: https://github.com/rhysd/marked-sanitizer-github

I can't say how secure it is but the point remains that marked doesn't need to ship with a sanitizer.

OrKoN:
I think it's nice if marked has sanitization built-in.

I tried to use `dompurify` but I find it too heavy and slow for Lambda environment/nodejs. I have tried `sanitize-html` but it seems to be simply removing everything that is not allowed and strips class names produced by highlight.js so requires a lot of tuning/testing.

UziTech:
We will most likely have a plug-in to do sanitization with a third party library in the future to make it easy for people to use, but it doesn't make sense for core marked to have sanitization built in because it is not part of the markdown specs.

UziTech:
[...]

PR #1361, opened 2018-10-19
Allow url protocol block/allow lists

aprotim:
**Markdown flavor:** all

## Description

Add the ability to set lists of allowable or blocked url protocols. Replaces the current "sanitize" code for URLs with explicit controls. (Existing behavior for sanitize is retained, for compatibility, but if/when sanitize is removed generally, urls can still be restricted to only desired protocols.

The included tests have examples.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1361,7 +1361,7 @@ function urlProtocolMatch(href, protocols) {
     .replace(/[^\w:]/g, '')
     .toLowerCase();
 
-  return protocols.some(function(protocol) { return prot.startsWith(protocol + ':')});
+  return protocols.some(function(protocol) { return prot.startsWith(protocol + ':'); });
 }
 
 function resolveUrl(base, href) {
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1593,7 +1593,7 @@ marked.getDefaults = function () {
     tables: true,
     xhtml: false,
     blockedUrlProtocols: null,
-    allowedUrlProtocols: null,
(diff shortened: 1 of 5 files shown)


aprotim:
9 commits later, I got the tests to finally pass.

styfle:
Hi @aprotim thanks again for the PR

This looks like it's adding new options which is not the desire for marked at this time. Especially since we are going to remove sanitization in the future (#1232).

A block-list or allow-list can be implemented outside of marked using something like `dompurify` or `sanitize-html`.

## Example sanitize

Here's an example of using an allow-list.

```js
var allowList = ["http", "https", "ftp", "mailto"];
var dirty = marked(md);
var clean = sanitizeHtml(dirty, {
  transformTags: {
    'a': function(tagName, attribs) {
        if (attribs.href && allowList.some(p => attribs.href.startsWith(p))) {
            delete attribs.href;
        }
        return { tagName: tagName, attribs: attribs };
    }
  }
});
```

aprotim:
I understand the desire to limit feature creep, to which end I've filed #1362.

I *would* like to point out that there's a lot of value of sanitizing the *input* and not (just) the output: keeping up with the ins and outs of the complex ways HTML can be abused is a harder problem than restricting the already smaller subset of functionality of markdown. (As far as I can tell, the vast majority, if not all, markdown-related XSS vulnerabilities could be avoided by disallowing embedded HTML and only allowing URLs from a small set of URL schemas.)

PR #1366, opened 2018-10-28
Fix builds: remove node v0.10 from travis matrix

styfle:
## Description

Fixes our failing builds.

Node v0.10 went End-of-Life two years ago which is why jasmine (our test framework) won't work anymore (looks like they dropped support awhile ago and just introduced breaking changes recently.

Related: https://github.com/jasmine/jasmine-npm/pull/144

## Contributor

- [x] no new tests required for this PR.

## Committer

- [ ] Merge PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ jobs:
 
     - stage: spec tests 👩🏽‍💻
       script: npm run test:specs
-    - node_js: v4
+      node_js: v4
     - node_js: lts/*
     - node_js: node
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ jobs:
 
     - stage: spec tests 👩🏽‍💻
       script: npm run test:specs
-      node_js: v0.10
     - node_js: v4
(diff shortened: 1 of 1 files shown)


styfle:
@UziTech That's a great question 🤔 

I think we could say this PR itself is not breaking because it doesn't actually break compatibility with older versions of node, its just a change to CI.

However, our next release that contains breaking changes (0.6.0) should mention that node v0.10 is no longer supported. 

It's a loop-hole but it will get CI green again and get the patch release out (0.5.2)

davisjam:
1. By itself dropping node v0.10 from CI is not a breaking change.
2. But yes it's a bit of a loophole, and could lead to breaking changes in the future.

Good candidate for release notes.

styfle:
@davisjam I will definitely add to release notes once you approve 😄

styfle:
I couldn't find if we officially support any browser/versions though. I know we don't run automated tests in any browsers. So my best way to communicate v0.10 as a browser is IE. Thoughts on this one?

PR #1367, opened 2018-10-28
Add files to package.json

styfle:
## Description

This will prevent publishing dev files to npm so that marked users who run `npm install marked` will not need to download any unnecessary files.

## Contributor

- [x] no new tests required for this PR.

## Committer

- [ ] Merge PR


UziTech:
I like it. That way we won't have to remember to add to .npmignore if we add other config files

Issue #123, opened 2013-02-22
GFM: Shouldn't require blank line before list


adam-p:
On Github, a blank line isn't required between paragraph text and a following list. For example:

```
I present this list:
* thing one
* thing two
```

I present this list:
- thing one
- thing two

But Marked gives:

```
> marked('I present this list:\n* thing one\n* thing two', {gfm: true, breaks: true})
'<p>I present this list:<br><em> thing one\n</em> thing two</p>\n'
```

I know Marked's behaviour is MD-standard,  but... GFM.

(Motivation: A user of [Markdown Here](https://github.com/adam-p/markdown-here) came across this organically, was confused, and asked me why it was happening.)

PR #1374, opened 2018-11-07
Fix Heading ID for Unicode text

k-utsumi:
I think defining `nonWordCharacters` will solve it.

Reference: [atom nonWordCharacters](https://github.com/atom/atom/blob/080d6ae1279c66e5f018d803186a0cde5c202c03/src/config-schema.js?fbclid=IwAR0BI3l0KgHls4gGlFw0HYQ_G_U7A84X0eTtmSvfEauo8NCnQxpRKuGJnQM#L450)

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -952,7 +952,7 @@ Renderer.prototype.heading = function(text, level, raw) {
       + level
       + ' id="'
       + this.options.headerPrefix
-      + raw.toLowerCase().replace(/[^\w]+/g, '-')
+      + raw.toLowerCase().replace(/[/\\()"':,.;<>~!@#$%^&*|+=[\]{}`?\-…_\s\t]+/g, '-')
       + '">'
       + text
       + '</h'
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -952,7 +952,7 @@ Renderer.prototype.heading = function(text, level, raw) {
       + level
       + ' id="'
       + this.options.headerPrefix
-      + raw.toLowerCase().replace(/[/\\()"':,.;<>~!@#$%^&*|+=[\]{}`?\-…_\s\t]+/g, '-')
+      + raw.toLowerCase().replace(/["\s]+/g, '-')
       + '">'
       + text
       + '</h'


UziTech:
according to the [spec](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute):

> When specified on HTML elements, the id attribute value must be unique amongst all the IDs in the element's tree and must contain at least one character. The value must not contain any ASCII whitespace.

so couldn't we change it to `/\s+/g`?

UziTech:
I suppose we would want to remove `"` as well so `/["\s]+/g`

All other characters seem to work fine in ids on chrome.

k-utsumi:
Maybe, need to consider the behavior of shared URL with fragment (section ID).

e.g. `## how to resolve #123 ? or ...`

styfle:
This feature (header id) is being removed in PR #1354 so I don't think it's worth pursuing in this PR.

k-utsumi:
I want to eliminate this problem quickly, so I reopened with a simple code adjustment.
If merged, I’ll make same PR for other services I use.

UziTech:
@styfle do we want to add a `seenIds` array to prevent collisions on this pr?

styfle:
@UziTech If we are not going to remove it, we could go all-out and implement like so:

```js
function Slugger () {	
  this.seen = {};	
}	

Slugger.prototype.slug = function (value) {	
  var slug = value	
    .toLowerCase()	
    .trim()	
    .replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,./:;<=>?@[\]^`{|}~]/g, '')	
    .replace(/\s/g, '-');	
  var count = this.seen.hasOwnProperty(slug) ? this.seen[slug] + 1 : 0;	
  this.seen[slug] = count;	
   if (count > 0) {	
    slug = slug + '-' + count;	
  }	
   return slug;	
};
```

I think we'll need some unit tests to see what will and will not match as far as unicode goes.

That was one of the reasons why I was trying to tear it out, so the implementation and tests could live in `github-slugger` package.

UziTech:
I think not having dependencies is one of the big benefits of marked so if we want to have ids be part of gfm we will need to incorporate slugger into our code.

k-utsumi:
@styfle 
Since I can't understand the change correctly,
Could you edit the code directory or create a new PR?

styfle:
Closing in favor of #1401

PR #1384, opened 2018-12-05
Make autolinks case insensitive

UziTech:
**Marked version:** master

## Description

Make urls case insensitive

implements https://github.com/markedjs/marked/pull/1350#issuecomment-427164249

```js
marked("hTtP://fOo.CoM");
// <p><a href=\"hTtP://fOo.CoM\">hTtP://fOo.CoM</a></p>
```

- closes #1350

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/test/specs/marked/marked.json
+++ b/test/specs/marked/marked.json
@@ -85,6 +85,18 @@
   },
   {
     "section": "Autolinks",
+    "markdown": "HTTP://FOO.COM",
+    "html": "<p><a href=\"HTTP://FOO.COM\">HTTP://FOO.COM</a></p>",
+    "example": 15
+  },
+  {
+    "section": "Autolinks",
+    "markdown": "hTtP://fOo.CoM",
+    "html": "<p><a href=\"hTtP://fOo.CoM\">hTtP://fOo.CoM</a></p>",
+    "example": 16
+  },
+  {
+    "section": "Autolinks",
     "markdown": "~~hello@email.com~~",
     "html": "<p><del><a href=\"mailto:hello@email.com\">hello@email.com</a></del></p>",
     "example": 1307
--- a/test/specs/marked/marked.json
+++ b/test/specs/marked/marked.json
@@ -93,7 +93,7 @@
     "section": "Autolinks",
     "markdown": "hTtP://fOo.CoM",
     "html": "<p><a href=\"hTtP://fOo.CoM\">hTtP://fOo.CoM</a></p>",
-    "example": 16
+    "example": 18
   },
   {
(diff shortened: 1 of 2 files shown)


test/specs/marked/marked.json
```diff
@@ -83,6 +83,18 @@
     "html": "<p>``<em>test`</em></p>",
     "example": 14
   },
+  {
+    "section": "Autolinks",
+    "markdown": "HTTP://FOO.COM",
+    "html": "<p><a href=\"HTTP://FOO.COM\">HTTP://FOO.COM</a></p>",
+    "example": 15
```
styfle:
```suggestion
    "example": 17
```

test/specs/marked/marked.json
```diff
@@ -83,6 +83,18 @@
...
+  },
+  {
+    "section": "Autolinks",
+    "markdown": "hTtP://fOo.CoM",
+    "html": "<p><a href=\"hTtP://fOo.CoM\">hTtP://fOo.CoM</a></p>",
+    "example": 16
```
styfle:
```suggestion
    "example": 18
```

test/specs/marked/marked.json
```diff
@@ -83,6 +83,18 @@
     "html": "<p>``<em>test`</em></p>",
     "example": 14
   },
+  {
+    "section": "Autolinks",
```
styfle: Is it a problem if the example numbers collide?

We just had example 15 added in master: https://github.com/markedjs/marked/blob/master/test/specs/marked/marked.json

styfle:
/cc @roygold7 since PR 1350 was started but didn't finish.

I want to make sure Roy is aware of this PR and that the reason for the new PR is no movement on the old PR.

I don't want to end up in [this scenario](https://twitter.com/timocov_/status/1065501522073665536).

test/specs/marked/marked.json
UziTech: I think the example numbers are just used in the shouldPassButFail array and the test descriptions.

The `marked` tests shouldn't ever fail so maybe we should replace `example` with `description` so they won't collide as often

roygold7:
@styfle I appreciate the comment but I'd be just as happy if the issue was resolved in a different PR. I'm not looking for credit.

test/specs/marked/marked.json
styfle: > maybe we should replace example with description so they won't collide as often

I'm fine with that. Let's save that change for a different PR though. This looks good now.

davisjam:
LGTM. This should not have a ReDoS impact.

Issue #1216, opened 2018-04-13
Defibrillator challenge


joshbruce:
**Marked version:** 0.3.19

**Markdown flavor:** CommonMark & GitHub Flavored Markdown

**Proposal type:** other

## What pain point are you perceiving?

See #1106, #1202, and #1211.

## What solution are you suggesting?

Create a badge for contributors to earn. The challenge is, hopefully, simple enough.

1. View the testbeds for [CommonMark](https://github.com/markedjs/marked/blob/master/test/specs/commonmark/commonmark-spec.js) and [GFM](https://github.com/markedjs/marked/blob/master/test/specs/gfm/gfm-spec.js).
2. Within each testbed are test suites for each section of the related [CommonMark](http://spec.commonmark.org/0.28/) and [GFM](https://github.github.com/gfm/) specifications, respectively. 
3. Within each test suite is a variable called `shouldPassButFails` with an array of example numbers that fail even though they should pass.
4. Pick at least one example to remove from the array.
5. Run `npm run test:cm|gfm` to confirm it still fails.
6. Write the code necessary to make it pass.
7. Submit the pull request.
8. Get added to the [AUTHORS](https://github.com/markedjs/marked/blob/master/docs/AUTHORS.md) page as part of the defibrillator gang.

PR #1398, opened 2018-12-20
Fix no-break space on lists

mccraveiro:
**Marked version:** master branch

## Description

**Important: This PR introduces parsing of invisible unicode characters. Please double check it!**

- Related to #1216 

## Input

```markdown
* a *
```

**Important: note that ` `  is not a normal space. It is a "no-break unicode space"**

## Expectation

```html
<p>* a *</p>
```

## Result

```html
<ul>
<li>a *</li>
</ul>
```

## What was attempted

Stops replacing "no-break" spaces and fixes the example with the correct unicode char

## Contributor

**Important: I had to change 2 old tests so they match the spec. Please double check it! :)**

## Committer

- [ ] Merge PR


UziTech:
I think we want to keep replacing `\u00a0` with a regular space by default. see #52 and #897

I think the conclusion to #897 was to create an option to keep `\u00a0`

mccraveiro:
@UziTech that would mean marked would never be 100% CommonMark compliant. Are we fine with that?

UziTech:
I'm ok with it not being 100% compliant by default as long as there are options that will make it 100% compliant.

UziTech:
I can't find the discussion but I think the original idea was to limit the number of breaking changes and when we get to v1.0 deprecate a bunch of things and remove things in v2.0

After looking at https://libraries.io/npm/marked/usage it looks like most dependents are pinned to pre v0.4 anyways so breaking changes shouldn't be as big of an issue as previously suspected.

PR #1305, opened 2018-07-03
enable CommonMark spec 468

Trott:
**Marked version:** 0.4.0

**Markdown flavor:** CommonMark

## Description

A minor adjustment to the href-matching regexp permits CommonMark spec
468 to pass. All other passing specs are unaffected.

The regexp was incorrectly excluding parentheses as valid characters in
a URL.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR); or,
- [ ] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] Draft GitHub release notes have been updated.
- [ ] CI is green (no forced merge required).
- [ ] Merge PR

PR #1408, opened 2019-01-13
Fix ReDOS #1405

### LINKED ISSUE #1405: Too many a tag, the CPU will be 100% when there is no split
**Describe the bug**
Too many a tag, the CPU will be 100% when there is no split.

**To Reproduce**

md is:
```
* 伪类：[:active](https://developer.mozilla.org/en-US/docs/Web/CSS/:active)、[:any-link](https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link)、[:blank](https://developer.mozilla.org/en-US/docs/Web/CSS/:blank)、[:checked](https://developer.mozilla.org/en-US/docs/Web/CSS/:checked)、[:current](https://developer.mozilla.org/en-US/docs/Web/CSS/:current)、[:default](https://developer.mozilla.org/en-US/docs/Web/CSS/:default)、[:defined](https://developer.mozilla.org/en-US/docs/Web/CSS/:defined)、[:dir()](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir)、[:disabled](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled)、[:drop](https://developer.mozilla.org/en-US/docs/Web/CSS/:drop)、[:empty](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty)、[:enabled](https://developer.mozilla.org/en-US/docs/Web/CSS/:enabled)、[:first](https://developer.mozilla.org/en-US/docs/Web/CSS/:first)、[:first-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child)、[:first-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type)、[:fullscreen](https://developer.mozilla.org/en-US/docs/Web/CSS/:fullscreen)、[:future](https://developer.mozilla.org/en-US/docs/Web/CSS/:future)、[:focus](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus)、[:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible)、[:focus-within](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within)、[:has()](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)、[:host](https://developer.mozilla.org/en-US/docs/Web/CSS/:host)、[:host()](https://developer.mozilla.org/en-US/docs/Web/CSS/:host())、[:host-context()](https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context())、[:hover](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover)、[:indeterminate](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate)、[:in-range](https://developer.mozilla.org/en-US/docs/Web/CSS/:in-range)、[:invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid)、[:is()](https://developer.mozilla.org/en-US/docs/Web/CSS/:is)、[:lang()](https://developer.mozilla.org/en-US/docs/Web/CSS/:lang)、[:last-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child)、[:last-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type)、[:left](https://developer.mozilla.org/en-US/docs/Web/CSS/:left)、[:link](https://developer.mozilla.org/en-US/docs/Web/CSS/:link)、[:local-link](https://developer.mozilla.org/en-US/docs/Web/CSS/:local-link)、[:not()](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)、[:nth-child()](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)、[:nth-col()](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-col)、[:nth-last-child()](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-child)、[:nth-last-col()](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-col)、[:nth-last-of-type()](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-of-type)、[:nth-of-type()](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type)、[:only-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child)、[:only-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-of-type)、[:optional](https://developer.mozilla.org/en-US/docs/Web/CSS/:optional)、[:out-of-range](https://developer.mozilla.org/en-US/docs/Web/CSS/:out-of-range)、[:past](https://developer.mozilla.org/en-US/docs/Web/CSS/:past)、[:placeholder-shown](https://developer.mozilla.org/en-US/docs/Web/CSS/:placeholder-shown)、[:read-only](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only)、[:read-write](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-write)、[:required](https://developer.mozilla.org/en-US/docs/Web/CSS/:required)、[:right](https://developer.mozilla.org/en-US/docs/Web/CSS/:right)、[:root](https://developer.mozilla.org/en-US/docs/Web/CSS/:root)、[:scope](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope)、[:target](https://developer.mozilla.org/en-US/docs/Web/CSS/:target)、[:target-within](https://developer.mozilla.org/en-US/docs/Web/CSS/:target-within)、[:user-invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:user-invalid)、[:valid](https://developer.mozilla.org/en-US/docs/Web/CSS/:valid)、[:visited](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited)、[:where()](https://developer.mozilla.org/en-US/docs/Web/CSS/:where)
* 伪元素：[::after (:after)](https://developer.mozilla.org/en-US/docs/Web/CSS/::after)、[::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop)、[::before (:before)](https://developer.mozilla.org/en-US/docs/Web/CSS/::before)、[::cue (:cue)](https://developer.mozilla.org/en-US/docs/Web/CSS/::cue)、[::first-letter (:first-letter)](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter)、[::first-line (:first-line)](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-line)、[::grammar-error](https://developer.mozilla.org/en-US/docs/Web/CSS/::grammar-error)、[::marker](https://developer.mozilla.org/en-US/docs/Web/CSS/::marker)、[::placeholder](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder)、[::selection](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection)、[::slotted()](https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted)、[::spelling-error](https://developer.mozilla.org/en-US/docs/Web/CSS/::spelling-error) 
```

**Expected behavior**
CPU is normal

Feder1co5oave:
Fixes #1405, reverts #1305
Something went wrong with the security review of that PR :(

I don't think we'll ever support multiple nested _anything_, _anywhere_. Regexes are just not made to do that easily.

I'd consider releasing this as a patch release to all minor versions available on npm.


Trott:
> I would hate to revert because I believe that @Trott fixed that specific test case because they were running into a problem while rendering the Node.js docs with marked.

If it helps at all: We're not using `marked` in Node.js anymore, so reverting it won't hurt Node.js, at least.

UziTech:
@davisjam on https://github.com/markedjs/marked/pull/1305#issuecomment-403339860:
> I will cryptically remark that its vulnerability status is unchanged.

I don't think this regex for link is any more secure then the one in #1305. It just happens to fix this redos attack.

I think this is a situation where we need to simplify the regex and use some other method to get the parts of the link, similar to what we do with tables.

Feder1co5oave:
@UziTech sorry I totally forgot about this issue.
I can't clearly explain why, but I'm pretty confident this regex is secure, while the one currently in master is not.
It has to do with the fact that, when trying to match `/\([^\s\x00-\x1f\\()]*\)/`, the engine requires an opening `(`, something that _certainly is not_ a closing `)`, and then the closing `)`. Simple as that.
Whereas the offending `/\([^\s\x00-\x1f\\]*\)/` is dangerous because to match it you need to:
- find the `(`
- match any characters in `[^\s\x00-\x1f\\(]`
  if a `)` is encountered, we cannot know whether it is the ending `)`, or one in the middle. We must try **both**.
  - try matching it and as belonging to `[^\s\x00-\x1f\\()]`, and keep going.
    - If `$` is encountered, it might be because the last `)` was in fact the ending one. Backtrack.
    - Anything in `[^\s\x00-\x1f\\(]` is found. Eat it.
    - Another `(` is encountered. We cannot know whether.... branch once again, and then backtrack if applicable.

For every `)` encountered in the input, exponential backtracking takes place.

UziTech:
@Feder1co5oave yes that could be why it fixes this situation, but I was saying the regex in general is no more secure. In other words we will have to change the regex in a different way to make it secure in the future.

atodorov:
How about a quick fix with this patch and then redo the regex into a function ?

UziTech:
I'd like to hear from @davisjam on this

Pablodotnet:
We need this to be merged to fix the vulnerability snyk is advertising https://snyk.io/vuln/SNYK-JS-MARKED-73637

UziTech:
#1414 also fixes the vulnerability and still allows urls with matching parentheses

PR #1409, opened 2019-01-13
Update devDependencies to last stable

Feder1co5oave:
Update devDependencies to versions found at https://david-dm.org/markedjs/marked?type=dev

Had to disable new eslint rule to avoid introducing codebase changes. We have mixed type of literal objects:
```javascript
{key: "value"}
{ key: "value" }
```

so we might go for one or the other, or leave them as they are.
If we consider linting more aggressively, also the rule `space-before-function-paren` can be enforced, and we might take the chance and change indentation warnings into errors so they're caught by the CI robot. Let me know.


styfle:
I personally prefer spaces in object literals such as `{ key: "value" }` and the same for destructuring such as `const { key } = obj`.

Martii:
@Feder1co5oave 

Apologies for the interruption...

> ... last stable


Do you mean "latest"? Last implies that something is EOL'd from a quick read in my emails/notifications.

Feder1co5oave:
@Martii all the versions are described as "stable" at the link above. I think in semver _everything_ is considered "stable" unless a postfix such as "-alpha" or "-rc" is appended.

Feder1co5oave:
`glob-to-regex@0.3.0` was introduced in #1004 but I can't find the specific reason it was pinned

UziTech:
I probably just forgot the caret (`^`) It doesn't look like v0.4.0 should break anything.

PR #1428, opened 2019-02-23
[WIP] Worker thread

UziTech:
**Marked version:** master

## Description

Use [worker threads](https://nodejs.org/api/worker_threads.html) on original tests.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -94,8 +94,10 @@ function runTests(engine, options) {
   if (!engine) {
     try {
       engine = require('./worker.js');
+      engine.newWorker();
       usingWorker = true;
     } catch (ex) {
+      console.log(ex);
       engine = marked;
     }
   }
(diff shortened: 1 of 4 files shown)


UziTech:
Just testing out node worker threads for the "new" and "original" tests.

This will allow us to run redos tests easier.

Worker threads are experimental in node 11 so it falls back to running tests on the main thread when worker threads are not available.

I set them to fail after 2 seconds.

I also changed the tests to use Promises but everything should work in node v4 and above.

UziTech:
worker_threads in node are [changing](https://github.com/nodejs/node/releases/tag/v11.11.0)

We should probably wait on this until after they are not experimental.

I don't think there is a current way to move a PR back to draft.

UziTech:
This will be obsolete with #1482

PR #1438, opened 2019-03-07
allow html without \n after

UziTech:
**Marked version:** master

## Description

Fix html without a new line after

```
marked('<img src="sdfg">\n');
// expected: '<img src="sdfg">'
// actual:   '<img src="sdfg">'

marked('<img src="sdfg">');
// expected: '<img src="sdfg">'
// actual:   '<p><img src="sdfg"></p>'
```

- Fixes https://github.com/markedjs/marked/issues/1232#issuecomment-470663325

## Contributor

## Committer

- [x] Merge PR


styfle:
There are some merge conflicts

lib/marked.js
```diff
@@ -27,8 +27,8 @@ var block = {
     + '|<![A-Z][\\s\\S]*?>\\n*' // (4)
     + '|<!\\[CDATA\\[[\\s\\S]*?\\]\\]>\\n*' // (5)
     + '|</?(tag)(?: +|\\n|/?>)[\\s\\S]*?(?:\\n{2,}|$)' // (6)
-    + '|<(?!script|pre|style)([a-z][\\w-]*)(?:attribute)*? */?>(?=\\h*\\n)[\\s\\S]*?(?:\\n{2,}|$)' // (7) open tag
-    + '|</(?!script|pre|style)[a-z][\\w-]*\\s*>(?=\\h*\\n)[\\s\\S]*?(?:\\n{2,}|$)' // (7) closing tag
```
UziTech: @Feder1co5oave I'm not sure what the `\\h*` is supposed to do. As far as I know `\H` isn't a special character.

davisjam: `\h` has meaning in some languages.

| Language | Meaning of `\h` |
| ------------- | --------------- |
| Ruby | Hex chars, e.g. `[0-9a-fA-F]` |
| Java, PHP, Perl | Horizontal whitespace, cf. `\v` |

davisjam: ...so I imagine this regex originated in a language (or in someone's head) where `\h` had meaning.

davisjam: ...I've written a research paper on this topic if @UziTech @styfle @Feder1co5oave would be interested.

lib/marked.js
```diff
@@ -27,8 +27,8 @@ var block = {
...
+    + '|<(?!script|pre|style)([a-z][\\w-]*)(?:attribute)*? */?>(?=\\n|$)[\\s\\S]*?(?:\\n{2,}|$)' // (7) open tag
```
davisjam: - Is there a test case that includes horizontal whitespace where the `\h` was?
- It's possible the original regex should have used something like `[ \t]` instead?

lib/marked.js
davisjam: - Is there a test case that includes horizontal whitespace where the \h was?
- It's possible the original regex should have used something like [ \t] instead?

UziTech: It looks like it should be `[ \\t]*` but I would like to get @Feder1co5oave point of viewq since he is the one that introduced it.

Feder1co5oave: It was intended to be `[ \t]*` I guess, I just don't remember why I used `\h` instead, and why it required a newline after the tag :(

UziTech: I added `[ \t]*` in place of `\h` but with some manual testing it doesn't seem to change anything. All of the tests I can think of still pass with or without `[ \t]*`

UziTech: That could be due to the fact that we strip white space before we check the output.

UziTech:
@davisjam @joshbruce This should be ready to merge. Once this is merged we can release [v0.6.2](https://github.com/markedjs/marked/pull/1441)

lib/marked.js
Feder1co5oave: We still strip whitespace when testing? Wasn't html-differ supposed to fix this?

UziTech: We were only using html-differ for the spec tests. The files were still using the manual method.

I fixed it in #1449

UziTech:
@davisjam @joshbruce We need one of you to approve this in order to merge it and release a new version of marked.

joshbruce:
Sorry. I'm gonna be a little slow on things for at least the next few days (family-related keeping me offline for long stretches). Maybe we should consider an alternative to keep from deadlock. Though not sure it will happen often.

Issue #1430, opened 2019-02-28
Task List Classes


oasiz:
**Describe the feature**
Hi folks! This Markdown:

```
- [ ] foo
- [x] bar
```

.. generates the following HTML:

```
<ul>
    <li><input type="checkbox" disabled=""> foo</li>
    <li><input type="checkbox" disabled="" checked=""> bar</li>
</ul>
```

Is it possible to add classes (as GitHub does) to make these style-able?

GitHub renders as such:

```
<ul class="contains-task-list">
    <li class="task-list-item"><input type="checkbox" id="" disabled="" class="task-list-item-checkbox"> foo</li>
    <li class="task-list-item"><input type="checkbox" id="" disabled="" class="task-list-item-checkbox" checked=""> bar</li>
</ul> 
```

**Why is this feature necessary?**
I want to add "list-style: none;" to the UL class, as it currently doesn't look good:

![tasklist](https://user-images.githubusercontent.com/2811813/53571933-ffbdb500-3b61-11e9-8b71-3f262bcdf5f6.png)

Apologies if I've missed a way to do this, I've looked through the docs but couldn't see a way.

Thanks for reading! :-)

UziTech:
You should be able to do this be extending the [renderer](https://marked.js.org/#/USING_PRO.md#renderer)

something like:

```js
// Create reference instance
const marked= require('marked');

// Get reference
const renderer = new marked.Renderer();
const renderListitem = renderer.prototype.listitem;

// Override function
renderer.listitem = function (text) {
	let html = renderListitem(text);
	
	if (html.includes("<input")) {
		html = html.replace("<li>", "<li class='task-list-item'>");
	}

	return html;
};

// Run marked
console.log(marked('* [ ] test', { renderer: renderer }));

// <ul><li class='task-list-item'><input disabled="" type="checkbox"> test</li></ul>
```

UziTech:
It could be easier if the renderer was passed a `task` boolean.

That would be a great PR if you want to work on it. 😁

oasiz:
> You should be able to do this be extending the [renderer](https://marked.js.org/#/USING_PRO.md#renderer)
> 
> something like:
> 
> ```js
> // Create reference instance
> const marked= require('marked');
> 
> // Get reference
> const renderer = new marked.Renderer();
> const renderListitem = renderer.prototype.listitem;
> 
> // Override function
> renderer.listitem = function (text) {
> 	let html = renderListitem(text);
> 	
[...]

PR #1440, opened 2019-03-08
Pass token boolean to the listitem function.

x13machine:
**Marked version:**

## Description

- Passes token boolean to the listitem function. In response to this issue: https://github.com/markedjs/marked/issues/1430
- Also changes "avaialable" to "available" in CONTRIBUTING.md

## Contributor

## Committer

- [ ] Merge PR


lib/marked.js
```diff
@@ -1258,18 +1258,19 @@ Parser.prototype.tok = function() {
...
       }
 
       while (this.next().type !== 'list_item_end') {
         body += !loose && this.token.type === 'text'
           ? this.parseText()
           : this.tok();
       }
-
-      return this.renderer.listitem(body);
+      return this.renderer.listitem(body, task);
```
UziTech: Should we also pass checked?

```suggestion
      return this.renderer.listitem(body, task, checked);
```

x13machine:
Do you think passing loose would be a good idea too?

UziTech:
> Do you think passing loose would be a good idea too?

That is really only useful for telling marked how to parse text in the list item.

styfle:
@x13machine Thanks for the PR!

Can you update the docs in `USING_PRO.md` so that listitem api matches the code?

Making a note for myself: this will be a breaking change.

UziTech:
I don't think this will be a breaking change since we are just adding parameters not changing any current parameters

UziTech:
I would say this is a minor version change, which in a beta version is the same as a patch

styfle:
Oh right, we aren't changing the `Renderer.prototype.listitem` to depend on the new args. So I guess that's not a breaking change 🤔

Issue #1423, opened 2019-02-19
Add --test option in man page


roipoussiere:
**Describe the bug**
The `--test` CLI option is not displayed in the man page.

**To Reproduce**

```
man marked
```

This appends in last version and simple to verify by looking at the [man](https://github.com/markedjs/marked/blob/master/man/marked.1) [files](https://github.com/markedjs/marked/blob/master/man/marked.1.txt).

**Expected behavior**

The `--test` CLI option should be displayed in the man page, since it is [a valid option in the marked binary](https://github.com/markedjs/marked/blob/master/bin/marked#L80).

x13machine:
I think, this should be closed

styfle:
Fixed in #1442

PR #1442, opened 2019-03-10
Add --test man documentation and changes argv parameters test/index.js

x13machine:
**Marked version:**

## Description

- Added --test man documentation in response to https://github.com/markedjs/marked/issues/1423. I don't know If I'm using the correct groff format. I ran "man -P cat ./marked.1 > marked.1.txt" to write the "marked.1.txt" file. I don't know if that's what I'm suppose to do. 
- Changed the parseArg code in test/index.js because it was using process.argv and not argv. https://github.com/markedjs/marked/blob/d96d1d37cdd94e1e6fa100cea9a2b8ccb202b0ca/test/index.js#L554

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/bin/marked
+++ b/bin/marked
@@ -78,7 +78,7 @@ function main(argv, callback) {
     arg = getarg();
     switch (arg) {
       case '--test':
-        return require('../test').main(process.argv.slice());
+        return require('../test').main(process.argv);
       case '-o':
       case '--output':
         output = argv.shift();
--- a/bin/marked
+++ b/bin/marked
@@ -78,7 +78,7 @@ function main(argv, callback) {
     arg = getarg();
     switch (arg) {
       case '--test':
-        return require('../test').main(process.argv);
+        return require('../test').main(process.argv.slice());
       case '-o':
       case '--output':
         output = argv.shift();
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1270,7 +1270,7 @@ Parser.prototype.tok = function() {
           ? this.parseText()
           : this.tok();
       }
-      return this.renderer.listitem(body, task, checked);
+      return this.renderer.listitem(body, task, checked, loose);
     }
     case 'html': {
       // TODO parse inline content if parameter markdown=1
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1270,7 +1270,7 @@ Parser.prototype.tok = function() {
           ? this.parseText()
           : this.tok();
       }
-      return this.renderer.listitem(body, task, checked, loose);
+      return this.renderer.listitem(body, task, checked);
     }
     case 'html': {
       // TODO parse inline content if parameter markdown=1
--- a/man/marked.1.txt
+++ b/man/marked.1.txt
@@ -26,13 +26,13 @@ OPTIONS
        -i, --input [input]
               Specify file input, otherwise use last argument as input file. If no input file is specified, read from stdin.
 
-       --test Makes sure test file(s) pass.
+       --test Makes sure the test(s) pass.
 
-              --glob [file] Specify test file to use.
+              --glob [file] Specify which test to use.
 
-              --bench Benchmarks the test file(s).
+              --bench Benchmarks the test(s).
 
-              --time Times The test file(s).
+              --time Times The test(s).
 
               --Run Runs test file(s) as minified.
 
--- a/man/marked.1.txt
+++ b/man/marked.1.txt
@@ -30,11 +30,13 @@ OPTIONS
 
               --glob [file] Specify which test to use.
 
+              --fix Fixes tests.
+
               --bench Benchmarks the test(s).
 
               --time Times The test(s).
 
-              --Run Runs test file(s) as minified.
+              --minified Runs test file(s) as minified.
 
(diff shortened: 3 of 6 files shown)


lib/marked.js
```diff
@@ -1270,7 +1270,7 @@ Parser.prototype.tok = function() {
           ? this.parseText()
           : this.tok();
       }
-      return this.renderer.listitem(body, task, checked);
+      return this.renderer.listitem(body, task, checked, loose);
```
styfle: I don't think adding `loose` should be part of this PR.

man/marked.1.txt
```diff
@@ -26,71 +19,72 @@ EXAMPLES
...
+
+       --test Makes sure the test(s) pass.
+
+              --glob [file] Specify which test to use.
+
+              --bench Benchmarks the test(s).
+
+              --time Times The test(s).
+
+              --Run Runs test file(s) as minified.
```
styfle: I don't see the implementation of `--Run`.
Where did you find that?

x13machine:
I meant --minified and not --run. The loose parameter was a mistake.

test/index.js
```diff
@@ -441,9 +441,8 @@ function fix() {
  * Argument Parsing
  */
 
-function parseArg() {
-  var argv = process.argv.slice(2),
```
styfle: This looks like a bug to me. Did you intend to remove the `slice(2)` here?

x13machine: It seemed it was meant to get the argv from here. https://github.com/markedjs/marked/blob/d96d1d37cdd94e1e6fa100cea9a2b8ccb202b0ca/bin/marked#L81 because main(argv) didn't use the argv parameter. https://github.com/markedjs/marked/blob/d96d1d37cdd94e1e6fa100cea9a2b8ccb202b0ca/test/index.js#L553 It would just make the code nicer, wouldn't change functionality. I tested it and nothing changed.

styfle: This code was added in 62467849142f92d85f2a7b7f1005fd5542fba4ae and it seems like `slice(2)` was intended because the first two arguments are typically `["node", "test.js"]`

/cc @UziTech

x13machine: It doesn't seem like it matters because "marked sdfg --test" works

UziTech: To prevent a breaking change we should slice the first two args off of `argv`.

```js
function parseArg(argv) {
  argv = argv.slice(2);
```

bin/marked
```diff
@@ -78,7 +78,7 @@ function main(argv, callback) {
     arg = getarg();
     switch (arg) {
       case '--test':
-        return require('../test').main(process.argv.slice());
+        return require('../test').main(process.argv);
```
UziTech: `.slice()` is supposed to be there to pass a copy of the array to prevent unwanted changes directly to `process.argv`

test/index.js
```diff
@@ -582,7 +582,7 @@ function main(argv) {
 
 if (!module.parent) {
   process.title = 'marked';
-  process.exit(main(process.argv.slice()) ? 0 : 1);
+  process.exit(main(process.argv) ? 0 : 1);
```
UziTech: We should keep `.slice()` to keep `process.argv` immutable

man/marked.1.txt
```diff
@@ -26,71 +19,76 @@ EXAMPLES
...
        Copyright (c) 2011-2014, Christopher Jeffrey (MIT License).
 
-
 SEE ALSO
        markdown(1), node.js(1)
 
-
-
-v0.3.1				  2014-01-31			     marked(1)
+v0.3.1                                                             2014-01-31                                                          marked(1)
```
UziTech: @styfle Should this version and date be updated any time we release a new version?

UziTech:
To make the tests pass you will have to change `/test/specs/original/specs-spec.js` line 6

```diff
-  if (!specTests({stop: true})) {
+  if (!specTests(['', '', '--stop'])) {
```

test/index.js
```diff
@@ -441,13 +441,13 @@ function fix() {
...
-function parseArg() {
-  var argv = process.argv.slice(2),
-      options = {},
+function parseArg(argv) {
+  var options = {},
       opt = '',
       orphans = [],
       arg;
 
+  argv = Array.isArray(argv) ? argv.slice(2) : [];
```
UziTech: I'm not sure this check should be there. If `argv` is not an array it should throw an error.

test/index.js
```diff
@@ -551,7 +551,7 @@ function camelize(text) {
  */
 
 function main(argv) {
-  var opt = parseArg();
+  var opt = parseArg(argv);
```
UziTech: You could do the check here. If `argv` is undefined `opt` should be set to `{}`

x13machine: I just added it to get the test error to go away and didn't remove it after I added your solution.

man/marked.1.txt
UziTech: and maybe the line shouldn't be so long?

styfle: Maybe we should remove the version from the man page and instead provide a `--version` flag.

PR #1449, opened 2019-03-13
Update tests to node 4 syntax

UziTech:
**Marked version:** master

## Description

Update all tests to use html-differ and use syntax for node 4

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/test/helpers/helpers.js
+++ b/test/helpers/helpers.js
@@ -1,8 +1,5 @@
 const marked = require('../../');
-const HtmlDiffer = require('@markedjs/html-differ').HtmlDiffer;
-const htmlDiffer = new HtmlDiffer({ignoreSelfClosingSlash: true});
-
-const EXCERPT_LENGTH = 30;
+const htmlDiffer = require('./html-differ.js');
 
 beforeEach(() => {
   marked.setOptions(marked.getDefaults());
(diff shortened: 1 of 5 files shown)


test/helpers/helpers.js
```diff
@@ -18,23 +15,8 @@ beforeEach(() => {
...
-
-                expectedHtml = expectedHtml.substring(
-                  Math.max(i - EXCERPT_LENGTH, 0),
-                  Math.min(i + EXCERPT_LENGTH, expectedHtml.length));
-
-                break;
-              }
-            }
-            result.message = 'Expected:\n' + expectedHtml + '\n\nActual:\n' + actualHtml;
+            const diff = htmlDiffer.firstDiff(actual, expected);
```
styfle:
```suggestion
            const { suggested, actual } = htmlDiffer.firstDiff(actual, expected);
```

test/helpers/helpers.js
```diff
@@ -18,23 +15,8 @@ beforeEach(() => {
...
+            result.message = 'Expected: ' + diff.expected + '\n  Actual: ' + diff.actual;
```
styfle:
```suggestion
            result.message = `Expected: ${expected}\nActual: ${actual}`;
```

test/index.js
```diff
@@ -289,13 +245,13 @@ function runBench(options) {
   }
   bench('marked (pedantic)', files, marked);
 
-  // showdown
+  // commonmark
```
styfle: Let's remove these comments since they are redundant

test/helpers/helpers.js
UziTech: destructuring is not available in node 4

test/index.js
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+'use strict';
+// 'use strict' is here so we can use let and const in node 4
```
davisjam: Delete this line?

UziTech: 'use strict' is required for block scope let and const in node v4. I figured I would write a comment so people don't remove it since it isn't needed in newer versions

styfle: It throws an error when you remove it, right? So if someone removes it the tests should fail (I assume).

UziTech: It throws an error that might not be obvious that they have to put `'use script';` back.

I also like the comment because it lets us know that `'use script';` can be removed once we no longer support testing on node 4 .

styfle: Sounds good to me 👍

PR #1454, opened 2019-03-22
NFU - Add option to add class to pre tag

funayman:
## Description

Adds an option (`preClass`) to add (for example) `class="language-go` to the `<pre>` element in a code block.

Example of using `preClass`:
```js
marked.setOptions({ sanitize: true, preClass: true, langPrefix: "language-", highlight: function(code, lang) {
  var Prism = require('prismjs');
  var loadLanguages = require('prismjs/components/');
  loadLanguages([lang]);
  return Prism.highlight(code, Prism.languages[lang], lang);
}});
```

## Expectation
```html
<pre class="language-go">
  <code class="language-go">fmt.Println("hello world")</code>
</pre>
```

## Result
```html
<pre>
  <code class="language-go">fmt.Println("hello world")</code>
</pre>
```
## What was attempted
Tried making a custom highlight function, but the `<pre>` is hard coded in.

## Contributor

## Committer

- [ ] Merge PR


UziTech:
This sounds very niche. We are trying to deprecate options and move them to extensions.

Out of curiosity why do you need the class on the pre?

UziTech:
You could do this pretty easily as a string replace after passing through marked.

Something like:

```js
const html = marked(markdownString)
const preClass = html.replace(/<pre>(\s*)<code class="([\w-]+)">/g, '<pre class="$2">$1<code class="$2">');
```

funayman:
Fair enough. I'm working on incorporating [schnack](https://schnack.cool/) into my blog, and I'm using prismjs for syntax highlighting. Without the class being added to the `<pre>` tag, code blocks are treated as inline code.

Thanks for your help! I'll give your code a go!

Issue #1456, opened 2019-03-28
Parsing ![]() is buggy


jackens:
**Describe the bug**
Parsing `![]()`  is buggy

**To Reproduce**

```md
![](img1.svg) (or ![](img2.svg))
```

generates

```html
<p><img src="img1.svg" alt="" title="or ![](img2.svg">)</p>
```

(conf. <https://marked.js.org/demo/?text=![](img1.svg)%20(or%20![](img2.svg))>)

**Expected behavior**
```html
<p><img src="img1.svg" alt=""> (or <img src="img2.svg" alt="">)</p>
```

UziTech:
This should be fixed in #1453

[demo](https://marked.js.org/demo/?outputType=html&text=!%5B%5D(https%3A%2F%2Fexample.com%2F1)%20(!%5B%5D(https%3A%2F%2Fexample.com%2F2))%20!%5B%5D(https%3A%2F%2Fexample.com%2F3)&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tables%22%3A%20true%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=pr%3A1435)

UziTech:
I'm not sure why the demo's master isn't working since #1435 was already merged, but this is already fixed in master.

UziTech:
demo master should be fixed in #1457

UziTech:
[...]

PR #1457, opened 2019-03-29
use latest commit for demo master

UziTech:
## Description

use `@master` in demo for master version

https://docs-ikjy0vxvf.now.sh/demo/

- fixes comment https://github.com/markedjs/marked/issues/1456#issuecomment-478090870

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/docs/demo/demo.js
+++ b/docs/demo/demo.js
@@ -34,7 +34,7 @@ var inputDirty = true;
 var $activeOutputElem = null;
 var search = searchToObject();
 var markedVersions = {
-  master: 'https://cdn.jsdelivr.net/gh/markedjs/marked@master/lib/marked.js'
+  master: 'https://cdn.jsdelivr.net/gh/markedjs/marked/lib/marked.js'
 };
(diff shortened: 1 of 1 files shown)


styfle:
I think the branch is cached by the CDN for a week or so which is maybe why it wasn’t showing the latest.

UziTech:
If that is the case than this won't fix the issue. I wonder If I should grab the latest commit hash and use that.

styfle:
@UziTech That might work. Another alternative is to add a "bust cache" button since there appears to be an [API](https://github.com/jsdelivr/jsdelivr#purge-cache) for that (requires an email).

See the full discussion here: https://github.com/jsdelivr/jsdelivr/issues/18101

UziTech:
Using the latest commit hash seems to be the right way to go.

https://docs-qaob045da.now.sh/demo/

UziTech:
I also fixed a bug when using a regular version

https://docs-ikjy0vxvf.now.sh/demo

PR #1472, opened 2019-04-14
Update to resolve js-yaml vulnerability

calvinchengx:


UziTech:
This is a dev dependency so it shouldn't affect any of our users. We should wait for an eslint update.

calvinchengx:
@UziTech but `npm audit` doesn't really care whether it is a dev dependency or not ?  It still spits out the warning.  Will an eslint update resolve this?

UziTech:
> Will an eslint update resolve this?

Yes, js-yaml is a dependency of eslint and front-matter. Once they release an update that updates js-yaml we will update those dependencies.

calvinchengx:
I see. Thanks for clarification.

styfle:
@calvinchengx You are likely cloning this repo and running `npm install` or `npm audit`.

For users who are using `marked`, they do not see (and are not affected) by dev dependency vulnerabilities.

### Example

```
mkdir example
cd example
npm init -y
npm install marked
```

### Output

```
+ marked@0.6.2
added 1 package from 1 contributor and audited 1 package in 1.081s
found 0 vulnerabilities
```

The caveat here is that a dev dependency vulnerability in something that emits code that gets published (for example uglify) could affect our end users. But eslint is nothing to worry about.

PR #1482, opened 2019-04-25
Clean up tests

UziTech:
**Marked version:** master

## Description

- Move all tests to jasmine (still `.md` and `.html` files for original and new)
- Remove old `gfm_` and `cm_` tests from new
- Bench uses CommonMark specs
- Create `npm run test:update` to update CommonMark and GFM specs
- Remove `standard` from eslint config for marked.js to disallow const and template strings for older versions of node and browsers
- Remove browser test that wasn't working anyways

## Contributor

## Committer

- [ ] Merge PR


bin/marked
```diff
@@ -13,7 +13,7 @@ var fs = require('fs'),
  * Man Page
  */
 
-function help() {
+function help () {
```
styfle: But why spaces?

UziTech: I figured it should be consistant and since we are using standard might as well use their recommendation.

I personally don't care one way or the other but they were inconsistent before.

styfle: I think it should have the fewest changes possible (preferably none) to `lib/marked.js` and `bin/marked`

UziTech: As for the philosophical question "why?" I think the argument is that is makes visual grepping easier.

`help()` = calling a help function
`help ()` = defining a help function

UziTech: I think I did pick the least changes. Making it consistent would have changed it either way.

joshbruce: My brain takes issue with this. Which standard/style/recommendation is this from (sorry, might have missed a mention somewhere)?

I haven't seen a language or style that adds a space between the name and the argument list container. Mainly just curious because it seems like it might add confusion not reduce it.

styfle: > I think I did pick the least changes

This is changing every single function. This is not necessary. We should make the tools follow the standards of the project, not conform to someone else's standard (unless there is clear value added).

UziTech: [standard js](https://standardjs.com/rules.html#javascript-standard-style) (which we [have been using](https://github.com/markedjs/marked/commit/2fa274284ea7e237608b947f3be03918eb31b5f1#diff-df39304d828831c44a2b9f38cd45289c) with eslint) has [space-before-function-paren](https://eslint.org/docs/rules/space-before-function-paren) eslint rule set to "always". We had that rule turned off which made our code inconsistant.

This PR removes `standard` from eslint since it doesn't allow us to catch es6 syntax in eslint but I added all of standard's rules to keep it consistant.

I changed it to never allow a space between a function and parens

UziTech:
https://github.com/eslint/eslint/issues/11658 is the reason I removed `"extends": "standard"` from eslint

UziTech:
@joshbruce this should be ready to merge

joshbruce:
@UziTech - Confilicts. Sorry - probably should have merged this first.

UziTech:
@joshbruce I fixed the conflicts

bin/marked
joshbruce: Thanks for the link.

PR #1495, opened 2019-06-04
Fix nested blockquote error

### LINKED ISSUE #1494: Uncaught TypeError: Cannot parse `>a\n>>b`
```
>a
>>b
```
cause an error `Uncaught TypeError: Cannot read property 'type' of undefined`.

See: [here](https://marked.js.org/demo/?outputType=html&text=%3Ea%0A%3E%3Eb&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tables%22%3A%20true%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master>)

It works fine in `v0.6.2`.

UziTech:
**Marked version:** master

## Description

Error thrown when nesting blockquote

- Fixes #1494 

## Contributor

## Committer

- [ ] Merge PR


UziTech:
I added more tests and fixed a few other bugs in #1464

lib/marked.js
```diff
@@ -1516,6 +1511,24 @@ function splitCells(tableRow, count) {
   return cells;
 }
 
+function lineMatch(str, reg, some) {
```
styfle: What does `some` represent?

UziTech: if it should match some lines or every line.

I could remove it. it isn't really needed here.

UziTech:
I'm actually thinking we should just revert #1464 

#1464 fixes recursion limit for fully nested blockquotes

```
>> a
>> b
```

but it doesn't solve recursion for partially nested blockquotes

```
> a
>> b
```

I'm not sure that we should have a special case for fully nested blockquotes.

markdown-it solves this problem by limiting the recursion to 100 nested blockquotes, and I'm not sure how commonmark.js gets around this.

styfle:
> markdown-it solves this problem by limiting the recursion to 100 nested blockquotes

That sounds pretty reasonable to me. I can't think of many scenarios where you would need that many nested blockquotes besides an email thread.

Issue #1461, opened 2019-04-08
Inserted <p> into <li> on different unordered lists


ybiquitous:
**Describe the bug**

When putting different two unordered lists (`-` and `*`) with blank lines, a `<p>` tag is inserted into `<li>` tag.

E.g.
```md
- a1
- a2

* b1
* b2
```

**To Reproduce**

1. [Marked Demo](https://marked.js.org/demo/?outputType=html&text=-%20a1%0A-%20a2%0A%0A*%20b1%0A*%20b2%0A&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tables%22%3A%20true%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=0.6.2)
2. [CommonMark Demo](https://spec.commonmark.org/dingus/?text=-%20a1%0A-%20a2%0A%0A*%20b1%0A*%20b2%0A%0A)

On terminal:

1. Install marked: `npm install --save marked@0.6.2`
2. Run marked:
```
$ cat <<EOF | ./node_modules/.bin/marked
- a1
- a2

* b1
* b2
EOF
```
3. Actual output (or error) is:
```
<ul>
<li><p>a1</p>
</li>
<li><p>a2</p>
</li>
<li><p>b1</p>
</li>
<li><p>b2</p>
</li>
</ul>
```

**Expected behavior**

Like CommonMark, I think marked should output the following HTML:
```
<ul>
<li>a1</li>
<li>a2</li>
</ul>
<ul>
<li>b1</li>
<li>b2</li>
</ul>
```

UziTech:
Adding an extra new line will make it render correctly but this is a bug.

ybiquitous:
I think so, too. 👍

[...]

PR #1496, opened 2019-06-05
modify the list regex to solve an issue with list rendering #1461

ansuz:
This commit makes many tests fail, but fixes my issue. I don't know enough to tell whether that means my _"fix"_ is really wrong, or if the tests are simply specifying incorrect behaviour for the purpose of consistency.

Please advise.

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -19,7 +19,7 @@ var block = {
   heading: /^ *(#{1,6}) *([^\n]+?) *(?:#+ *)?(?:\n+|$)/,
   nptable: noop,
   blockquote: /^( {0,3}> ?(paragraph|[^\n]*)(?:\n|$))+/,
-  list: /^( {0,3})(bull) [\s\S]+?(?:hr|def|\n{2,}(?! )(?!\1bull )\n*|\s*$)/,
+  list: /^( {0,3})(bull) [\s\S]+?(?:hr|def|\n{1,}(?! )(?!\1bull )\n*|\s*$)/,
   html: '^ {0,3}(?:' // optional indentation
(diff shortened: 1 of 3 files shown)


UziTech:
This is the wrong fix. It breaks commonmark examples like [spec#285](https://spec.commonmark.org/0.29/#example-285)

ansuz:
I haven't found a fix, but I made a test that reproduces the issue.

If you agree that the test is correct then I can see about trying to fix the issue without breaking everything else.

UziTech:
A list with a single blank space is considered a "loose" list ([spec example](https://spec.commonmark.org/0.29/#example-284))

the resulting html [according to commonmark](https://spec.commonmark.org/dingus/?text=*%20pew%0A*%20pew%0A*%20pew%0A%0A*%20pew%0A*%20pew) should be a single list with each element as a paragraph.

```html
<ul>
  <li>
    <p>pew</p>
  </li>
  <li>
    <p>pew</p>
  </li>
  <li>
    <p>pew</p>
  </li>
  <li>
    <p>pew</p>
  </li>
  <li>
    <p>pew</p>
  </li>
</ul>
```

ansuz:
Is that in the original markdown spec, or is it undefined behaviour anywhere other than commonmark?

UziTech:
according to [the original markdown spec](https://daringfireball.net/projects/markdown/syntax#list): 

> If list items are separated by blank lines, Markdown will wrap the items in `<p>` tags in the HTML output.

UziTech:
That is the correct test. Now to find a solution that doesn't break other tests.

UziTech:
This PR should be rebased and code changed in the `/src/` folder. Feel free to reopen this if you are still working on it.

PR #1504, opened 2019-06-26
Sanitize hardening

koczkatamas:
**Marked version:** master
**Markdown flavor:** all

## Description

- Fixes a security vulnerability discussed in email

## Expectation

HTML sanitized properly if `sanitize: true` is used

## Result

There is a known bypass of sanitization

## What was attempted

Fixed the known bypass, but I expect other bypasses, so Marked now warns its users not to depend in `sanitize: true`, but use other means of sanitization (e.g. DOMPurify).

-->

## Contributor

## Committer

- [ ] Merge PR


UziTech:
I think we should have a list of sanitizers in our docs and recommend using one of them instead of just recommending DOMPurify. something like [this list](https://github.com/markedjs/marked/issues/1232#issuecomment-414308183)

test/specs/run-spec.js
```diff
@@ -16,6 +16,9 @@ function runSpecs(title, dir, showCompletionTable, options) {
           spec.options = Object.assign({}, options, (spec.options || {}));
           const example = (spec.example ? ' example ' + spec.example : '');
           const passFail = (spec.shouldFail ? 'fail' : 'pass');
+          if (spec.options.sanitizerRemoveHtml) {
+            spec.options.sanitizer = () => '';
+          }
```
UziTech: Could we change this to just use sanitizer instead of creating a new option

somthing like:

```js
if (spec.options.sanitizer) {
  // eslint-disable-next-line no-eval
  spec.options.sanitizer = eval(spec.options.sanitizer);
}
```

then the frontmatter would just look like:

```markdown
---
sanitize: true
sanitizer: () => ''
---
```

koczkatamas:
> I think we should have a list of sanitizers in our docs and recommend using one of them instead of just recommending DOMPurify. something like [this list](https://github.com/markedjs/marked/issues/1232#issuecomment-414308183)

Works for me!

I personally recommended DOMPurify because its creators (Cure53) are one of (if not the) experts of the fields with a proven track record of finding critical bugs in even most well-known libraries / services.

Sanitize-html also looks good, but I am not sure about the other two.

Should we create a docs .MD file about these options and link to that file from the code and from other documentation?

test/specs/run-spec.js
koczkatamas: Yeah sure, I was thinking about that also, but my security-oriented mind screams every time I have to write down `eval`... 🙈

But we are talking here about only tests, so it's probably okay...

UziTech:
> Should we create a docs .MD file about these options and link to that file from the code and from other documentation?

I would just add the list under the options and link to that section in USING_ADVANCED.md

README.md
```diff
@@ -39,7 +39,7 @@ Also read about:
 
 ## Usage
 
-### Warning: 🚨 Marked does not [sanitize](https://marked.js.org/#/USING_ADVANCED.md#options) the output HTML by default 🚨
+### Warning: 🚨 Marked does not [sanitize](https://marked.js.org/#/USING_ADVANCED.md#options) the output HTML. Please use a sanitize library, like [DOMPurify](https://github.com/cure53/DOMPurify) (recommended), [sanitize-html](https://github.com/apostrophecms/sanitize-html) or [insane](https://www.npmjs.com/package/insane) on the output HTML! 🚨
```
styfle: Some of these use github URLs and some use npm URLs. Let's be consistent use all github URLs. 

https://github.com/bevacqua/insane

koczkatamas: Fixed.

test/specs/run-spec.js
koczkatamas: Modified the code according your suggestion.

README.md
```diff
@@ -39,7 +39,7 @@ Also read about:
 
 ## Usage
 
-### Warning: 🚨 Marked does not [sanitize](https://marked.js.org/#/USING_ADVANCED.md#options) the output HTML by default 🚨
+### Warning: 🚨 Marked does not [sanitize](https://marked.js.org/#/USING_ADVANCED.md#options) the output HTML. Please use a sanitize library, like [DOMPurify](https://github.com/cure53/DOMPurify) (recommended), [sanitize-html](https://github.com/apostrophecms/sanitize-html) or [insane](https://github.com/bevacqua/insane) on the output HTML! 🚨
```
UziTech: Could we have this list in one place and just link to it. I feel like it will be hard to keep these lists in sync if we ever have to update them.

koczkatamas: Any suggestion where would you like to put that list _exactly_? It's your project afterall and I feel like this is the point where this change has nothing to do with the security fix.

Also there are already two README.mds (`/README.md` and `/docs/README.md`) with similar content, so you have to keep those in sync too...

davisjam:
OK, finally had a chance to review this.

If I understand correctly, I see 2-3 problems being discussed.
1. Marked's default sanitizer is problematic.
2. The sanitizer is not always applied properly.
3. (my own addition) Marked does not clearly document when the sanitizer is applied and what the default behavior of the sanitizer is.

I agree that deprecating the sanitization feature seems like a good path forward. Marked's job is converting markdown to HTML. The sanitization of this HTML (whether security-related or not) is an orthogonal problem and should be handled by the application.

lib/marked.js
```diff
@@ -434,7 +434,7 @@ Lexer.prototype.token = function(src, top) {
           : 'html',
         pre: !this.options.sanitizer
           && (cap[1] === 'pre' || cap[1] === 'script' || cap[1] === 'style'),
-        text: cap[0]
+        text: this.options.sanitize ? (this.options.sanitizer ? this.options.sanitizer(cap[0]) : escape(cap[0])) : cap[0]
```
davisjam: This will be a breaking change for anyone using this feature, right?

koczkatamas: Yep, for example if their code allowed XSS then now it won't. 😉

There are no test cases on how the sanitizer should work, so basically any previous change could change this feature also.

joshbruce:
Looking for @davisjam to **approve this one to the point of merge** as I don't have the security knowledge to feel right in doing it. With that said, given the plan to get out of the sanitizer business altogether by 1.0 - I'm pretty okay here.

If the community can hit 100% CommonMark compliance and no known security issues, 1.0 should be released and we can begin the work of refactoring and re-engineering toward the single responsibility of parsing markdown and letting the inputs and outputs be handled by extension or somewhere else altogether.

README.md
joshbruce: @koczkatamas: Thank you so much for this!

@styfle: Where would the best place be to store this list? Thinking it would be a nice list for users moving forward as we are planning to get out of the sanitizer business and working looking for things like this anyway. (I have a note to catch up on what I've missed with all the weirdness of life soon.)

UziTech:
Here is our current gfm and commonmark compliance chart:

```
--------------------------------------
|                 GFM                |
|                                    |
| Tables               8 of  8  100% |
| Task list items      2 of  2  100% |
| Strikethrough        2 of  2  100% |
| Autolinks           11 of 11  100% |
| Disallowed Raw HTML  0 of  1    0% |
--------------------------------------

------------------------------------------------------------
|                        CommonMark                        |
|                                                          |
| Tabs                                     10 of  11   91% |
| Precedence                                1 of   1  100% |
| Thematic breaks                          19 of  19  100% |
| ATX headings                             15 of  18   83% |
| Setext headings                          22 of  27   81% |
| Indented code blocks                     12 of  12  100% |
| Fenced code blocks                       28 of  29   97% |
| HTML blocks                              43 of  43  100% |
| Link reference definitions               25 of  28   89% |
| Paragraphs                                8 of   8  100% |
| Blank lines                               1 of   1  100% |
| Block quotes                             22 of  25   88% |
| List items                               35 of  48   73% |
| Lists                                    13 of  26   50% |
| Inlines                                   1 of   1  100% |
| Backslash escapes                        11 of  13   85% |
| Entity and numeric character references  13 of  17   76% |
| Code spans                               20 of  22   91% |
| Emphasis and strong emphasis             84 of 131   64% |
| Links                                    68 of  87   78% |
| Images                                   15 of  22   68% |
| Autolinks                                15 of  19   79% |
| Raw HTML                                 19 of  21   90% |
| Hard line breaks                         15 of  15  100% |
| Soft line breaks                          2 of   2  100% |
| Textual content                           3 of   3  100% |
------------------------------------------------------------
```

We have a long way to go on some of the sections

lib/marked.js
davisjam: What I mean is, if anyone is currently invoking marked with the sanitize option, the output may change if we land this PR.

1. Yes, they will be (less) XSS-able
2. But since we didn't really document what "sanitize" did, they might have enjoyed its previous behavior (e.g. assuming that by sanitize we meant removing unclosed HTML tags or similar).

As a result I'm not sure how to semver this.

PR #1460, opened 2019-04-03
Improve worst-case performance of inline.text regex

andersk:
The old regex may take quadratic time to scan for potential email addresses starting at every point.  Fix it to avoid scanning from points that would have been in the middle of a previous scan.

**Marked version:**

0.1.3 and later (problem introduced by commit 00f1f7a)

**Markdown flavor:** GitHub Flavored Markdown

## Description

- Fixes DoS issue reported privately.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR); or,
- [ ] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] Draft GitHub release notes have been updated.
- [ ] CI is green (no forced merge required).
- [ ] Merge PR

PR #1507, opened 2019-06-28
Fix breaks

### LINKED ISSUE #1505: 0.6.1 -> 0.6.2 broke `breaks` option (Regression)
**Describe the bug**
After updating from `0.6.1` to `0.6.2`, I noticed that my linebreaks got lost.

**To Reproduce**
`0.6.1` with option `breaks` enabled:
```
A
B
```
output:
```
<p>A<br>B</p>
```
(which is correct)

with `0.6.2` it becomes:
```
<p>A
B</p>
```

UziTech:
**Marked version:** 0.6.2

**Markdown flavor:** GitHub Flavored Markdown with breaks

## Description

the `breaks: true` option broke in v0.6.2 because of #1460 

This reverts the change only when the `breaks` option is set to true.

- Fixes #1505 

## Contributor

## Committer

- [ ] Merge PR


UziTech:
This fixes the breaks but it adds back the redos in [quadradic_br.js](https://github.com/markedjs/marked/pull/1460/files#diff-f1f1b91e274a613c8041d4c39925bb12) when `breaks: true` option is used.

UziTech:
This should probably be released soon since this probably affects a lot of marked users

davisjam:
Let's get this merged. I think fixing a behavior regression is more important than a performance regression.

fdietze:
Thank you!

PR #1511, opened 2019-07-01
use correct options in specs

### LINKED ISSUE #1510: md title parse error
**Describe the bug**
According to markdown syntax:
one of the ways to create a title
```
First Title
=
Second Title
-
```
it doesn't work in marked.js
but results the following in github

First Title
=
Second Title
-

UziTech:
**Marked version:** master

## Description

I am trying to implement the changes from @Feder1co5oave in #1407 

- CommonMark tests are running with `gfm: false`
- GFM tests include CommonMark tests but are using `gfm: true`
- Original tests are running with `pedantic: true`
- **BREAKING CHANGE** Move fences to CommonMark
- **BREAKING CHANGE** Move tables to GFM
- Improve `heading`, `lheading`, and `paragraph`

These changes improve CommonMark compliance significantly:
```
--------------------------------------------------------
|                        CommonMark                    |
|                                         before  after|
| Tabs                                       91%   91% |
| Precedence                                100%  100% |
| Thematic breaks                           100%  100% |
| ATX headings                               83%   94% |
| Setext headings                            81%   89% |
| Indented code blocks                      100%  100% |
| Fenced code blocks                         97%   97% |
| HTML blocks                               100%  100% |
| Link reference definitions                 89%   89% |
| Paragraphs                                100%  100% |
| Blank lines                               100%  100% |
| Block quotes                               88%   92% |
| List items                                 73%   73% |
| Lists                                      50%   58% |
| Inlines                                   100%  100% |
| Backslash escapes                          85%   85% |
| Entity and numeric character references    76%   76% |
| Code spans                                 91%   91% |
| Emphasis and strong emphasis               64%   64% |
| Links                                      78%   78% |
| Images                                     68%   68% |
| Autolinks                                  79%  100% |
| Raw HTML                                   90%   90% |
| Hard line breaks                          100%  100% |
| Soft line breaks                          100%  100% |
| Textual content                           100%  100% |
--------------------------------------------------------
```

## Issues

closes #1407
fixes #1510

## Contributor

## Committer

- [ ] Merge PR


UziTech:
I also agree with @Feder1co5oave that we should remove the `tables` option. I can't see why anyone would want GFM without tables.

lib/marked.js
```diff
@@ -14,10 +14,9 @@
 var block = {
   newline: /^\n+/,
   code: /^( {4}[^\n]+\n*)+/,
-  fences: noop,
+  fences: /^ {0,3}(`{3,}|~{3,})([^`\n]*)\n(?:|([\s\S]*?)\n)(?: {0,3}\1[~`]* *(?:\n+|$)|$)/,
```
davisjam: This is quadratic due to the sub-regex `...(`{3,}|~{3,})([^`\n]*)...` <-- note that the second group does not exclude `~`. An attack string of the form `~~...~` will parse slowly.

lib/marked.js
```diff
@@ -14,10 +14,9 @@
...
   hr: /^ {0,3}((?:- *){3,}|(?:_ *){3,}|(?:\* *){3,})(?:\n+|$)/,
-  heading: /^ *(#{1,6}) *([^\n]+?) *(?:#+ *)?(?:\n+|$)/,
-  nptable: noop,
+  heading: /^ {0,3}(#{1,6}) +([^\n]*?)(?: +#+)? *(?:\n+|$)/,
```
davisjam: Looks OK

lib/marked.js
```diff
@@ -31,9 +30,12 @@ var block = {
     + '|</(?!script|pre|style)[a-z][\\w-]*\\s*>(?=[ \\t]*(?:\\n|$))[\\s\\S]*?(?:\\n{2,}|$)' // (7) closing tag
     + ')',
   def: /^ {0,3}\[(label)\]: *\n? *<?([^\s>]+)>?(?:(?: +\n? *| *\n *)(title))? *(?:\n+|$)/,
+  nptable: noop,
   table: noop,
-  lheading: /^([^\n]+)\n {0,3}(=|-){2,} *(?:\n+|$)/,
-  paragraph: /^([^\n]+(?:\n(?!hr|heading|lheading| {0,3}>|<\/?(?:tag)(?: +|\n|\/?>)|<(?:script|pre|style|!--))[^\n]+)*)/,
+  lheading: /^([^\n]+)\n {0,3}(=+|-+) *(?:\n+|$)/,
```
davisjam: OK

lib/marked.js
```diff
@@ -31,9 +30,12 @@ var block = {
...
+  // regex template, placeholders will be replaced according to different paragraph
+  // interruption rules of commonmark and the original markdown spec:
+  _paragraph: /^([^\n]+(?:\n(?!hr|heading|lheading|blockquote|fences|list|html)[^\n]+)*)/,
```
davisjam: OK

lib/marked.js
UziTech: I just copied that from gfm.fences. So it doesn't make marked any less secure.

Could I fix it just by adding a `~` to the second group?

lib/marked.js
```diff
@@ -91,28 +97,18 @@ block.normal = merge({}, block);
  */
 
 block.gfm = merge({}, block.normal, {
-  fences: /^ {0,3}(`{3,}|~{3,})([^`\n]*)\n(?:|([\s\S]*?)\n)(?: {0,3}\1[~`]* *(?:\n+|$)|$)/,
-  paragraph: /^/,
-  heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/
+  nptable: /^ *([^|\n ].*\|.*)\n *([-:]+ *\|[-| :]*)(?:\n((?:.*[^>\n ].*(?:\n|$))*)\n*|$)/,
```
davisjam: Super-linear.

1. The construct `.*\|.*` can be exploited through a run of `|||...|`. Can we replace with `[^|]*|[^|]*` ?
2. I believe the same problem applies later in the `(?:.*[^>\n ].*` section.

lib/marked.js
```diff
@@ -91,28 +97,18 @@ block.normal = merge({}, block);
...
+  table: /^ *\|(.+)\n *\|?( *[-:]+[-| :]*)(?:\n((?: *[^>\n ].*(?:\n|$))*)\n*|$)/
```
davisjam: Same genre of problem as the `nptable`.

```
if (/^ *\|(.+)\n *\|?( *[-:]+[-| :]*)(?:\n((?: *[^>\n ].*(?:\n|$))*)\n*|$)/.exec(' | \n' + ' '.repeat(50000))) {
  console.log('match');
}
```

(Note, here and earlier I'm just checking regexes, not full exploitability).

lib/marked.js
davisjam: Yes, that would be linear.

UziTech:
I updated the `fences` regex but `nptable` and `table` kept breaking tests when I was trying to fix them. 

`nptable` and `table` were just moved from `block.tables` anyway.

The only changed regexes were `heading`, `lheading`, and `paragraph`

UziTech:
I did remove the `tables` option since this PR would make it do nothing anyway.

test/specs/run-spec.js
```diff
@@ -35,8 +35,8 @@ function runSpecs(title, dir, showCompletionTable, options) {
   });
 }
 
-runSpecs('GFM', './gfm', true, { gfm: true });
-runSpecs('CommonMark', './commonmark', true, { headerIds: false });
-runSpecs('Original', './original', false, { gfm: false });
+runSpecs('GFM', './gfm', true, { gfm: true, pedantic: false, headerIds: false });
+runSpecs('CommonMark', './commonmark', true, { gfm: false, pedantic: false, headerIds: false });
```
styfle: I thought `pedantic` was default false so why assign it here?

test/specs/new/toplevel_paragraphs.md
```diff
@@ -13,13 +13,6 @@ text inside block code
 paragraph before hr
 * * *
 
-paragraph before head with hash
```
styfle: Why is this part of the test removed?

docs/USING_ADVANCED.md
```diff
@@ -25,7 +25,6 @@ marked.setOptions({
   },
   pedantic: false,
   gfm: true,
-  tables: true,
```
styfle: Can `tables: true` be used when `gfm: false`? It seems like the answer is no, but I want to make sure there are no other cases I'm missing.

test/specs/run-spec.js
UziTech: Ya I thought it was better to be verbose so it is easier to know what is going on, and in case the defaults change in the future.

docs/USING_ADVANCED.md
UziTech: no `tables` is like `breaks` and only works when `gfm: true` is set

test/specs/new/toplevel_paragraphs.md
UziTech: It was moved to another file with `pedantic: true` I'm not entirely sure why but it was like that in #1407

Lmillan123:
:+1:

PR #1513, opened 2019-07-01
Remove baseUrl if its just a slash

### LINKED ISSUE #1466: baseUrl with relative prefix not working as expected.
**Describe the bug**
For usage in github pages I'd like to use the baseUrl option to add a prefix of the project name for relative links. For instance with a github pages site like ``` matt.github.io/derp``` with ```derp``` being the project name. I'd like relative links to change from ```/link``` to ```/derp/link```.

Marked options
```javascript
const options = {
    baseUrl: '/derp/',
}
```

Markdown
```markdown
![alt text](/favicon.ico "relative")'
```
Output html
```html
<p><img src="/derp/favicon.ico" alt="alt text" title="relative"></p>
```
What is actually happening is:
```html
<p><img src="/derp//favicon.ico" alt="alt text" title="relative"></p>
```

Maybe it is as simple as not inverting rtrim here:
https://github.com/markedjs/marked/blob/master/lib/marked.js#L1417

**To Reproduce**
This code sandbox should show the issue. Check the console
https://codesandbox.io/s/9ql03o5114

I was able to reproduce this in both 0.5.0 and 0.6.2.

**Expected behavior**
I expect html output to look like this:
```html
<p><img src="/derp/favicon.ico" alt="alt text" title="relative"></p>
```

VitaliyR:
**Marked version:** master

## Description

- Fixes #1466 

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -1119,6 +1119,9 @@ function Parser(options) {
   this.tokens = [];
   this.token = null;
   this.options = options || marked.defaults;
+  if (this.options.baseUrl && this.options.baseUrl.trim() === '/') {
+    delete this.options.baseUrl;
+  }
   this.options.renderer = this.options.renderer || new Renderer();
   this.renderer = this.options.renderer;
   this.renderer.options = this.options;
--- a/test/specs/new/base_url_root.html
+++ b/test/specs/new/base_url_root.html
@@ -1,0 +1,3 @@
+<p>
+  <img alt="Image" src="/images/pic01.jpg" />
+</p>
--- a/test/specs/new/base_url_root.md
+++ b/test/specs/new/base_url_root.md
@@ -1,0 +1,4 @@
+---
+baseUrl: "/"
+---
+![Image](/images/pic01.jpg)


UziTech:
I was just looking at the code and it looks like the issue is that `baseUrl` is supposed to be an absolute url.

https://github.com/markedjs/marked/blob/124c57690eaf97b0b173c8bcfb780ba7ec74db7e/lib/marked.js#L1435-L1436 

I'm not sure if that was the intention when `baseUrl` was proposed.

@styfle Instead of creating a special case for `baseUrl: '/'` should we fix `resolveUrl` to work with a relative `baseUrl`?

We should create a test file like [relative_urls.md](https://github.com/markedjs/marked/blob/master/test/specs/new/relative_urls.md) with `baseUrl: '/base/'`.

styfle:
The docs say `baseUrl` is a prefix, but the name `URL` makes me think it should be a complete URL like the `<base>` tag from HTML https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base.

UziTech:
So the ultimate questions is should we update `resolveUrl` to work with a relative `baseUrl` or update the docs to make it clear that `baseUrl` should be an absolute url?

I'm not exactly sure why `baseUrl` was added as an option. It looks like @joshbruce added it in https://github.com/markedjs/marked/commit/e5b2998326bd3911cff76f6b79e1aad9469e24fa.

VitaliyR:
I think typically baseUrl should be available - and works just as `path.join`, not concatenating. I mean in case of href of link/img in the parser -> path.join(baseUrl || '', href)

UziTech:
I think it works more like `path.resolve(base, href)`.

- Any href starting with `/` is considered absolute relative to the domain.
- Any href starting with `//` will replace the domain but not the protocol.
- Any href starting with `.*://` is considered absolute and the baseUrl will be ignored.

styfle:
I'm starting to think `baseUrl` was a bad choice. Instead, we should have a callback to allow the user to specify what to do with urls.

Just like the `highlight` function, we could have a `url` function.

Or is this a job for the Renderer?

UziTech:
I think it should be a job for the renderer.

I'm not exactly sure why this option was added in the first place. It was added by @joshbruce in https://github.com/markedjs/marked/commit/e5b2998326bd3911cff76f6b79e1aad9469e24fa without a good explanation

styfle:
@joshbruce Can we remove this `baseUrl` option in favor of overriding the Renderer?

VitaliyR:
@styfle remove it at all? So is there a way to prepend some url to all urls which are transformed by marked? 

Or it will be via overriding Renderer.image && link?

UziTech:
I don't think we will remove it since there are definitely people using it but this is one option we will be deprecating in the future and probably have an extension that overrides the renderer.

UziTech:
I feel like in the meantime we should make `baseUrl` work with relative urls. @VitaliyR do you want to change this PR to handle that or should I create a new PR?

VitaliyR:
@UziTech 
What exactly do you mean? Because now:

baseUrl: 'http://example.com/root'
url: /pic/img.jpg

result: http://example.com/pic/img.jpg (without `root` in the path)

So what should we do in this case? Because we might break something

UziTech:
the baseUrl needs to end with a slash

```js
marked("![](pic/img.jpg)", {baseUrl: "http://example.com/root/"});
// <p><img src="https://example.com/root/pic/img.jpg" alt=""></p>
```

PR #1543, opened 2019-08-27
Update htmldiffer async matcher

UziTech:
**Marked version:** master

## Description

update html-differ and make tests async

This will remove the [security alert](https://github.com/markedjs/marked/network/alerts) for diff

I had to remove testing for node v4 since html-differ uses async/await but will still catch any es6 code with eslint.

## Contributor

## Committer

- [x] Merge PR


package.json
```diff
@@ -29,20 +29,21 @@
     "html"
   ],
   "devDependencies": {
-    "@markedjs/html-differ": "^2.0.1",
+    "@markedjs/html-differ": "^3.0.0",
+    "acorn": "^7.0.0",
```
styfle: Is acorn used?

UziTech: I'm not sure where that came from? everything seems to work without it.

PR #1552, opened 2019-09-24
Typo: Fix link to USING_PRO.md

vmlaguna:
**Marked version:** 0.7.0

**Markdown flavor:** n/a

## Description

Documentation link was not pointing to the correct location, because link can't be relative, hardcoding the official documentation website.

## Expectation

Documentation works seamlessly. 

## Result

N/A

## What was attempted

N/A

PR #1526, opened 2019-07-18
relative baseUrl

UziTech:
**Marked version:** master

## Description

Allow `baseUrl` option to be a relative url.

- Closes #1513 
- Fixes #1466 

## Expectation

```html
<p><img src="/favicon.ico" alt="alt text" title="relative"></p>
```

## Result

```html
<p><img src="/derp//favicon.ico" alt="alt text" title="relative"></p>
```

## What was attempted

```js
marked('![alt text](/favicon.ico "relative")', {baseUrl: "/derp/"});
```

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression.
- [x] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] Draft GitHub release notes have been updated.
- [ ] CI is green (no forced merge required).
- [ ] Merge PR

PR #1571, opened 2019-12-04
0.8.0

UziTech:
# Release Notes

## Breaking changes

- Remove substitutions #1532
- Separate source into modules #1563 #1572 #1573 #1575 #1576 #1581

## Fixes

- Fix relative urls in `baseUrl` option #1526 
- Loose task list #1535
- Fix image parentheses #1557

## Docs

- Update examples with es6+ #1521
- Fix link to USING_PRO.md page #1552
- Fix typo in USING_ADVANCED.md #1558
- Node worker threads are stable #1555

## Dev Dependencies

- Update deps #1516
- Update eslint #1542
- Update htmldiffer async matcher #1543

---

# Publisher

- [x] `$ npm version` has been run.
- [x] Release notes in [draft GitHub release](https://github.com/markedjs/marked/releases) are up to date
- [x] Release notes include which flavors and versions of Markdown are supported by this release
- [ ] Committer checklist is complete.
- [ ] Merge PR.
- [ ] Publish GitHub release using `master` with correct version number.
- [ ] `$ npm publish` has been run.
- [ ] Create draft GitHub release to prepare next release.

Note: If merges to `master` occur after submitting this PR and before running `$ npm pubish` you should be able to

1. pull from `upstream/master` (`git pull upstream master`) into the branch holding this version,
2. run `$ npm run build` to regenerate the `min` file, and
3. commit and push the updated changes.

# Committer

In most cases, this should be someone different than the publisher.

- [x] Version in `package.json` has been updated (see [PUBLISHING.md](https://github.com/markedjs/marked/blob/master/docs/PUBLISHING.md)).
- [x] The `marked.min.js` has been updated; or,
- [ ] release does not change library.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "marked",
-  "version": "0.7.0",
+  "version": "0.8.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "marked",
   "description": "A markdown parser built for speed",
   "author": "Christopher Jeffrey",
-  "version": "0.7.0",
+  "version": "0.8.0",
   "main": "./src/marked.js",
   "bin": "./bin/marked",
   "man": "./man/marked.1",


Andarist:
While this is going to be a breaking change - maybe you would like to consider changing a few more things? I would like to propose changing exported shape for better tree-shakeability. Mainly I don't believe using a default export is a good fit here (especially the way it is used right now) and also I would like to propose to skip using static properties on functions - but this one would have to be evaluated first because by a quick glance at the source code I'm not sure why they are used in the first place (maybe they are for some reason needed, not sure)

UziTech:
@Andarist if you want to create a PR with your proposed changes we could squeeze it in here if we think it is necessary.

Andarist:
@UziTech Ok, I'm going to prepare something later tonight. Stay tuned!

Andarist:
Keep in mind that your main & module entries (cjs & esm) have incompatible shapes right now. `module.exports = value`  is not an equivalent to `export default value`. This can break your consumers in a situation as follows:
```js
// webpack consumer
require('./foo')
// foo.js
var localMarked = require('marked') // webpack will expose `{ default: marked }` as `localMarked`
localMarked('**md**') // error!
```

UziTech:
This seems to be a known issue with webpack https://github.com/webpack/webpack/issues/4742

I have found the following solutions:
1. Leave it as is since this is a known issue with webpack and document the solution.
2. remove `module` field in `package.json`
3. Create named export `marked`
   This would allow anyone to use
   ```js
   ```
   We would still keep the default export for anyone who wants to use it.

I prefer #1 so when webpack fixes this we won't need to change anything.

styfle:
I don't think webpack is going to change, the issue is over 2 years old. I also think the [maintainer is correct](https://github.com/webpack/webpack/issues/4742#issuecomment-295115576), packages should not export different things from module and main.

What are the implications for number 2: remove `module` field from package.json? That seems to be the best solution.

web-padawan:
> What are the implications for number 2: remove module field from package.json? That seems to be the best solution.

Some existing tools rely on that field as indicator of the "modern" packages shipping ES modules.

See for example `@pika/web` https://github.com/pikapkg/web and all the tools by @pikapkg

UziTech:
@styfle I believe `#2` would prevent code splitting/tree shaking although I'm not sure if that is actually an issue with a library like marked that only has one export and pretty much the entire codebase is used for a single purpose.

Without the `module` field anyone wanting to use the esm version in node would need to use `import marked from 'marked/lib/marked.esm.js';`

We could move the esm build to `/esm/index.js` so they could use `import marked from 'marked/esm';`

UziTech:
I think removing the module field would be the easiest solution. It would make marked backwards compatible and we can always fix code splitting/tree shaking in a later version if we need to.

styfle:
> Without the module field anyone wanting to use the esm version in node 

I don't think Node supports the `module` field.

Perhaps you are thinking of the [`"type": "module"`](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field) property?

Andarist:
Yes - if you dont want to change CJS export shape right now then I would also recommend dropping the "module" field right now. I would love "module" being added - as the community, I believe we should move towards ESM as much as we can as it's the part of the language - but only if it's done right (so without any drawbacks).

At the moment `marked` is not tree-shakeable anyway (but it could! if you'd be willing to change exports shape 😉 ).

UziTech:
I think this is ready to publish once we get one more approval from @joshbruce or @davisjam

UziTech:
v0.8.0 has been released 🎉

@Andarist Thanks for your PRs and the help with this release.

Andarist:
No problem. If you'd like to discuss tree-shakeability and changing exports shape in the future I would gladly help you out on this one. Configuring things "right" for bundlers and other possible consumers is surprisingly hard and I have good expertise on that matter.

styfle:
Thanks @Andarist 👍 

A related issue has been opened in pikapkg that you might want to engage in.

https://github.com/pikapkg/analyze-npm/issues/9

Issue #1514, opened 2019-07-01
allow CommonMark uncompliant mixing of Markdown in HTML blocks


NattyNarwhal:
**What pain point are you perceiving?.**

The change to >=0.4 significantly broke expectations for me. Before PR #1135, it was easy to do say, `<p class="foo">*bar*</p>` or indented HTML (for blocks of `<div>`) and have it do as a user, what I expect it to. However, the CommonMark behaviour introduced is unintuitive and it took be a while to figure out what was happening. For me, it's a regression.
 
**Describe the solution you'd like**

Some option to revert to the 0.3.x behaviour. I don't care about CommonMark compliance, and would rather mix markup freely.

UziTech:
Every change breaks someone's [workflow](https://xkcd.com/1172/) 😄

You can pin marked to v0.3.19 to get the old behavior

pyromanci:
There is a solution for this, but needs you do some extra work. That is to override the html() function of the Renderer. This is what I did for div tags. 

```
const _renderer = new marked.Renderer();
				
_renderer.html = function(html) {
	if(_.startsWith(_.trim(html), '<div') && _.endsWith(_.trim(html), '</div>')){
		const openTag = html.substring(0, html.indexOf('>')+1);
		html = html.substring(html.indexOf('>')+1);
		html = html.substring(0, html.lastIndexOf('</div>'));
		return `${openTag} ${marked(html)} </div>`;
	}
	return html;
}
```

PR #1574, opened 2019-12-04
Remove changeDefaults

Andarist:
As far as I can tell this has stayed unused - `changeDefaults` was only changing a local variable and that didn't affect anything. There are also no tests associated with it.

I've wanted also to remove `defaults` as it seems a duplication of `getDefaults`, but you actually indirectly rely on it - because it's not a constant value, this actually is a static property of `marked` and you allow for consumers to mutate this with `setOptions`. I would like to strongly argue that this is not good - if one wants to use marked with custom options they should provide it for a particular call/instance. It's highly dangerous to allow mutation on such a global/shared state - other module using marked and mutating this might affect how marked behave for me and this is totally unexpected from the user's point of view.

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -1,6 +1,4 @@
-let defaults = getDefaults();
-
-function getDefaults() {
+module.exports = function getDefaults() {
   return {
     baseUrl: null,
     breaks: false,
@@ -19,14 +17,4 @@ function getDefaults() {
     smartypants: false,
     xhtml: false
   };
-}
-
-function changeDefaults(newDefaults) {
-  defaults = newDefaults;
-}
-
-module.exports = {
-  defaults,
-  getDefaults,
-  changeDefaults
 };
--- a/src/InlineLexer.js
+++ b/src/InlineLexer.js
@@ -1,5 +1,5 @@
 const Renderer = require('./Renderer.js');
-const { defaults } = require('./defaults.js');
+const getDefaults = require('./defaults.js');
 const { inline } = require('./rules.js');
 const {
   findClosingBracket,
@@ -11,7 +11,7 @@ const {
  */
 module.exports = class InlineLexer {
   constructor(links, options) {
-    this.options = options || defaults;
+    this.options = options || getDefaults();
     this.links = links;
     this.rules = inline.normal;
     this.options.renderer = this.options.renderer || new Renderer();
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -1,4 +1,4 @@
-const { defaults } = require('./defaults.js');
+const getDefaults = require('./defaults.js');
 const { block } = require('./rules.js');
 const {
   rtrim,
@@ -13,7 +13,7 @@ module.exports = class Lexer {
   constructor(options) {
     this.tokens = [];
     this.tokens.links = Object.create(null);
-    this.options = options || defaults;
+    this.options = options || getDefaults();
     this.rules = block.normal;
 
     if (this.options.pedantic) {
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -2,7 +2,7 @@ const Renderer = require('./Renderer.js');
 const Slugger = require('./Slugger.js');
 const InlineLexer = require('./InlineLexer.js');
 const TextRenderer = require('./TextRenderer.js');
-const { defaults } = require('./defaults.js');
+const getDefaults = require('./defaults.js');
 const {
   merge,
   unescape
@@ -15,7 +15,7 @@ module.exports = class Parser {
   constructor(options) {
     this.tokens = [];
     this.token = null;
-    this.options = options || defaults;
+    this.options = options || getDefaults();
     this.options.renderer = this.options.renderer || new Renderer();
     this.renderer = this.options.renderer;
     this.renderer.options = this.options;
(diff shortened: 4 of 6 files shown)


UziTech:
changeDefaults was added to allow backwards compatibility for the way `marked.setOptions` worked.

When `setOptions` was used to change the options it changed the defaults for every other call to `require('marked')`. I know it's not good to have mutable defaults but unfortunately certain dependents have come to rely on that feature.

Since `defaults` are required in multiple files it could no longer be a global so I used `changeDefaults` to actually change what `require('./defaults.js')` returns for defaults.

src/defaults.js
```diff
@@ -19,14 +17,4 @@ function getDefaults() {
     smartypants: false,
     xhtml: false
   };
-}
-
-function changeDefaults(newDefaults) {
-  defaults = newDefaults;
```
UziTech: It looks like this should actually be `module.exports.defaults = newDefaults;`

UziTech: I think this worked as it is when the files were ES modules but when I switched to CommonJS I forgot to update this line.

UziTech:
this is actually a known issue https://github.com/markedjs/marked/issues/907 and will be deprecated at some point.

src/defaults.js
Andarist: This could have happened because ESM exports are actually live bindings but it doesn't work like that with CJS (as you have noticed you'd have to assign it to module.exports to make it work).

Andarist:
@UziTech as this is a long-standing issue (dated to 2017), wouldn't it make sense to fix this right now at a convenient opportunity of releasing other breaking changes anyway?

UziTech:
I would like to but I think that would affect too many dependents. Once we reach 1.0 we will start deprecating stuff.

Andarist:
I'm not a native speaker, so discussing in a friendly-sounding-way is a real challenge to me. Please assume my best intentions.

I'd like to argue that this is, in my opinion, a bad decision. You are right now releasing other breaking changes and bumping your minor version accordingly (it's also worth noting that for npm minor bump for 0.x versions is treated as a major bump and by definition, it's not treated ad compatible with older versions). It's a perfect time to release other breaking changes. My premise is that breaking changes are **good** - if only they are communicated well (and by creating a proper changelog entry + raising a minor you don't have to do anything extra). Breaking changes exist for a reason - they are introduced to fix issues with the older design and are **beneficial** to your consumers. After all - we can agree that the current design in this regard is bad and there is a plan to remove it, but by keeping it in place you are putting your consumers at risk of wasting numerous hours of debugging, because it's not an obvious thing and might take a while to find the problem - especially for less experienced developers. 

> I would like to but I think that would affect too many dependents.

By introducing any breaking changes as new major (in this case minor) version you cannot affect your dependents automatically - they have to opt-in into upgrading to a new version, when doing so they should assume that the bump was there for a reason and they should check out your changelog. So it's not like you can accidentally affect their projects. To aid the migration you can even add development-only code (hidden behind `process.env.NODE_ENV` check) that would inform them about things being removed in a more explicit way - you can even throw an error in such a case as failing fast is most noticeable for developers and will help them migrate quickly to a new release.

UziTech:
Marked isn't only used through npm. Marked is a client side library so it could be used by just including
```html
```
in the html, in which case the version number doesn't mean anything.

We are trying to move slowly since we have lots of dependencies and not a lot of time to work on marked. Every change/fix breaks someones workflow (https://github.com/markedjs/marked/issues/1514) so if we don't have to change something than we probably shouldn't.

Plus I'm pretty sure the original intent was for setOptions to change options **globally**.

Andarist:
> Marked isn't only used through npm. Marked is a client side library so it could be used by just including

This is a valid thing to consider now - but if you are not going to change anything about your CDN URL policy then this thing will keep you hostage **forever**. We all use versioned software for exactly this reason - it has benefits for both sides, gives stability for consumers but also gives an ability to change stuff for authors. I don't believe that providing an URL pointing always to the latest version is a good thing.

I understand that you can't just remove it now - but what I would propose is to commit a built version of marked as this file, maybe with some kind of a deprecation warning included and leave it untouched for the rest of eternity. In the meantime develop the package and only provide versioned URLs.

> We are trying to move slowly since we have lots of dependents and not a lot of time to work on marked.

Totally understandable - your work is highly appreciated ❤️ But also because the development of the package has a slow pace I think it's only the additional argument for grouping breaking changes together as much as possible. Taking a chance to introduce a wanted change when there is a time for it.

> Every change/fix breaks someones workflow (#1514) so if we don't have to change something than we probably shouldn't.

Yes - that's always true. Even a bugfix might break somebody, that's why semver in practice is just a guideline - otherwise every change would have to result in a new major version.

> Plus I'm pretty sure the original intent was for setOptions to change options globally.

Right, this was probably the original intent, but I think we've agreed before (or maybe I've misunderstood your words) that, after seeing this approach downsides, doing such a thing globally is not a good thing/design and ideally, in the future, it should be removed.

UziTech:
Our plan is to get all Common Mark and GFM spec tests passing then release v1.0 and deprecate a bunch of things and remove them in v2.0.

The process is going to take a while but if you want to help speed it up you can try to get the specs with `"shouldFail": true` to pass when you set it to `false`.

https://github.com/markedjs/marked/blob/ed18cd58218ed4ab98d3457bec2872ba1f71230e/test/specs/commonmark/commonmark.0.29.json#L57

PR #1593, opened 2020-01-07
Add tests to show parser breaking when links end with underscores

StefKors:
**Marked version:** 0.8.0

## Description

When parsing a block of markdown content that contain inline links, links ending with underscores are not parsed correctly. The following link https://www.google.com/maps_ should be parsed as [https://www.google.com/maps_](https://www.google.com/maps_)

I've created a test case which highlights the issue. I'll dig around to see if I can find where this happens.

## Expectation

Expected: [https://www.google.com/maps_](https://www.google.com/maps_)

## Result

Result: [https://www.google.com/maps](https://www.google.com/maps)_

### DIFF
--- a/test/specs/new/underscore_link.html
+++ b/test/specs/new/underscore_link.html
@@ -1,0 +1,1 @@
+<p><a href="https://www.google.com/maps_">https://www.google.com/maps_</a></p>
--- a/test/specs/new/underscore_link.md
+++ b/test/specs/new/underscore_link.md
@@ -1,0 +1,4 @@
+---
+sanitize: true
+---
+https://www.google.com/maps_


UziTech:
IT looks like github doesn't include the underscore as part of the link.

https://www.google.com/maps_

accroding to the [GFM spec](https://github.github.com/gfm/#autolinks-extension-):

> Trailing punctuation (specifically, `?`, `!`, `.`, `,`, `:`, `*`, `_`, and `~`) will not be considered part of the autolink, though they may be included in the interior of the link

You can fix this by using brackets around the link

```
<https://www.google.com/maps_>
```

<https://www.google.com/maps_>

StefKors:
@UziTech I see, is it possible to add support for this through an option or flag? or would that break the specs?

UziTech:
We are looking at dropping most of the current options in the future in favor of just going with the spec and making marked more extensible.

So unless you are able to change the spec this doesn't seem like something marked will support directly in the future.

Your best bet would be to create a package that extends marked and changes the behavior.

StefKors:
Thanks for the info!

PR #1598, opened 2020-02-05
Fix GFM tables not breaking on block-level structures

### LINKED ISSUE #1467: Heading immediately following a table is being rendered in the table element
**Describe the bug**
I have some Markdown that contains a table immediately followed by a heading on a new line. The subsequent heading is being rendered as part of the table. 

**To Reproduce**
Steps to reproduce the behaviour:

Markdown Sample:

```markdown
| Heading | Description |
| ----------- | ----- |
| Heading 1 | Some description | 
| Heading 2 | Some other description |
## A Title that is not a part of the table
```

HTML output:

```html
<table>
  <thead>
    <tr>
      <th>Heading</th>
      <th>Description</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Heading 1</td>
      <td>Some description</td>
    </tr>
    <tr>
      <td>Heading 2</td>
      <td>Some other description</td>
    </tr>
    <tr>
      <td>## A title that is not a part of the table</td>
      <td></td>
    </tr>
  </tbody>
</table>
```

Demo:
[Marked Demo](https://marked.js.org/demo/?text=%7C%20Heading%20%7C%20Description%20%7C%0A%7C%20-----------%20%7C%20-----%20%7C%0A%7C%20Heading%201%20%7C%20Some%20description%20%7C%20%0A%7C%20Heading%202%20%7C%20Some%20other%20description%20%7C%0A%23%23%20A%20title%20that%20is%20not%20a%20part%20of%20the%20table&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tables%22%3A%20true%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)

**Expected behaviour**
The heading should be rendered separately from the table.

calculuschild:
**Marked version:** 8.0

## Description

- Fixes #1467 

## What was attempted

Edited rules.js to break GFM tables properly using similar syntax used to break paragraphs.

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -111,8 +111,6 @@ block.gfm.table = edit(block.gfm.table)
   .replace('tag', block._tag) // pars can be interrupted by type (6) html blocks
   .getRegex();
 
-console.log(block.gfm.table);
-
 /**
  * Pedantic grammar (original John Gruber's loose markdown specification)
  */
--- a/src/rules.js
+++ b/src/rules.js
@@ -105,7 +105,7 @@ block.gfm.table = edit(block.gfm.table)
   .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
   .replace('blockquote', ' {0,3}>')
   .replace('code', ' {4}[^\\n]')
-  .replace('fences', ' {0,3}(?:`{3,}|~{3,})[^`\\n]*\\n')
+  .replace('fences', ' {0,3}(?:`{3,}(?=[^`\\n]*\\n)|~{3,})[^\\n]*\\n')
   .replace('list', ' {0,3}(?:[*+-]|1[.)]) ') // only lists starting from 1 can interrupt
   .replace('html', '</?(?:tag)(?: +|\\n|/?>)|<(?:script|pre|style|!--)')
   .replace('tag', block._tag) // pars can be interrupted by type (6) html blocks
--- a/src/rules.js
+++ b/src/rules.js
@@ -69,7 +69,7 @@ block.html = edit(block.html, 'i')
 
 block.paragraph = edit(block._paragraph)
   .replace('hr', block.hr)
-  .replace('heading', ' {0,3}#{1,6} +')
+  .replace('heading', ' {0,3}#{1,6} ')
   .replace('|lheading', '') // setex headings don't interrupt commonmark paragraphs
   .replace('blockquote', ' {0,3}>')
   .replace('fences', ' {0,3}(?:`{3,}|~{3,})[^`\\n]*\\n')
@@ -101,7 +101,7 @@ block.gfm = merge({}, block.normal, {
 
 block.gfm.table = edit(block.gfm.table)
   .replace('hr', block.hr)
-  .replace('heading', ' {0,3}#{1,6} +')
+  .replace('heading', ' {0,3}#{1,6} ')
   .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
   .replace('blockquote', ' {0,3}>')
   .replace('code', ' {4}[^\\n]')
--- a/src/rules.js
+++ b/src/rules.js
@@ -104,7 +104,7 @@ block.gfm.table = edit(block.gfm.table)
   .replace('heading', ' {0,3}#{1,6} +')
   .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
   .replace('blockquote', ' {0,3}>')
-  .replace('code', ' {4}[^\\n]+')
+  .replace('code', ' {4}[^\\n]')
   .replace('fences', ' {0,3}(?:`{3,}|~{3,})[^`\\n]*\\n')
   .replace('list', ' {0,3}(?:[*+-]|1[.)]) ') // only lists starting from 1 can interrupt
   .replace('html', '</?(?:tag)(?: +|\\n|/?>)|<(?:script|pre|style|!--)')
--- a/src/rules.js
+++ b/src/rules.js
@@ -94,8 +94,24 @@ block.normal = merge({}, block);
 
 block.gfm = merge({}, block.normal, {
   nptable: /^ *([^|\n ].*\|.*)\n *([-:]+ *\|[-| :]*)(?:\n((?:.*[^>\n ].*(?:\n|$))*)\n*|$)/,
-  table: /^ *\|(.+)\n *\|?( *[-:]+[-| :]*)(?:\n((?: *[^>\n ].*(?:\n|$))*)\n*|$)/
+  table: '^ *\\|(.+)\\n' // Header
+    + ' *\\|?( *[-:]+[-| :]*)' // Align
+    + '(?:\\n((?:(?!^|>|\\n| |hr|heading|lheading|code|fences|list|html).*(?:\\n|$))*)\\n*|$)' // Cells
 });
+
+block.gfm.table = edit(block.gfm.table)
+  .replace('hr', block.hr)
+  .replace('heading', ' {0,3}#{1,6} +')
+  .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
+  .replace('blockquote', ' {0,3}>')
+  .replace('code', ' {4}[^\\n]+')
+  .replace('fences', ' {0,3}(?:`{3,}|~{3,})[^`\\n]*\\n')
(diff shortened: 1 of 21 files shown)


src/rules.js
```diff
@@ -94,9 +94,25 @@ block.normal = merge({}, block);
...
+  .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
+  .replace('blockquote', ' {0,3}>')
+  .replace('code', ' {4}[^\\n]+')
+  .replace('fences', ' {0,3}(?:`{3,}|~{3,})[^`\\n]*\\n')
+  .replace('list', ' {0,3}(?:[*+-]|1[.)]) ') // only lists starting from 1 can interrupt
+  .replace('html', '</?(?:tag)(?: +|\\n|/?>)|<(?:script|pre|style|!--)')
+  .replace('tag', block._tag) // pars can be interrupted by type (6) html blocks
+  .getRegex();
+
+console.log(block.gfm.table);
```
styfle: Remove `console.log`

UziTech:
By including it all in the regex I'm a little worried about this introducing a new [ReDos](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS) vulnerability.

I'll look into this but it might take some time.

/cc @davisjam

src/rules.js
calculuschild: Got it. Good catch.

calculuschild:
> By including it all in the regex I'm a little worried about this introducing a new [ReDos](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS) vulnerability.

If it helps, the only change was in the detection of table cells (the third captured regex group) where I replaced 

```
 *[^>\n ]
```

with 

```
(?!^|>|\\n| |hr|heading|lheading|code|fences|list|html)
```

and then apply the helper.js "edit" function in identical manner to the paragraph breaking logic earlier in the document. Note that this new bit of code does not [apply repetition (“+”, “*”) to a complex subexpression](https://www.geeksforgeeks.org/understanding-redos-attack/)

Given that, if the paragraph code has been deemed safe from REDOS attacks there shouldn't be anything different here except that I also include regex for `code` and `lheading` which are ignored in the paragraph regex.

test/specs/new/lheading_following_table.html
```diff
@@ -0,0 +1,19 @@
...
+      <td>bar</td>
+      <td>foo</td>
+    </tr>
+    <tr>
+      <td>baz</td>
+      <td>boo</td>
+    </tr>
+  </tbody>
+</table>
+<h1 id="title">title</h1>
```
styfle: GitHub seems to render this one differently

image

calculuschild: Hmmm.... The spec says ["The table is broken at the first empty line, or **beginning of another block-level structure**"](https://github.github.com/gfm/#example-201), so that would seem to break their rule. If it's an issue, it's a simple matter of just removing the associated lines that check for `lheading`.

styfle: Good point, your test seems correct. I also double checked to make sure our GFM spec tests included 201 `[extension] Tables` and it is passing.

src/rules.js
```diff
@@ -94,9 +94,23 @@ block.normal = merge({}, block);
...
   nptable: /^ *([^|\n ].*\|.*)\n *([-:]+ *\|[-| :]*)(?:\n((?:.*[^>\n ].*(?:\n|$))*)\n*|$)/,
-  table: /^ *\|(.+)\n *\|?( *[-:]+[-| :]*)(?:\n((?: *[^>\n ].*(?:\n|$))*)\n*|$)/
+  table: '^ *\\|(.+)\\n' // Header
+    + ' *\\|?( *[-:]+[-| :]*)' // Align
+    + '(?:\\n((?:(?!^|>|\\n| |hr|heading|lheading|code|fences|list|html).*(?:\\n|$))*)\\n*|$)' // Cells
 });
 
+block.gfm.table = edit(block.gfm.table)
+  .replace('hr', block.hr)
+  .replace('heading', ' {0,3}#{1,6} +')
```
UziTech:
```suggestion
  .replace('heading', ' {0,3}#{1,6} ')
```

No need to check for more than one space if one will match.

davisjam:
Is this stable enough for me to look at the regexes for ReDoS?

src/rules.js
UziTech: This could also be fixed in the paragraph regex.

src/rules.js
```diff
@@ -94,9 +94,23 @@ block.normal = merge({}, block);
...
+  .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
+  .replace('blockquote', ' {0,3}>')
+  .replace('code', ' {4}[^\\n]+')
```
UziTech:
```suggestion
  .replace('code', ' {4}[^\\n]')
```
No need to check for more than one char if one will match.

UziTech:
@davisjam yes that would be great. I kind of looked it over but I'm sure you are better at finding them.

test/specs/new/lheading_following_table.html
UziTech: from the [spec](https://github.github.com/gfm/#setext-heading):

> In general, a setext heading need not be preceded or followed by a blank line. However, it cannot interrupt a paragraph, so when a setext heading comes after a paragraph, a blank line is needed between them.

Maybe since it cannot interrupt a paragraph it also shouldn't interrupt a table? I'm leaning toward following the implementation and posing this as an issue to [GitHub](https://github.com/github/cmark-gfm) to see what they have to say.

We could leave the regex in and just comment it out with a note so we can change it easily in the future.

src/rules.js
```diff
@@ -94,9 +94,23 @@ block.normal = merge({}, block);
...
+  .replace('fences', ' {0,3}(?:`{3,}|~{3,})[^`\\n]*\\n')
```
calculuschild:
```suggestion
  .replace('fences', ' {0,3}(?:`{3,}[^`\\n]*|~{3,}[^\\n]*)\\n')
```
Looking at the other snippets I copied from the paragraph regex, I think the check for fences might also be out of spec. Looks like only code fences made with ` [cannot have backticks in the rest of the line](https://github.github.com/gfm/#example-115), but fences made with ~ [can have both backticks and tildes](https://github.github.com/gfm/#example-116).

Not sure if there's a cleaner way to make this fix, but If it's going into this PR it should also be fixed in the paragraph regex. Any thoughts?

UziTech: maybe this should be a separate pr looks like the fences regex should be changed to pass example 116:

```diff
- (`{3,}|~{3,})([^`~\n]*)\n
+ (`{3,}(?=[^`\n]*\n)|~{3,})([^\n]*)\n
```

calculuschild: Does the lookahead `(?=` make the fence regex less prone to REDOS as well? I can't remember but it seems like lookaheads acted like atomic groups which probably helps.

UziTech: spec tests marked `shouldFail: true` (like example 116) are ones that are currently failing.

if you remove that line from the json file it will fail in the tests. With this change it will pass that example. I'm not sure about being less prone to redos.

calculuschild: Here ya go: Lucky number #1600.

UziTech: 👍 you should commit this change and remove the (`+`) on `heading` and this should be good to go once we hear back from @davisjam

src/rules.js
calculuschild: Committed

src/rules.js
calculuschild: Committed.

calculuschild:
> Is this stable enough for me to look at the regexes for ReDoS?

Has there been any progress on this step? Is there anything I can help with?

UziTech:
@davisjam have you had time to look at these?

calculuschild:
High five! ✋

PR #1599, opened 2020-02-05
Cleanup now.json

styfle:
This PR removes unnecessary properties from `now.json`.


UziTech:
Do you know if there is a way to only deploy when the docs change instead of on every pull request?

I also created an issue on zeit now about this https://github.com/zeit/now/issues/3744 but I noticed you recently started working for Zeit (congrats :tada: by the way)

styfle:
I added`.nowignore`.

At some point, it might be nice to have the PR version use the version of marked from the PR rather than needing link to the raw github. This could be added as a build step.

PR #1600, opened 2020-02-10
Fix GFM Example 116

calculuschild:
**Marked version:** 8.0

**Markdown flavor:** CommonMark & GitHub Flavored Markdown

## Description

- Fixes GFM ([Example 116](https://github.github.com/gfm/#example-116)). Only code fences made from backticks [cannot have other backticks in the same line](https://github.github.com/gfm/#example-115). Fences with tildes [can have both tildes and backticks.](https://github.github.com/gfm/#example-116). This change allows the tilde case to work correctly.

Also adds tests for detection of these fences if they should or not break paragraphs.

Implementing code as discussed in https://github.com/markedjs/marked/pull/1598#discussion_r377252499

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/src/rules.js
+++ b/src/rules.js
@@ -10,7 +10,7 @@ const {
 const block = {
   newline: /^\n+/,
   code: /^( {4}[^\n]+\n*)+/,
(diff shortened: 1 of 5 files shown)


calculuschild:
Cool. Does anything else need to happen to merge this?

PR #1616, opened 2020-03-06
Release 0.8.1

UziTech:
# Release Notes

## Fixes

- Fix `marked --help` #1588
- Fix GFM Example 116 code fences #1600
- Send inline html to renderer #1602 (fixes #1601)
- Improve docs example for invoking highlight.js #1603
- Fix block-level elements breaking tables #1598 (fixes #1467)
- break nptables on block-level structures #1617

----

## Publisher

- [ ] `$ npm version` has been run.
- [ ] Release notes in [draft GitHub release](https://github.com/markedjs/marked/releases) are up to date
- [ ] Release notes include which flavors and versions of Markdown are supported by this release
- [ ] Committer checklist is complete.
- [ ] Merge PR.
- [ ] Publish GitHub release using `master` with correct version number.
- [ ] `$ npm publish` has been run.
- [ ] Create draft GitHub release to prepare next release.

Note: If merges to `master` occur after submitting this PR and before running `$ npm pubish` you should be able to

1. pull from `upstream/master` (`git pull upstream master`) into the branch holding this version,
2. run `$ npm run build` to regenerate the `min` file, and
3. commit and push the updated changes.

## Committer

In most cases, this should be someone different than the publisher.

- [x] Version in `package.json` has been updated (see [PUBLISHING.md](https://github.com/markedjs/marked/blob/master/docs/PUBLISHING.md)).
- [x] The `marked.min.js` has been updated; or,
- [ ] release does not change library.
- [x] CI is green (no forced merge required).

PR #1617, opened 2020-03-09
break nptables on block-level structures

UziTech:
**Marked version:** master

## Description

This PR copies #1598 to add breaking block-level rules to `nptable` (tables that don't start or end with a pipe)

e.g:
```markdown
# nptable

 abc | def
 --- | ---
  1  |  2
  3  |  4

# table

| abc | def |
| --- | --- |
|  1  |  2  |
|  3  |  4  |
```

I also removed allowing setext header from breaking a table since it doesn't break on github. We can change it later easily if github changes. see https://github.com/markedjs/marked/pull/1598#pullrequestreview-355648462

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR).

## Committer

- [ ] Merge PR


src/rules.js
```diff
@@ -93,22 +93,38 @@ block.normal = merge({}, block);
...
+  .replace('list', ' {0,3}(?:[*+-]|1[.)]) ') // only lists starting from 1 can interrupt
+  .replace('html', '</?(?:tag)(?: +|\\n|/?>)|<(?:script|pre|style|!--)')
+  .replace('tag', block._tag) // tables can be interrupted by type (6) html blocks
+  .getRegex();
+
 block.gfm.table = edit(block.gfm.table)
   .replace('hr', block.hr)
   .replace('heading', ' {0,3}#{1,6} ')
-  .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
+  // .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
```
styfle: I think we should use git log history instead of leaving a comment like this

src/rules.js
```diff
@@ -93,22 +93,38 @@ block.normal = merge({}, block);
...
   table: '^ *\\|(.+)\\n' // Header
     + ' *\\|?( *[-:]+[-| :]*)' // Align
-    + '(?:\\n((?:(?!^|>|\\n| |hr|heading|lheading|code|fences|list|html).*(?:\\n|$))*)\\n*|$)' // Cells
+    + '(?:\\n *((?:(?!\\n|hr|heading|lheading|blockquote|code|fences|list|html).*(?:\\n|$))*)\\n*|$)' // Cells
 });
 
+block.gfm.nptable = edit(block.gfm.nptable)
+  .replace('hr', block.hr)
+  .replace('heading', ' {0,3}#{1,6} ')
+  // .replace('lheading', '([^\\n]+)\\n {0,3}(=+|-+) *(?:\\n+|$)')
```
styfle: I think we should use git log history instead of leaving a comment like this

UziTech:
@joshbruce could you review this so we can release [v0.8.1](https://github.com/markedjs/marked/pull/1616)

joshbruce:
So glad to see all the tests.

PR #1622, opened 2020-03-20
Render html in heading

### LINKED ISSUE #1621: TypeError: this.renderer.html is not a function (Error in 0.8.1, OK in 0.8.0)
**Describe the bug**

Bug in **0.8.1**. In version 0.8.0 everything works fine. 
 
```bash
(node:17048) UnhandledPromiseRejectionWarning: TypeError: this.renderer.html is not a function
    at InlineLexer.output (C:\Users\XYZ\workspace\node_modules\marked\src\InlineLexer.js:85:30)
    at Parser.tok (C:\Users\XYZ\workspace\node_modules\marked\src\Parser.js:97:36)
    at Parser.parse (C:\Users\XYZ\workspace\node_modules\marked\src\Parser.js:47:19)
    at Function.parse (C:\Users\XYZ\workspace\node_modules\marked\src\Parser.js:30:19)
    at marked (C:\Users\XYZ\workspace\node_modules\marked\src\marked.js:103:19)
```

UziTech:
**Marked version:** 0.8.1

## Description

- Add html to TextRenderer for html in headings.
- Fix Slugger to remove html tags.

Fixes #1621

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/src/Slugger.js
+++ b/src/Slugger.js
@@ -13,6 +13,9 @@ module.exports = class Slugger {
     let slug = value
       .toLowerCase()
       .trim()
+      // remove html tags
+      .replace(/<!?\/?[\w-]+(?: .*)?\/?>/g, '')
+      // remove unwanted chars
       .replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,./:;<=>?@[\]^`{|}~]/g, '')
       .replace(/\s/g, '-');
 
--- a/src/Slugger.js
+++ b/src/Slugger.js
@@ -14,7 +14,7 @@ module.exports = class Slugger {
       .toLowerCase()
       .trim()
       // remove html tags
-      .replace(/<!?\/?[\w-]+(?: .*)?\/?>/g, '')
+      .replace(/<[!\/a-z].*?>/ig, '')
       // remove unwanted chars
       .replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,./:;<=>?@[\]^`{|}~]/g, '')
       .replace(/\s/g, '-');
--- a/src/TextRenderer.js
+++ b/src/TextRenderer.js
@@ -20,6 +20,10 @@ module.exports = class TextRenderer {
     return text;
   }
 
+  html(text) {
+    return text;
+  }
+
   text(text) {
     return text;
   }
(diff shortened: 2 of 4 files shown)


UziTech:
@davisjam could you check that `/<!?\/?[\w-]+(?: .*)?\/?>/g` isn't vulnerable to redos?

joshbruce:
Would like to hold off on my approval until @davisjam (or someone) can review for REDOS.

Gerrit0:
According to http://redos-checker.surge.sh/, it is vulnerable, which makes sense since the lookahead can also match characters in `[\w-]+`. I suggest `/<[!\/\w-].*?>/g`, which provides mostly similar results without being vulnerable.

That said, both of these options break rather easily... regex isn't a good html parser. The `<em>` test added implies that the contents within a tag should be preserved... but what should be done here?

```html
<a href="BREAK>" target="_blank">tag</a>
```

UziTech:
`?:` isn't a look ahead. It is a non-capturing group. And http://redos-checker.surge.sh says `<!?\/?[\w-]+(?: .*\/?>|\/?>)` is not vulnerable which is the same as `<!?\/?[\w-]+(?: .*)?\/?>`.

`safe-regex` which is used by http://redos-checker.surge.sh just checks the star height of the regex so it has a lot of false positives. It is designed so if it says it is NOT vulnerable than you know it is safe but if it says it is vulnerable it still might not be vulnerable.

as for the `<a href="BREAK>" target="_blank">tag</a>` example CommonMark/GFM spec says `'>'` must be percent encoded in valid markdown so it is ok to assume `'>'` is the end of a tag.

UziTech:
That being said your regex might be better since it is simpler. I would like to match github but I don't know what GitHub uses for it's heading ids.

UziTech:
I simplified the regex to `/<[!\/a-z].*>/ig` so it must start with a `!`, `/`, or a letter to be considered an html tag.

@joshbruce this regex is definitely not vulnerable.

joshbruce:
I’m finding myself being pulled to family and friends at the moment. Can we temporarily move to a single review model - as long as it doesn’t introduce a security vulnerability? Further, if it does, the only required review would be for the security piece?

joshbruce:
Not in a position to complete second approves flow. @styfle??

UziTech:
I got it.

PR #1627, opened 2020-03-25
Inline tokens

UziTech:
## Description

Right now the `Lexer` returns the block tokens and the `Parser` sends the text to the `InlineLexer` which sends the information to the `Renderer`. There is no way to get the inline tokens before sending them to the `Renderer`.

This PR changes the `Lexer` to return all tokens with block tokens that have tokens inside of them in a `tokens` property on the token and changes the `Parser` to act on all tokens instead of the text from the block tokens.

Each token that has inline or block tokens will have a `tokens` property that is an array of tokens.

The `list` token will have an `items` property which will be an array of list items. Each list item will have a `tokens` property with the tokens inside the item.

This PR also adds a `raw` property to the tokens so a user can get the markdown that created that token.

fixes #1626
fixes #361

## TODO:

- [x] Write unit tests for `Lexer`, and `Parser`
- [x] Make sure demo works
	- You can see the new tokens [here](https://markedjs-t4fq3vpiu.now.sh/demo/?outputType=lexer&version=pr%3A1627)
- [x] Update docs
- [x] Refactor to make benchmarks faster
	- About 5-10% faster than master
	- It looks like this sped the compiled es5 version (`lib/marked.js`) up even more.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression.
- [x] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] Draft GitHub release notes have been updated.
- [ ] CI is green (no forced merge required).
- [ ] Merge PR

PR #1632, opened 2020-03-31
Split Lexer into extendable functions

### LINKED ISSUE #1373: Extending with custom tags
I want to create custom "tags", like text alignment blocks, inline text color, etc.
Documentation is not very clear to me how to create totally new tags or tokens.

There are other examples?

calculuschild:
List, NPTable, and Table functions slightly rewritten to remove need to access "src" variable within the functions (i.e. no "Backtracking" steps)

**Marked version:**

8.2

## Description

- Fixes #1373 

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -54,342 +54,77 @@ module.exports = class Lexer {
    */
   token(src, top) {
     src = src.replace(/^ +$/gm, '');
-    let next,
-      loose,
-      cap,
-      bull,
-      b,
-      item,
-      listStart,
-      listItems,
-      t,
-      space,
-      i,
-      tag,
-      l,
-      isordered,
-      istask,
-      ischecked;
+    let cap;
 
     while (src) {
       // newline
       if (cap = this.rules.newline.exec(src)) {
-        src = src.substring(cap[0].length);
-        if (cap[0].length > 1) {
-          this.tokens.push({
-            type: 'space'
-          });
-        }
+        if (this.newline(cap)) { src = src.substring(cap[0].length); continue; }
       }
 
       // code
       if (cap = this.rules.code.exec(src)) {
-        const lastToken = this.tokens[this.tokens.length - 1];
-        src = src.substring(cap[0].length);
-        // An indented code block cannot interrupt a paragraph.
-        if (lastToken && lastToken.type === 'paragraph') {
-          lastToken.text += '\n' + cap[0].trimRight();
-        } else {
-          cap = cap[0].replace(/^ {4}/gm, '');
-          this.tokens.push({
-            type: 'code',
-            codeBlockStyle: 'indented',
-            text: !this.options.pedantic
-              ? rtrim(cap, '\n')
-              : cap
-          });
-        }
-        continue;
+        if (this.code(cap)) { src = src.substring(cap[0].length); continue; }
       }
 
       // fences
       if (cap = this.rules.fences.exec(src)) {
-        src = src.substring(cap[0].length);
-        this.tokens.push({
-          type: 'code',
-          lang: cap[2] ? cap[2].trim() : cap[2],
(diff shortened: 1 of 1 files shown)


calculuschild:
This is my initial attempt at splitting the Lexer into extendable functions for each Markdown feature. Passes all tests, no noticeable slowdown through benchmark (that I can see). Each function returns `true` if it successfully generated tokens, e.g., the `table` function might detect an invalid table with headers and delimiter rows with different cell counts, and just `continue` out without consuming from `src`. Hope that makes sense.

Every function is passed `cap` which contains the captured text from the regex in `rules.js`. Some functions (`list` and `blockquote`) also require `top` as they can recursively contain other elements. 

So, further questions:
1. Since some functions require `top`, should we pass it to every function to allow users full customization in case they want some weird custom nesting behavior (and for consistency)?
2. Should we also expose `src` to each function? I explicitly rewrote a couple sections so it would not be necessary, but perhaps some users would want to be able to view the current text being parsed? It should minimally impact performance as strings are passed by reference.
3. Should the entire regex rule check (`if (cap = this.rules.html.exec(src))` also just be moved to the functions as well?
4. The remaining 14 copies of `src = src.substring(cap[0].length);` just bug me. Is there a clean way to eliminate that repetition?

UziTech:
1. no it will build when the PR is merged.
2. maybe
3. maybe we should pass an object with all of these variables?
4. That might be better.
5. Probably not without introducing overhead. There are quite a few things we do to keep it as fast as possible

I am working on getting the InlineLexer to output tokens as well in PR #1627 maybe you want to build on that PR to allow the inline tokens (e.g strong, em ,link, etc.) to be extendable as well.

calculuschild:
> maybe you want to build on that PR to allow the inline tokens (e.g strong, em ,link, etc.) to be extendable as well.

@UziTech Is that PR pretty stable where it is? Do I just check out a new branch based on that PR?

UziTech:
I'm thinking the easiest way for users to be able to extend the lexers would be to create a `Tokenizer` class that holds those functions and each function is passed an object with those variables. The functions can return `false` to continue to the next function or edit the src and return a token to add to the lexer's `tokens` array.

The user could modify the `Tokenizer` and add the modified `Tokenizer` as an option the same way the `Renderer` is modified for the `Parser`

calculuschild:
Ok, so let me make sure I have this right:

1. Add a Tokenizer class into a new `Tokenizer.js` file?
2. The Tokenizer will hold functions that are all passed an object with `src` and `top`
    - *note:* if `rules.js` regex is checked within the function, no need to pass `cap`
3. Each function will do the following:
    1. Compare the current `src` with regex rules from `rules.js`
    2. If match, parse the matched text into tokens
    3. If tokens are successfully added to the Lexer's `tokens` array, alter `src` and return `true`
    4. If some error occurs or no token is generated, return `false`
4. `Lexer` will simply loop through each Tokenizer function in sequence until it receives a `true` to continue to the next iteration, ending when all of `src` is consumed. 

If this makes sense to you I can use #1627 as the starting point for this.

UziTech:
> If tokens are successfully added to the Lexer's `tokens` array, alter `src` and return `true`

I think it would be better if the function returns the token if it is successful and alters `src`. Then the Lexer can add the token to the array and continue the while loop.

It might be easier if we just collaborate on #1627 together than trying to make two PR's to marked. You can contribute to that PR by sending PRs to the [UziTech/marked/inline-tokens](https://github.com/UziTech/marked/tree/inline-tokens) branch

calculuschild:
> I think it would be better if the function returns the token if it is successful and alters `src`.

Gotcha. Misread your earlier post.

Ok. I can work on that branch.

calculuschild:
Superseded by #1637 .

Issue #545, opened 2015-02-16
Table of contents


mazing:
Is it possible to get a list of all headers in order to create a table of contents?

paxidently:
Just define custom renderer. Like this:

````js
var renderer = new marked.Renderer();

var toc = [];

renderer.heading = function(text, level, raw) {
    var anchor = this.options.headerPrefix + raw.toLowerCase().replace(/[^\w]+/g, '-');
    toc.push({
        anchor: anchor,
        level: level,
        text: text
    });
    return '<h'
        + level
        + ' id="'
        + anchor
        + '">'
        + text
        + '</h'
        + level
        + '>\n';
};

marked.setOptions({
    renderer: renderer
});
````

**Comment by anaran** [comment]
Worked great for me, thanks!

```
  // See
  // https://github.com/chjj/marked/issues/545#issuecomment-74505539
  var toc = [];
  var renderer = (function() {
    var renderer = new marked.Renderer();
    renderer.heading = function(text, level, raw) {
      var anchor = this.options.headerPrefix + raw.toLowerCase().replace(/[^\w]+/g, '-');
      toc.push({
        anchor: anchor,
        level: level,
        text: text
      });
      return '<h'
      + level
      + ' id="'
      + anchor
      + '">'
      + text
      + '</h'
      + level
      + '>\n'
      + '<a href="#table-of-contents">Table of Contents<a>\n';
    };
    return renderer;
  })();

  marked.setOptions({
    renderer: renderer,
    gfm: true,
    tables: true,
    breaks: false,
    pedantic: false,
    sanitize: true,
    smartLists: true,
    smartypants: false
  });

          var html = marked(request.response);
          var tocHTML = '<h1 id="table-of-contents">Table of Contents</h1>\n<ul>';
          toc.forEach(function (entry) {
            tocHTML += '<li><a href="#'+entry.anchor+'">'+entry.text+'<a></li>\n';
          });
          tocHTML += '</ul>\n';
          document.querySelector('#render_markdown').innerHTML = tocHTML + html;
```

**Comment by joshbruce** [comment]
Closing as out of scope based on current focus.

See #956

**Comment by yzfdjzwl** [comment]
Those answers above, only support English, do not support English.So I add some rules based on the example above:

```javascript
var anchor = this.options.headerPrefix + raw.toLowerCase().replace(/[^\w\\u4e00-\\u9fa5]]+/g, '-');
```

Hope it can help we Chinese.

joshbruce:
[...]

PR #1637, opened 2020-04-03
Tokenizer

UziTech:
## Description

This PR builds on #1627 to create a Tokenizer class that can be extended to influence the Lexer.

## TODO:

- [x] Write unit tests
- [x] Make benchmarks faster (currently same as master)
- [x] Update docs

## Contributor

## Committer

- [x] Merge PR


calculuschild:
Nice. Looks like you got it going. I think I'll go ahead and close #1632 then.

UziTech:
This should probably be cleaned up. Right now the Tokenizer functions get passed an object that has a `src` and `out` property that needs to be edited along with returning a token. It seems like a weird way to make the user edit the `src` and `out`.

calculuschild:
What does the `out` property do?

UziTech:
it provides the text for inline tokens

UziTech:
After some refactoring it looks like we could remove the `out` property if we dont care to have the text for inline tokens that can have other inline tokens inside of them (`strong`, `em`, and `del`).

calculuschild:
> It seems like a weird way to make the user edit the `src` and `out`.

In my attempt #1632 I also removed editing the `src` from the Token functions. I imagine the same could be done here, so `src` is only really needed to do the initial regex capture and then if a token is returned the `lexer` will update `src` appropriately.

UziTech:
Yes but I think the original regex capture is important to be able to extend/change. And how would we deal with backtracking? If we want to be 100% spec compliant I think we will need to do more of that.

calculuschild:
> I think the original regex capture is important to be able to extend/change.

Right, performing the regex capture makes sense, so we pass in `src` for that, but I don't think it's necessary to edit `src` in the tokenizer. Then again, isn't that why the `rules` is already extendable? 

> And how would we deal with backtracking?

In #1632 I found ways around the backtracking in the `list` token. If that's not enough, perhaps we can simply use the length of the "raw" text in the tokens and consume that much of `src`.

UziTech:
I like the idea of using raw length 👍. We will still need to pass the `src` to the tokenizer functions but they won't need to update it.

UziTech:
I think this is just about ready. A few things to point out:

1. The block text function is called `text` and the inline text function is called `inlineText`.
2. I left `smartypants` and `mangle` as public functions so the user could extend them if they want.
3. Some tokens (`em`, `strong`, `del`, `link`) don't have a `text` property since they have inline tokens. We could add a `text` property from the `tokens` property but I don't think it is necessary.
4. I am passing `src` and other variables by value instead of in an object but I didn't see any slow down in the benchmarks.

calculuschild:
> I am passing src and other variables by value instead of in an object but I didn't see any slow down in the benchmarks.

My understanding is objects, strings, and arrays all default to pass by reference in Javascript anyway, so really `top` is the only one that would make a difference in speed, no?

UziTech:
> My understanding is objects, strings, and arrays all default to pass by reference in Javascript anyway, so really top is the only one that would make a difference in speed, no?

Correct.

calculuschild:
Cool. This is awesome work!

UziTech:
The other thing I would like to do for extensibility is to allow multiple extensions to alter different parts of the renderer and tokenizer. For example if one extension only needs to alter `code` and another extension only affects `html` they won't override each other.

I am thinking of something like this interface:

```js
// marked_code_extension
module.exports = {
	tokenizer: {
		code(lexer, src, tokens, top) {
			// code tokenizer
		}
	}
	renderer: {
		code(code, infostring, escaped) {
			// code renderer
		}
	}
};
```

```js
// marked_html_extension
module.exports = {
	tokenizer: {
		html(lexer, src, tokens, top) {
			// html tokenizer
		}
	}
	renderer: {
		html(html) {
			// html renderer
		}
	}
};
```

```js
marked.use(require('marked_code_extension'));
marked.use(require('marked_html_extension'));

const html = marked(markdown);
// code and html extensions will both be used

```

calculuschild:
> I am thinking of something like this interface:

This would be handy. Would this maybe fit in a separate PR?

UziTech:
> This would be handy. Would this maybe fit in a separate PR?

Yes I would do a separate PR for this but I wanted to get the idea out there before I start updating the docs.

calculuschild:
Is there anything in particular blocking this?

UziTech:
I'd like to decouple the tokenizer and the lexer and make it easier to change. Hopefully not have to pass the lexer to each function. I will have more time to work on it next week.

UziTech:
I wasn't sure how much control we should give to the user. I left checking the rules to the user in case they want to use something other than a regex to check for a token.

If we move the rules check back to the lexer we could send more meaningful data to the Tokenizer functions. At that end of the spectrum the Tokenizer functions would just be there to construct the tokens and all other processing of the source markdown would be done by the Lexer. I didn't think that was very useful since they could just get the tokens that the lexer currently outputs and run that through a processor to do the same thing without the Tokenizer being needed.

The other end of the spectrum is to give them access to the markdown, rules, lexer, and all the other state that the lexer currently holds. That seems like too much control. At that point they might as well build their own lexer and just use marked to send the tokens to the parser.

@calculuschild @styfle Is this a good compromise or should the user have more or less control?

UziTech:
@joshbruce 

1. Lexer takes the markdown and sends it to the Tokenizer functions.
2. Tokenizer uses rules create tokens.
3. Parser takes tokens and sends them to the Renderer functions.
4. Renderer returns html for output.

calculuschild:
I think this is a good compromise. With this, users can
1. change the rules for capturing text (via extending `rules`) (*is this possible?*)
2. change what that captured text means (via extending `tokenizer')
3. change what output is generated (via extending 'renderer')

This seems like a pretty reasonable amount of control while keeping functionality compartmentalized.

UziTech:
> change the rules for capturing text (via extending rules) (is this possible?)

They are (not very easily) extendable by monkey patching them from the Lexer currently

```js
marked.Lexer.rules.block.html = // change default html rule
``` 

I would like to do a PR after this PR gets merged to make extending the rules, Renderer, Tokenizer much easier. something like:

```js
marked.use({
	rules: {
		block: {
			html: // change default html rule
		}
	}, 
	tokenizer: {
		html: // change html function in tokenizer
	}, 
	renderer: {
		html: // change html function in renderer
	}
});
```

That would allow extensions to just return that object so users could use marked with extensions like:

```js
marked.use(require("some_marked_extension"));
```

docs/USING_PRO.md
```diff
@@ -89,9 +89,93 @@ slugger.slug('foo-1') // foo-1-2
...
+      type: 'codespan',
+      raw: match[0],
+      text: match[1].trim()
+    };
+  }
+  return originalCodespan.apply(this, arguments);
+};
+
+// Run marked
+console.log(marked('$ latex code $', { tokenizer }));
```
styfle, line 119: Nice example!

docs/USING_PRO.md
```diff
@@ -89,9 +89,93 @@ slugger.slug('foo-1') // foo-1-2
 - image(*string* href, *string* title, *string* text)
 - text(*string* text)
 
+<h2 id="tokenizer">The tokenizer</h2>
+
+The tokenizer defines how to turn markdown text into tokens.
+
+**Example:** Overriding default `codespan` tokenizer to include latex.
```
styfle, line 96:
```suggestion
**Example:** Overriding default `codespan` tokenizer to include LaTeX.
```

docs/USING_PRO.md
styfle: Can we also achieve a dynamic TOC based on headings using a custom tokenizer?

UziTech: It seems like it would be easier to use the Renderer for that. [example](https://github.com/markedjs/marked/issues/545#issuecomment-74505539)

calculuschild:
[These lines](https://github.com/markedjs/marked/pull/1637/files#diff-0d98a8505a813bcc95beb15072d50725R199-R200) still mentioned rules as part of the Lexer. They are now only accessible from the Tokenizer, no? (Github is not letting me comment directly on those lines...)

UziTech:
Nice catch. yes the rules are set on the tokenizer based on the options and the Lexer has a static property `rules` where the user can access all of the rules.

docs/USING_PRO.md
styfle: Oh yeah. So then couldn't this LaTeX example also be achieved using the Renderer instead of Tokenizer?

UziTech: No because the dollar sign (`$`) isn't a valid code token starter normally so it would just be counted as text.

I'm sure the example code would fail on edge cases but it is more about showing how to extend the Tokenizer than create a robust LaTeX interpreter.

UziTech:
@calculuschild have you had a chance to review this PR. I would like your approval before merging it since you had the original idea for the Tokenizer. Does this provide the fix for your use case?

calculuschild:
@UziTech Yes, this should work just fine for what I need. Thank you.

Issue #1488, opened 2019-05-13
renderer.code includes space at beginning of each line of code


lucaschen:
Indented code blocks are rendering with a space at the beginning of each line. Consider the following code:

```md
This is some text.

1. This is a list element.

	```js
	const x = 5;
	const y = x + 5;
	```
```

This will render with a single space before both lines of code. The problem persists regardless of whether spaces or tabs are used for indentation.

I tried doing a log inside of `renderer.code` like so:

```js
const renderer = new marked.Renderer();

renderer.code = (code, language) => {
  console.log(code);
}
```

In `code`, I can see a space before each line.

UziTech:
This should be fixed in #1645

PR #1645, opened 2020-04-15
fix: renderer.code includes space at beginning of each line of code

Scrum:
## Description

- Fixes renderer.code includes space at beginning of each line of code, issue #1488    

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -3,7 +3,8 @@ const {
   rtrim,
   splitCells,
   escape,
-  findClosingBracket
+  findClosingBracket,
+  indentCodeCompensation
 } = require('./helpers.js');
 
 function outputLink(cap, link, raw) {
@@ -77,11 +78,14 @@ module.exports = class Tokenizer {
   fences(src) {
     const cap = this.rules.block.fences.exec(src);
     if (cap) {
+      const raw = cap[0];
+      const text = indentCodeCompensation(raw, cap[3] || '');
+
       return {
         type: 'code',
-        raw: cap[0],
+        raw,
         lang: cap[2] ? cap[2].trim() : cap[2],
-        text: cap[3] || ''
+        text
       };
     }
   }
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -3,8 +3,7 @@ const {
   rtrim,
   splitCells,
   escape,
-  findClosingBracket,
-  indentCodeCompensation
+  findClosingBracket
 } = require('./helpers.js');
 
 function outputLink(cap, link, raw) {
@@ -28,6 +27,38 @@ function outputLink(cap, link, raw) {
       title
     };
   }
+}
+
+function indentCodeCompensation(raw, text) {
+  if (raw === undefined) {
+    return text;
+  }
+
+  const matchIndentToCode = raw.match(/^(\s+)(?:```)/);
+
+  if (matchIndentToCode === null) {
+    return text;
+  }
+
+  const indentToCode = matchIndentToCode[matchIndentToCode.length - 1];
+
+  return text
+    .split('\n')
+    .map(node => {
+      const matchIndentInNode = node.match(/^\s+/);
+      if (matchIndentInNode === null) {
+        return node;
+      }
+
+      const [indentInNode] = matchIndentInNode;
+
+      if (indentInNode.length >= indentToCode.length) {
+        return node.slice(indentToCode.length);
+      }
+
+      return node;
+    })
+    .join('\n');
 }
 
 /**
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -40,7 +40,7 @@ function indentCodeCompensation(raw, text) {
     return text;
   }
 
-  const indentToCode = matchIndentToCode[matchIndentToCode.length - 1];
+  const indentToCode = matchIndentToCode[1];
 
   return text
     .split('\n')
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -30,10 +30,6 @@ function outputLink(cap, link, raw) {
 }
 
 function indentCodeCompensation(raw, text) {
-  if (raw === undefined) {
-    return text;
-  }
-
   const matchIndentToCode = raw.match(/^(\s+)(?:```)/);
 
   if (matchIndentToCode === null) {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -228,6 +228,38 @@ function checkSanitizeDeprecation(opt) {
   }
 }
 
+function indentCodeCompensation(raw, text) {
+  if (raw === undefined) {
+    return text;
+  }
+
+  const matchIndentToCode = raw.match(/^(\s+)(?:```)/);
+
+  if (matchIndentToCode === null) {
+    return text;
+  }
+
+  const [indentToCode] = matchIndentToCode.reverse();
+  
+  return text
+    .split('\n')
+    .map(node => {
+      const matchIndentInNode = node.match(/^\s+/);
+      if (matchIndentInNode === null) {
+        return node;
+      }
+
+      const [indentInNode] = matchIndentInNode;
+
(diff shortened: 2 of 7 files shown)


Scrum:
@UziTech @styfle LGTM

UziTech:
It looks like that test passes without the fix.

Our tests use [`@markedjs/html-differ`](https://github.com/markedjs/html-differ) to compare html and it is probably removing the space before checking. You can try writing a unit test to test the exact output.

Scrum:
@UziTech Hi
> You can try writing a unit test to test the exact output.

done

test/helpers/helpers.js
```diff
@@ -37,6 +38,21 @@ beforeEach(() => {
           return result;
         }
       };
-    }
+    },
+    toDeepEqual: () => ({
+      compare: async(spec, expected) => {
+        const result = {};
+        const actual = marked(spec.markdown, spec.options);
+        result.pass = assert.deepStrictEqual(expected, actual) === undefined;
```
UziTech, line 46: I feel like just using `===` would be good enough since we know both will always be strings.

```suggestion
        result.pass = expected === actual;
```

test/helpers/helpers.js
```diff
@@ -37,6 +38,21 @@ beforeEach(() => {
...
+
+        if (result.pass) {
+          result.message = `${spec.markdown}\n------\n\nExpected: Should Fail`;
+        } else {
+          const diff = await htmlDiffer.firstDiff(actual, expected);
```
UziTech, line 51: I don't think htmlDiffer will be useful here since it might not find a difference if it is just in whitespace.

we will probably need to loop over the strings to find the index where they change and show 30 chars on each side.

test/helpers/helpers.js
```diff
@@ -37,6 +38,21 @@ beforeEach(() => {
           return result;
         }
       };
-    }
+    },
+    toDeepEqual: () => ({
```
UziTech, line 42:
```suggestion
    toRenderExact: () => ({
```

test/specs/run-spec.js
```diff
@@ -16,17 +16,22 @@ function runSpecs(title, dir, showCompletionTable, options) {
...
             // eslint-disable-next-line no-eval
             spec.options.sanitizer = eval(spec.options.sanitizer);
           }
+
           (spec.only ? fit : (spec.skip ? xit : it))('should ' + passFail + example, async() => {
             const before = process.hrtime();
             if (spec.shouldFail) {
               await expectAsync(spec).not.toRender(spec.html);
+            } else if (spec.options.deepEqual) {
+              await expectAsync(spec).toDeepEqual(spec.html);
```
UziTech, line 34:
```suggestion
            } else if (spec.options.renderExact) {
              await expectAsync(spec).toRenderExact(spec.html);
```

test/helpers/helpers.js
Scrum: This is solved by using `assert.deepStrictEqual` in case of discrepancy, he shall provide comprehensive information on the discrepancy 
Screenshot 2020-04-17 at 09 25 14

I updated this method, removing all unnecessary

UziTech:
The text should be fixed when creating the token instead of when rendering. If you merge the latest changes you can change the `Tokenizer.fences` function to fix `token.text` before parsing.

https://github.com/markedjs/marked/blob/904c9743624a9eefbcab4e6183ba9cc84a9cbab0/src/Tokenizer.js#L77-L87

And you should inline the fix instead of adding a helper since the code is only going to run in one place. No need for the extra function call.

Scrum:
@UziTech I don’t quite understand, does this relate exactly to my task?

UziTech:
Yes, currently you are taking the text from the token and fixing it before sending it to the renderer.

I'm suggesting that you fix the text before it goes into the token so we can still just take the text from the token and send it to the renderer without changing it.

Scrum:
ok now understand

Scrum:
@UziTech done

UziTech:
If you merge the latest changes from master you will see a new file `src/Tokenizer.js`. This fix should go in that file.

Scrum:
@UziTech done

src/helpers.js
```diff
@@ -228,6 +228,38 @@ function checkSanitizeDeprecation(opt) {
   }
 }
 
+function indentCodeCompensation(raw, text) {
```
UziTech, line 231: Can you move this code into the `Tokenizer.fences` function? It doesn't make sense to have it in `helpers` since it is only ever going to be used in one place.

test/helpers/helpers.js
```diff
@@ -37,6 +38,16 @@ beforeEach(() => {
...
         }
       };
-    }
+    },
+    toRenderExact: () => ({
+      compare: async(spec, expected) => {
+        const result = {};
+        const actual = marked(spec.markdown, spec.options);
+
+        result.pass = assert.deepStrictEqual(expected, actual) === undefined;
```
UziTech, line 47:
```suggestion
        result.pass = assert.strictEqual(expected, actual) === undefined;
```

since `expected` and `actual` are always strings there is no need to do a deep equal.

test/helpers/helpers.js
```diff
@@ -1,5 +1,6 @@
 const marked = require('../../src/marked.js');
 const htmlDiffer = require('./html-differ.js');
+const assert = require('assert').strict;
```
UziTech, line 3:
```suggestion
const assert = require('assert');
```

`.strict` is unnecessary since we only use `strictEqual`

test/helpers/helpers.js
Scrum: done

test/helpers/helpers.js
Scrum: done

src/helpers.js
```diff
@@ -228,6 +228,38 @@ function checkSanitizeDeprecation(opt) {
...
+    return text;
+  }
+
+  const matchIndentToCode = raw.match(/^(\s+)(?:```)/);
+
+  if (matchIndentToCode === null) {
+    return text;
+  }
+
+  const [indentToCode] = matchIndentToCode.reverse();
```
styfle, line 242:
```suggestion
  const indentToCode = matchIndentToCode[matchIndentToCode.length - 1];
```

No need to iterate over each item in the array if you only need the last one.

Scrum:
@UziTech @styfle Thank you for making me wiser and the code cleaner

src/Tokenizer.js
```diff
@@ -29,6 +29,38 @@ function outputLink(cap, link, raw) {
...
+    return text;
+  }
+
+  const matchIndentToCode = raw.match(/^(\s+)(?:```)/);
+
+  if (matchIndentToCode === null) {
+    return text;
+  }
+
+  const indentToCode = matchIndentToCode[matchIndentToCode.length - 1];
```
UziTech, line 43: The spaces will always be at `[1]`

```suggestion
  const indentToCode = matchIndentToCode[1];
```

Scrum: it's more like using a magic number

calculuschild: Does the first match at [0] not pick this up as well? If so can you just use `raw.exec` instead of `raw.match`?

UziTech: `[0]` will return the spaces and ` ``` ` not just the spaces. We could use the whole match if we do a look ahead `(?=```)` instead of a non-matching group `(?:```)` but I don't know if one is better than the other.

src/Tokenizer.js
```diff
@@ -29,6 +29,38 @@ function outputLink(cap, link, raw) {
   }
 }
 
+function indentCodeCompensation(raw, text) {
+  if (raw === undefined) {
+    return text;
+  }
```
UziTech, line 35: `raw` should always be a string. I think if it is `undefined` it should throw an error on the `.match` call below instead of returning `text`. It would be easier to debug.
```suggestion
```

Scrum:
@styfle LGTM

PR #1646, opened 2020-04-19
marked.use

UziTech:
**Marked version:** master

## Description

Add function `marked.use` which takes an options object and merges it into marked options.

The `renderer` and `tokenizer` can be an object with functions that will be merged into the `renderer` and `tokenizer` respectively.

The `renderer` and `tokenizer` functions can return `false` to fallback to the previous function.

```js
const extension = {
	breaks: true,
	renderer: {
		html(html) {
			if (html.startsWith('<div')) {
				return html.replace(/</g, '&lt;').replace(/>/g, '&gt;');
			}
			return false;
		}
	}
};
marked.use(extension);
const html = marked(`
<div>
html renderer will use extension
</div>

<span>will fallback to regular html renderer</span>

options from extension
will be used.
`);
console.log(html);
/* will log:
&lt;div&gt;
html renderer will use extension
&lt;/div&gt;

<span>will fallback to regular html renderer</span>

<p>options from extension<br>will be used.</p>

*/
```

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -112,7 +112,7 @@ const tokenizer = {
       };
     }
 
-    // return false to use original codespan renderer
+    // return false to use original codespan tokenizer
     return false;
   }
 };
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -2,6 +2,16 @@
 
 To champion the single-responsibility and open/closed principles, we have tried to make it relatively painless to extend marked. If you are looking to add custom functionality, this is the place to start.
 
+<h2 id="use">marked.use()</h2>
+
+`marked.use(options)` is the recommended way to extend marked. The options object can contain any [option](#/USING_ADVANCED.md#options) available in marked.
+
+The `renderer` and `tokenizer` options can be an object with functions that will be merged into the `renderer` and `tokenizer` respectively.
+
+The `renderer` and `tokenizer` functions can return false to fallback to the previous function.
+
+All other options will overwrite previously set options.
+
 <h2 id="renderer">The renderer</h2>
 
 The renderer defines the output of the parser.
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -12,24 +12,25 @@ The renderer defines the output of the parser.
 // Create reference instance
 const marked = require('marked');
 
-// Get reference
-const renderer = new marked.Renderer();
-
 // Override function
-renderer.heading = function(text, level) {
-  const escapedText = text.toLowerCase().replace(/[^\w]+/g, '-');
+const renderer = {
+  heading(text, level) {
+    const escapedText = text.toLowerCase().replace(/[^\w]+/g, '-');
 
-  return `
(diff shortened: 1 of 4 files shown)


UziTech:
This can be used to extend any option in marked. The renderer and tokenizer will be merged but all other options will overwrite previous options.

UziTech:
@davisjam I added a section explaining `marked.use`

UziTech:
This PR doesn't change current ways of extending marked with `marked.setOptions(options)` or adding options as a second parameter to `marked(md, options)`.

It just adds a new method (`marked.use(options)`) that will merge the `tokenizer` and `renderer` instead of overriding them completely. This method should be recommended over other ways of changing options and maybe in the future we can change other ways to merging them as well.

UziTech:
The idea is to create extensions in the form

```js
// some-marked-extension
module.exports = {
	renderer: {...},
	tokenizer: {...}
};
```

```js
// some-other-marked-extension
module.exports = {
	renderer: {...},
	tokenizer: {...}
};
```

that can be used by users without overriding eachother

```js
marked.use(require('some-marked-extension'));
marked.use(require('some-other-marked-extension'));
...
```

src/marked.js
```diff
@@ -127,6 +127,43 @@ marked.getDefaults = getDefaults;
...
+ * Use Extension
+ */
+
+marked.use = function(extension) {
+  const opts = merge({}, extension);
+  if (extension.renderer) {
+    const renderer = marked.defaults.renderer || new Renderer();
+    for (const prop in extension.renderer) {
+      const prevRenderer = renderer[prop];
+      renderer[prop] = (...args) => {
```
styfle, line 140: I can't remember if we we supporting IE 11 anymore or which version of Node, but this syntax might be a breaking change.

docs/USING_PRO.md
```diff
@@ -99,30 +110,34 @@ The tokenizer defines how to turn markdown text into tokens.
...
+    if (match) {
+      return {
+        type: 'codespan',
+        raw: match[0],
+        text: match[1].trim()
+      };
+    }
+
+    // return false to use original codespan tokenizer
+    return false;
```
styfle, line 126: What is the difference between returning `false` vs `undefined`?

UziTech: `undefined` is a valid return for the tokenizer basically saying skip this token for the current `src`. I think `false` is a good thing to check because it is a way for the extension author to explicitly let marked know it wants it to fall back to the previous function instead of skipping the token.

We could change it to `true` or something else but I don't think it should always fall back with `undefined`.

src/marked.js
UziTech: this source will be compiled by babel to es5 in `lib/marked.js` which is currently used to support ie11.

I think we are just supporting node `lts` and `latest` which should support this syntax.

docs/USING_PRO.md
styfle: That answers my question, thanks!

PR #1652, opened 2020-04-24
Codespan newline

### LINKED ISSUE #1651: v1.0.0 does not respect multiline code spans in list items
v1.0.0:

```console
$ echo $'- `code\ncode`' | bin/marked
<ul>
<li>`code
code`</li>
</ul>
```

v0.8.2:

```console
$ echo $'- `code\ncode`' | bin/marked
<ul>
<li><code>code
code</code></li>
</ul>
```

This regression was introduced in 2806c4913c4e26b811f58eafbe73a17a9500d6d4.

### LINKED ISSUE #1655: v1.0.0 newlines inside lists
**Describe the bug**
newlines no longer work inside lists; this seems to be the case regardless if you use break:true or add the double-whitespace

**To Reproduce**
https://marked.js.org/demo/?text=-%20a%0A%20%20a&options=%7B%22breaks%22%3Atrue%2C%22gfm%22%3Atrue%7D&version=master

**Expected behavior**
* a
  a

UziTech:
**Marked version:** v1.0.0

## Description

- Code spans with a newline should turn the newline into a space according to the [commonmark spec](https://spec.commonmark.org/0.29/#example-335).

- Code spans only remove one space from each end if the text begins and ends with white-space

- fixes any list items with more than one line of text.

- Fixes #1651
- Fixes #1655
## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -112,7 +112,7 @@ module.exports = class Lexer {
    */
   blockTokens(src, tokens = [], top = true) {
     src = src.replace(/^ +$/gm, '');
-    let token, i, l;
+    let token, i, l, lastToken;
 
     while (src) {
       // newline
@@ -127,7 +127,13 @@ module.exports = class Lexer {
       // code
       if (token = this.tokenizer.code(src, tokens)) {
         src = src.substring(token.raw.length);
-        tokens.push(token);
+        if (token.type) {
+          tokens.push(token);
+        } else {
+          lastToken = tokens[tokens.length - 1];
+          lastToken.raw += '\n' + token.raw;
+          lastToken.text += '\n' + token.text;
+        }
         continue;
       }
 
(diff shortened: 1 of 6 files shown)


src/Lexer.js
```diff
@@ -219,9 +225,15 @@ module.exports = class Lexer {
       }
 
       // text
-      if (token = this.tokenizer.text(src)) {
+      if (token = this.tokenizer.text(src, tokens)) {
         src = src.substring(token.raw.length);
-        tokens.push(token);
+        if (token.type) {
```
davisjam, line 230: DRY? Same code added in two places?

UziTech: That code is not really related so they could change independently.

UziTech: I don't think we should combine that code into a function since it could change and need to be moved back anyway. This is one reason the best practice for DRY is to only combine code that needs to change together and only if it is in three or more places. [Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

Issue #1358, opened 2018-10-17
*"Yo"* not outputted correctly


cloudkite:
**Describe the bug**
Emphasize not working when text contains double quotes

**To Reproduce**

[marked demo](https://marked.js.org/demo/?text=*%22Yo%22*)
[commonmark demo](https://spec.commonmark.org/dingus/?text=*%22Yo%22*)

styfle:
Hi Jonas, thanks for the bug report!

This is indeed a bug in 0.5.1 and also in the latest code on master.

Would you like to submit a PR to fix this?

dialex:
It's been more than a year, and this is still an issue (as reported in https://github.com/docsifyjs/docsify/issues/976 and https://github.com/docsifyjs/docsify/issues/787). Is anyone fixing it? Please prioritise it 🙏

UziTech:
@dialex if you want to submit a PR I would be happy to review it 😁

dialex:
I would if I could, I don't know enough of JS or MarkedJS to be of any help.
I'll ask Santa for Christmas 🎅😁 

**Workaround**: use `_"Yo"_` instead of `*"Yo"*`

UziTech:
This was fixed in #1686 released in v1.1.1

PR #1654, opened 2020-04-24
Fix/em yo

Scrum:
## Description

- Fixes *"Yo"* not outputted correctly, issue #1358

## Contributor

## Committer

- [ ] Merge PR


test/specs/gfm/commonmark.0.29.json
```diff
@@ -2850,7 +2850,7 @@
   },
   {
     "markdown": "a*\"foo\"*\n",
-    "html": "<p>a*&quot;foo&quot;*</p>\n",
+    "html": "<p>a<em>&quot;foo&quot;</em></p>\n",
```
UziTech, line 2853: according to the [commonmark spec](https://spec.commonmark.org/0.29/#example-352) that should not change.

UziTech: These common mark tests are derived from the spec so they should not be changed unless a new version of the spec comes out.

If a spec is marked with `shouldFail: true` that means it is failing and needs to be fixed in marked.

Scrum: marked as shouldFail

UziTech:
Seems like this is moving backwards in spec compliance. Could you find a way to fix `"yo"` while not failing spec tests that were passing?

Scrum:
I don't really understand

1. https://spec.commonmark.org/dingus/?text=*%22Yo%22* here it is clearly visible that he is working out correctly, I don't know where the error is

2. I don't really understand the problem that needs to be solved so that the same value is handled differently

UziTech:
The problem is that <https://spec.commonmark.org/dingus/?text=a*%22Yo%22*> is not working correctly with this PR but it is working without this PR.

Scrum:
> The problem is that https://spec.commonmark.org/dingus/?text=a*%22Yo%22* is not working correctly with this PR but it is working without this PR.

Screenshot 2020-04-29 at 09 08 58

> Can you find a way to make *"Yo"* = <em>"Yo"</em> work as well as a*"Yo"* = a*"Yo"*?

To do this, just return the rule to the state of the master, but then this PR has no value

UziTech:
> To do this, just return the rule to the state of the master, but then this PR has no value

`*"Yo"* = <em>"Yo"</em>` isn't working in master

UziTech:
`marked('*"Yo"*')` is supposed to output `<em>"Yo"</em>` which works with this PR but in master it outputs `*"Yo"*`

`marked('a*"Yo"*')` is supposed to output `a*"Yo"*` which works in master but not with this PR.

Can you update this PR to make both work with this PR?

Scrum:
no since it is one and the same `marked('*"Yo"*')` `marked('a*"Yo"*')` you see the difference here if you miss the symbol `a` ?

UziTech:
It is definitely possible. You will have to change more than just the `em` regex to make it work.

I don't think this PR should be merged if it could break current spec compliant markdown.

PR #1664, opened 2020-05-04
add walkTokens and fix highlight async

### LINKED ISSUE #1656: highlight callback is undefined
Latest version also prev. 0.8.x, when i run 
```ts
marked.setOptions({
            highlight: (code, language, callback) => {
                hljs.highlightAuto(code, [language])
                    .pipe(take(1))
                    .subscribe((highlight) => {
                        callback(null, highlight.value);
                    }, error => callback(error));
            }
        });
```
callback is refered as undefined

UziTech:
**Marked version:** 1.0.0

## Description

Using an async highlighting function was broken in v1.0.0 when we moved block tokens in blockquote and lists from the top level of the tokens array to a `tokens` property on the blockquote or listitem token.

This PR adds a public method to easily iterate over all tokens and send each token to a function specified by the user (`marked.walkTokens(tokens, callback)`) and uses that function to highlight code blocks asynchronously.

This also adds a `walkTokens` option so extensions can use it to run a function for each token.

**Example:** Overriding heading tokens to start at h2.

```js

// Override function
const walkTokens = (token) => {
  if (token.type === 'heading') {
    token.depth += 1;
  }
};

marked.use({ walkTokens });

// Run marked
console.log(marked('# heading 2\n\n## heading 3'));
```

**Output:**

```html
<h2 id="heading-2">heading 2</h2>
<h3 id="heading-3">heading 3</h3>
```

- Fixes #1656 

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/src/marked.js
+++ b/src/marked.js
@@ -179,47 +179,31 @@ marked.use = function(extension) {
 };
 
 /**
- * Iterate over every token
+ * Run callback for every token
  */
 
 marked.walkTokens = function(tokens, callback) {
-  let ret;
   for (const token of tokens) {
-    ret = callback(token);
-    if (ret === false) {
-      return false;
-    }
+    callback(token);
     switch (token.type) {
       case 'table': {
         for (const cell of token.tokens.header) {
-          ret = marked.walkTokens(cell, callback);
-          if (ret === false) {
-            return false;
-          }
+          marked.walkTokens(cell, callback);
         }
         for (const row of token.tokens.cells) {
           for (const cell of row) {
-            ret = marked.walkTokens(cell, callback);
-            if (ret === false) {
-              return false;
-            }
+            marked.walkTokens(cell, callback);
           }
         }
         break;
       }
       case 'list': {
-        ret = marked.walkTokens(token.items, callback);
-        if (ret === false) {
-          return false;
-        }
+        marked.walkTokens(token.items, callback);
         break;
       }
       default: {
         if (token.tokens) {
-          ret = marked.walkTokens(token.tokens, callback);
-          if (ret === false) {
-            return false;
-          }
+          marked.walkTokens(token.tokens, callback);
         }
       }
     }
--- a/src/marked.js
+++ b/src/marked.js
@@ -28,18 +28,17 @@ function marked(src, opt, callback) {
       + Object.prototype.toString.call(src) + ', string expected');
   }
 
-  if (callback || typeof opt === 'function') {
-    if (!callback) {
-      callback = opt;
-      opt = null;
-    }
+  if (typeof opt === 'function') {
+    callback = opt;
+    opt = null;
+  }
 
-    opt = merge({}, marked.defaults, opt || {});
-    checkSanitizeDeprecation(opt);
+  opt = merge({}, marked.defaults, opt || {});
+  checkSanitizeDeprecation(opt);
+
+  if (callback) {
     const highlight = opt.highlight;
-    let tokens,
-      pending,
-      i = 0;
+    let tokens;
 
     try {
       tokens = Lexer.lex(src, opt);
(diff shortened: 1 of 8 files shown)


src/marked.js
```diff
@@ -164,6 +154,50 @@ marked.use = function(extension) {
   marked.setOptions(opts);
 };
 
+/**
+ * Iterate over every token
+ */
+
+marked.iterateTokens = function(tokens, callback) {
```
UziTech, line 161: I'm not sure what the best name for this function should be. A few names that I thought would be good:

- iterateTokens
- mapTokens (this might be confusing because we don't update the tokens based on the return value)
- updateTokens

/cc @styfle @joshbruce @davisjam

src/marked.js
```diff
@@ -76,34 +70,30 @@ function marked(src, opt, callback) {
...
-    if (!pending) return done();
+    if (!tokens.length) return done();
 
-    for (; i < tokens.length; i++) {
-      (function(token) {
-        if (token.type !== 'code') {
-          return --pending || done();
-        }
-        return highlight(token.text, token.lang, function(err, code) {
-          if (err) return done(err);
```
UziTech, line 87: I realized the callback sent to marked could be called multiple times because of this line. I'm not sure if that was intentional or a bug but I left it in case anyone relies on it being called for multiple highlighting errors.

src/marked.js
```diff
@@ -164,6 +154,50 @@ marked.use = function(extension) {
...
+ * Iterate over every token
+ */
+
+marked.iterateTokens = function(tokens, callback) {
+  let ret;
+  for (const token of tokens) {
+    ret = callback(token);
+    if (ret === false) {
+      return false;
+    }
```
UziTech, line 167: if the callback function returns `false` when iterating over the tokens the iterating will stop.

davisjam: Is this a breaking change? AFAICT it would previously touch every token, and just eat failures?

src/marked.js
davisjam: iterateTokens seems fine

src/marked.js
```diff
@@ -164,6 +154,50 @@ marked.use = function(extension) {
   marked.setOptions(opts);
 };
 
+/**
+ * Iterate over every token
```
davisjam, line 158: Possibly worth documenting a few things here:

1. That this is recursive (could be a source of weird bugs on huge documents).
2. That the visit order is a pre-order DFS.
3. That if callback returns false, the iteration ceases.

src/marked.js
UziTech: The async highlighting still iterates over every token since it doesn't ever return `false`. I just added this to the new `iterateTokens` function so there is a way to stop if the user wants to.

This behavior is similar to other library functions such as [jquery.each](https://api.jquery.com/each/).

src/marked.js
styfle: I vote no for `mapTokens` because the return type is not an array of tokens.

I think `iterateTokens` is okay but maybe something like `walkTokens` would be better because it implies walking the tree recursively. https://en.wikipedia.org/wiki/Tree_traversal

davisjam: Ah, I like `walkTokens`.

test/unit/marked-spec.js
```diff
@@ -141,6 +153,29 @@ describe('use extension', () => {
...
+    const extension2 = {
+      walkTokens(token) {
+        walkedOnce++;
+        token.walkedOnce = true;
+      }
+    };
+    marked.use(extension1);
+    marked.use(extension2);
+    marked('text');
+    expect(walkedOnce).toBe(2);
```
styfle, line 175: walkedOnce is 2? Maybe it should be renamed?

UziTech: That variable counts how many tokens were walked at least once

davisjam: Perhaps `nWalkedOnce` ?

PR #1686, opened 2020-05-21
Follow GFM spec on EM and STRONG delimiters

calculuschild:
Starting toward better adherence to the GFM spec on Emphasis, specifically Left-flanking-delimiter-runs.

**Marked version: 1.1.0**

## Description

-  Fixes Examples:
(em) 341, 367, 368, 371, 372, 379, 390, 406, 417, 441, 444
(strong) 391, 397, 399, 400, 401, 431, 443, 471, 475, 476, 479, 480

## What was attempted

Applying the GFM spec for Left-flanking-delimiter-runs and right-flanking-delimiter runs more accurately for EM tags and STRONG tags.

## Contributor

## Committer

- [x] Merge PR


UziTech:
I don't have a lot of time to look at this right now but it could be a conflict with lists since `*` could be the start of a list as well as em and strong.

calculuschild:
OK, so I figured out what the issue was. I now have it working quite well except for two test cases I could use some help on:

1) `Original strong_and_em_together`
This test passes fine normally, but when run with the 'pedantic' tag it breaks. I'm not totally sure what the pedantic tag is actually doing so I would like some feedback.

2) `CommonMark Links example 519`
I have a pattern in there to skip over any `\[ stuff \]` to avoid conflict with links. This last example breaks on things between brackets that aren't actually links. I'm not sure how to handle this because  it could be a reflink in certain cases.

Also note, I removed a line from the `New em_2char` test that was invalid according to the [spec](https://github.github.com/gfm/#example-452). I don't handle it properly here either, but my code isn't "breaking" it because it wasn't correct to begin with.

UziTech:
The `pedantic` option uses the [pedantic rules](https://github.com/markedjs/marked/blob/master/src/rules.js#L130) to tokenize the markdown.

Basically making marked use the [original spec](https://daringfireball.net/projects/markdown/syntax) instead of CommonMark.

calculuschild:
OK gotcha. My change to the em tokenizer confused it. I can fix this pretty easily.

Any insight on detecting whether or not a set of square brackets is a reflink or not?

calculuschild:
Can anyone help me out with this?

calculuschild:
Tada!! My solution was to inject known reflink labels into the em rules right after the **block** sequence is finished. This way the em rule can properly skip over any links that might contain `*`, but still allow emphasis inside of `[...]` if it is not a link.

I'm hoping this tweaking of the lexer is alright. I'm not sure if there's some way people could inject malicious regex this way by giving their link labels some weird names, but I assume it would just require some further character escaping on the label names before injection.

If this looks good, I can move forward with the `_` variant of the em tags as well as `\*\*strong\*\*` which uses a lot of the same logic.

calculuschild:
Underscore em rules added. Fixes 3 more examples (371, 372, 406)

I would love some feedback on this PR!

test/specs/commonmark/commonmark.0.29.json
```diff
@@ -2766,7 +2766,7 @@
     "start_line": 6003,
     "end_line": 6007,
     "section": "Code spans",
-    "shouldFail": true
+    "shouldFail": false
```
UziTech, line 2769: instead of setting `shouldFail` to `false` could you just remove the line?

src/Lexer.js
```diff
@@ -267,7 +274,7 @@ module.exports = class Lexer {
         case 'text':
         case 'heading': {
           token.tokens = [];
-          this.inlineTokens(token.text, token.tokens);
+          this.inlineTokens(token.text, token.tokens, undefined, undefined);
```
UziTech, line 277: Isn't passing `undefined` is the same thing as not passing anything?

src/Lexer.js
```diff
@@ -102,6 +103,12 @@ module.exports = class Lexer {
 
     this.blockTokens(src, this.tokens, true);
 
+    // Insert known reflinks into em rules to properly skip over them
+    const rep = Object.keys(this.tokens.links).join('|').replace(/\*/g, '\\*');
```
UziTech, line 107: Do we need to escape the links incase they have a regexp character (like `|`)?

I feel like a better way to do this would be to change the em rule to over collect then check for reflinks in the tokenizer.

something like:
1. `inline.em` collects `"*em [ref*link]*"`
2. em tokenizer finds that `"ref*link"` is not in `tokens.links`
3. reduce collection to `"*em [ref*"`

I'm not sure which would be faster but it would be less prone to ReDOS and easier to prevent errors based on reflink text.

UziTech: We do something similar with links to allow nesting an unlimited number of matched parentheses.

brainchild0:
Is support for nested emphasis of the same type included in this change set (e.g. `****superstrong****` going to `<strong><strong>superstrong</strong></strong>`, or `*emphasis *nests* inside*` going to `<em>emphasis <em>nests</em> inside</em>`)? 

CommomMark gives rules for resolving arbitrary levels of delimiter  runs.

src/Lexer.js
calculuschild: I will see if I can get this to work. It does look like a "cleaner" way to do it.

calculuschild:
> Is support for nested emphasis of the same type included in this change

Both of those cases work as you have written, at least on my local copy.

src/Lexer.js
calculuschild: @UziTech Ah, so I have this working on the example you wrote above, but issues arise in cases like this:
`*theres [not*a*link] here*`
which should render to `<em>theres [not<em>a</em>link] here</em>` because we still need to match the first `*` to the final `*`. So essentially we need to run the em regex again over the same section of text but this time not skipping over the `[ ]` when looking for `*`. This requires a whole separate em regex. And *then*, if it happens to be a line with multiple brackets and some are links and some aren't, it again stops working.
For example `*theres [not*a*link] here but [this*is*alink] here*` should become
`<em>theres [not<em>a</em>link] here but <a href="url">this<em>is</em>alink</a> here</em>`

calculuschild: I guess there's also the possibility of an internal asterisk requiring us to *expand* the capture instead of reducing, which will also not work with overcollection. For example:

`*Hello [my *name] is* Joe*`
Which should render to `<em>Hello [my <em>name] is</em> Joe</em>`. The initial capture would skip over the brackets and just get us `*Hello [my *name] is*`, but then by opening up the bracket and finding it is not a link, we now have a mismatched asterisk.

UziTech: My example was just to illustrate that the `inline.em` regex doesn't have to be the only thing we use to determine the end of the `em` token. We can choose how much we "over collect" for example the link regex grabs everything that is not white-space for the href and hands it to the tokenizer to determine the actual end of the href.

The `em` regex could just grab everything up until the last `*`, or the end of the string if no `*` exists past the first character, and let other regexes or functions inside the tokenizer decide where the end should actually be. One thing we will want to watch out for though are slow functions. regex search can be extremely fast but regular expressions can't parse everything (unlimited nesting for example).

UziTech: you can look at commonmarks recommended parsing of emphasis for an idea of what the tokenizer might need to do.

https://spec.commonmark.org/0.29/#-process-emphasis-

calculuschild: Right, I hear what you're saying. But the point I'm stuck on is that from what I can see, we still need to detect a pattern that considers the entirety of the sequence, intelligently skipping over or entering into bracketed spaces as needed, all at once. Individually looking over bracketed blocks and checking for known links is simple enough, but then we still need to look at the text as a whole, since skipping one asterisk and not another will propagate changes to the pairing of `em` spans all the way to the end.

I will keep thinking on this though.

src/Lexer.js
calculuschild: Done.

test/specs/commonmark/commonmark.0.29.json
calculuschild: Done.

src/Lexer.js
calculuschild: Moved the logic into Tokenizer and just overwrite any found reflinks with the string `[aaaaa]` of equal length to hide any em delimiters inside. Then run the em regex over the new string; all remaining `*` and `_` are valid, so we get a good result. Then just use the positions of this capture to slice the original string to restore the reflinks.

Hope this is more what you had in mind. :)

calculuschild:
Added logic to the `strong` delimiter now. Fixes several more examples.

test/specs/gfm/commonmark.0.29.json
```diff
@@ -2976,16 +2975,15 @@
     "start_line": 6455,
     "end_line": 6459,
     "section": "Emphasis and strong emphasis",
-    "shouldFail": true
+    "shouldFail":false
```
UziTech, line 2978: remove `"shouldFail": false` line

calculuschild: Done.

src/rules.js
```diff
@@ -168,19 +168,35 @@ const inline = {
   link: /^!?\[(label)\]\(\s*(href)(?:\s+(title))?\s*\)/,
   reflink: /^!?\[(label)\]\[(?!\s*\])((?:\\[\[\]]?|[^\[\]\\])+)\]/,
   nolink: /^!?\[(?!\s*\])((?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]])*)\](?:\[\])?/,
-  strong: /^__([^\s_])__(?!_)|^\*\*([^\s*])\*\*(?!\*)|^__([^\s][\s\S]*?[^\s])__(?!_)|^\*\*([^\s][\s\S]*?[^\s])\*\*(?!\*)/,
-  em: /^_([^\s_])_(?!_)|^_([^\s_<][\s\S]*?[^\s_])_(?!_|[^\s,punctuation])|^_([^\s_<][\s\S]*?[^\s])_(?!_|[^\s,punctuation])|^\*([^\s*<\[])\*(?!\*)|^\*([^\s<"][\s\S]*?[^\s\[\*])\*(?![\]`punctuation])|^\*([^\s*"<\[][\s\S]*[^\s])\*(?!\*)/,
+  preStrong: /^(?:\*\*)|(?:__)/,
```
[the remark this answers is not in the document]
calculuschild: Done.

src/Tokenizer.js
```diff
@@ -489,25 +489,64 @@ module.exports = class Tokenizer {
...
+      if (links) {
+        links = Object.keys(links);
+        const reg = /(?:\[.*?\]\[.*?\])|(?:\[.*?\](?!\())/g;
+        let match;
+        while ((match = reg.exec(text)) != null) {
+          if (links.includes(match[0].slice(match[0].lastIndexOf('[') + 1, -1))) {
+            text = text.slice(0, match.index) + '[' + 'a'.repeat(match[0].length - 2) + ']' + text.slice(reg.lastIndex);
+          }
+        }
+      }
```
UziTech, line 506: Masking every reflink every time we use the `em` or `strong` tokenizers seems like a very brute force way to do it.

Is there a way we could instead do a search for a `*` or `_` and then check if it is in a reflink?

UziTech: Instead of having an `inline.em` rule we could have an `inline.beginEm` and `inline.endEm` rules.

1. check `beginEm` and prevChar
2. search for next `endEm`
3. if current end is part of reflink goto 2

calculuschild: > 3. if current end is part of reflink goto 2

The issue with this is that we can't simply check the position of the ending `*`. The pattern must correctly skip or not skip over all reflinks leading up to that point.

Preprocessing the string to handle the reflinks might seem "brute force", but it also nullifies all of them in a single step, rather than iteratively checking each individual reflink for the presence of asterisks or vice-versa, which is a multi-step process that would have to be repeated for each em step anyway.

UziTech: > The issue with this is that we can't simply check the position of the ending *. The pattern must correctly skip or not skip over all reflinks leading up to that point.

That is what the `goto 2` is for. After finding an end that satisfies the other requirements we should check if it is part of a reflink. If it is then we should check for the next end and repeat until we find an end that is not part of a reflink.

It seems to be fine to mask the reflinks for the tests and the benchmarks but if someone has a lot of `em`s and a lot of reflinks it is On^2 and doing a lot of delicate work.

Also, I haven't had time to check, but it is fairly likely that the `inline.em` regex is vulnerable to redos by checking the beginning and end at the same time.

calculuschild: I think we are having some miscommunication issues. Is there a possibility of chatting live somewhere to hash out these details so we are on the same page?

UziTech: @calculuschild Sure how would you like to chat? my email is Tony@Brix.ninja if you want to send me something private.

calculuschild:
Ok @UziTech, how about this? Moved the masking step to the Lexer at the beginning of the `inlineTokens` function so each inline string is only masked once. This is as high up in the hierarchy as I can go to avoid redundant masking, right?

Also, the filter you applied to the links `links = Object.keys(links).filter(l => l.match(/[*_]/));` was making it fail in the commonmark example 530 case, where the `*` is in the link text rather than the label like this, because `ref` was getting filtered out:

```
*[foo*][ref]

[ref]: /uri\n
```

UziTech:
@davisjam any chance you would have time to check these regexs for vulnerabilities?

There are the big ones that changed:

```
inline.reflinkSearch: /!?\[((?:\[(?:\\.|[^\[\]\\])*\]|\\.|`[^`]*`|[^\[\]\\`])*?)\]\[(?!\s*\])((?:\\[\[\]]?|[^\[\]\\])+)\]|!?\[(?!\s*\])((?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]])*)\](?:\[\])?(?!\()/g

inline.strong: /^(?:(\*\*(?=[*!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~]))|\*\*)(?![\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)\*){2})+?)(?:(?<![!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])\*\*(?!\*)|(?<=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~])\*\*(?!\*)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s]|$)))|^__(?![\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)_){2})+?)(?:(?<![\s])__(?!_)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])|$))/

inline.em: /^(?:(\*(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~]))|\*)(?![*\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)\*){2})*?)(?:(?<![!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])\*(?!\*)|(?<=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~])\*(?!\*)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s]|$)))|^_(?![_\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)_){2})*?)(?:(?<![\s])_(?!_)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])|$))/

inline.pedantic.em: /^()\*(?=\S)([\s\S]*?\S)\*(?!\*)|^_(?=\S)([\s\S]*?\S)_(?!_)/
```

UziTech:
Looks good to me. Benchmarks don't seem to have changed too much. I just want to make sure none of the new regexs have simple ReDOS vulnerabilities in them.

src/rules.js
```diff
@@ -168,19 +168,36 @@ const inline = {
...
+  preEm: /^[*_]/,
+  // (1) returns if starts w/ punctuation  | (2)   ⬐Check groups to skip over ⬐ skip if needed ⬐repeat logic for inner *'s (must be in pairs)⬎     ⬐last char can't be punct OR final * must also be followed by punct (or endline)  | (3) Underscores ⬐Check groups to skip over ⬐skip if needed ⬐repeat logic for inner _'s (must be in pairs)⬎  ⬐last char can't be a space, and final _ must preceed punct or \s (or endline)
+  em: /^(?:(\*(?=[punctuation]))|\*)(?![*\s])((?:(?:(?!emSkip)(?:[^*]|[\\\s]\*)|emSkip)|(?:(?:(?!emSkip)(?:[^*]|[\\\s]\*)|emSkip)*?(?<!\\)\*){2})*?)(?:(?<![punctuation\s])\*(?!\*)|(?<=[punctuation])\*(?!\*)(?:(?=[punctuation\s]|$)))|^_(?![_\s])((?:(?:(?!emSkip)(?:[^_]|[\\\s]_)|emSkip)|(?:(?:(?!emSkip)(?:[^_]|[\\\s]_)|emSkip)*?(?<!\\)_){2})*?)(?:(?<![\s])_(?!_)(?:(?=[punctuation\s])|$))/,
```
UziTech, line 176: I believe we still support Internet explorer so regex lookbehinds will not work

UziTech: unless babel transpiles them somehow but I don't think they do.

calculuschild: Oh geez. OK. This might take a while to figure out then. You don't happen to know any alternative regex for lookbehinds?

calculuschild: > I believe we still support Internet explorer

Wait, is this written somewhere?

UziTech: I'm not sure if it is explicit but we will definitely get a lot of people complaining if it just stops working for IE.

We stopped testing with es5 code a while ago but I think the idea is that we will still support IE if issues arise.

calculuschild: Ok. Maybe that would be helpful to put into the contributing guidelines or whatever for the future. I will keep poking at this and see if I can get a workaround without lookbehinds.

davisjam:
TLDR: 1,4 are safe, 2 is exponential, 3 is probably exponential (looks bad but I didn't try an attack).

I didn't look at the diff; its possible these have been exponential for awhile and this change doesn't affect that (in which case we can land it anyway).

1. `inline.reflinkSearch: /!?\[((?:\[(?:\\.|[^\[\]\\])*\]|\\.|`[^`]*`|[^\[\]\\`])*?)\]\[(?!\s*\])((?:\\[\[\]]?|[^\[\]\\])+)\]|!?\[(?!\s*\])((?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]])*)\](?:\[\])?(?!\()/g`

Looks fine.

2. `inline.strong: /^(?:(\*\*(?=[*!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~]))|\*\*)(?![\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)\*){2})+?)(?:(?<![!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])\*\*(?!\*)|(?<=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~])\*\*(?!\*)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s]|$)))|^__(?![\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)_){2})+?)(?:(?<![\s])__(?!_)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])|$))/`

Here's the NFA (I removed the lookbehinds because Regexper does not support them):
image

Note in Group 2, the three sub-patterns below the negative lookbehind can all overlap with each other because they **do not** restrict the "any character" construct. Combined with the outer loop, this can lead to exponential backtracking.

This is the problematic construct (I think we've debugged this before?): `\[.*?\]\(.*?\)|`.*?`|<.*?>)`. Note that it also occurs in the bottom path, presumably with similar ill effect.

Here's a demo attack on this regex. Unclear if such input can reach the input, I didn't check.

```
var at = "**" + "`" + "`<>".repeat(100) + "";

if (/^(?:(\*\*(?=[*!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~]))|\*\*)(?![\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)\*){2})+?)(?:(?<![!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])\*\*(?!\*)|(?<=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~])\*\*(?!\*)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s]|$)))|^__(?![\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)_){2})+?)(?:(?<![\s])__(?!_)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])|$))/.exec(at)) {
  console.log("Match");
} else {
  console.log("No match");
}
```

(Should hang when you run it).

3. `inline.em: /^(?:(\*(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~]))|\*)(?![*\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^*]|[\\\s]\*)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)\*){2})*?)(?:(?<![!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])\*(?!\*)|(?<=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~])\*(?!\*)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s]|$)))|^_(?![_\s])((?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)|(?:(?:(?!\[.*?\]\(.*?\)|`.*?`|<.*?>)(?:[^_]|[\\\s]_)|\[.*?\]\(.*?\)|`.*?`|<.*?>)*?(?<!\\)_){2})*?)(?:(?<![\s])_(?!_)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])|$))/`

Here's the NFA from Regexper, trimmed of lookbehinds again:

image (1)

Eyeballing it, it has the same problematic sub-patterns as the `inline.strong`. (Not surprising, they seem to differ only in the prefix of `*` vs. `**`). I didn't test an attack on this one.

4. `inline.pedantic.em: /^()\*(?=\S)([\s\S]*?\S)\*(?!\*)|^_(?=\S)([\s\S]*?\S)_(?!_)/`

Looks fine. Is there a reason that this begins with an empty capture group?

calculuschild:
> the three sub-patterns below the negative lookbehind can all overlap with each other because they **do not** restrict the "any character" construct.

@davisjam What does this mean? Is this saying we just need to replace the `<*>` with something like `<[^>]*>`?

davisjam:
> @davisjam What does this mean? Is this saying we just need to replace the <*> with something like <[^>]*>?

Yes, that would work (if semantically valid). The problem is that the current pattern is ambiguous -- the regex can parse a `>` as either a closing bracket, or in the `.*` . If you convert `.*` to `[^>]*` then the ambiguity goes away; a `>` can only be parsed in one way.

calculuschild:
Ok that's easy enough. Still passes all the tests so I assume its Ok. @davisjam Any other dangerous spots I should tweak besides needing to replace the lookbehinds for compatibility?

davisjam:
> calculusChild

Great. Can you post the new regexes in their entirety (i.e. after the interpolated content is injected)?

calculuschild:
@davisjam Here ya go. This one is the new version of #3.

/^(?:(\*(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~]))|\*)(?![*\s])(?:(?:(?!\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)(?:[^*]|[\\\s]\*)|\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)|(?:(?:(?!\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)(?:[^*]|[\\\s]\*)|\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)*?(?<!\\)\*){2})*?(?:(?<![!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])\*(?!\*)|(?<=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~])\*(?!\*)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s]|$)))|^_(?![_\s])((?:(?:(?!\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)(?:[^_]|[\\\s]_)|\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)|(?:(?:(?!\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)(?:[^_]|[\\\s]_)|\[[^\]]*?\]\([^\)]*?\)|`[^`]*?`|<[^>]*?>)*?(?<!\\)_){2})*?)(?:(?<![\s])_(?!_)(?:(?=[!"#$%&'()+\-.,/:;<=>?@\[\]`{|}~\s])|$))/

davisjam:
@calculuschild The new version of number 3 looks safe to me.

UziTech:
@davisjam FYI since it doesn't look like regexper has been updated in a while and seems abandoned I created a new version at https://uzitech.github.io/regexper that is updated to allow named groups and lookbehinds.

calculuschild:
Whoooo boy. That was rough. Split EM and Strong regex into "start", "end", and "middle". First a suitable start is found. Then a suitable end is found. Then the middle is matched to see if the inside is valid. If not, grab the next possible end and repeat until a match is found.

Had to go further with the "masking" to also mask everything that used to be in the "emSkip" tag, since splitting up into start and end doesn't let that work anymore.

On the plus side, this now passes 2 more commonmark tests: 390 and 471.

test/specs/new/em_2char.md
```diff
@@ -20,10 +20,6 @@ _123 _
 
 _ 123_
 
-_1__
-
-*1**
```
styfle, line 25: Are you sure you want to remove these tests?

It looks like @UziTech added these in PR #1181.

calculuschild: They do not confirm to the commonmark spec, so trying to get them to pass was, obviously, not working.

styfle: Maybe we can update the output of these two tests to match commonmark's output:

```html
<p><em>1</em>_</p>
<p><em>1</em>*</p>
```

calculuschild: We could... but this PR does not parse it correctly either.

Since it wasn't correct in the earlier version, I feel it would be OK to let this one slide for now (this PR isn't breaking anything that was working before), and work on fixing that test in a future development. Getting this one obscure test to work that wasn't even working before involves a lot of obnoxious workarounds and I don't know if it's worth focusing on it for this PR.

styfle: Yep that makes sense. I think this corner case will probably affect very few people so it can be addressed in a different PR 👍

src/Tokenizer.js
```diff
@@ -489,25 +489,53 @@ module.exports = class Tokenizer {
...
-        text: cap[6] || cap[5] || cap[4] || cap[3] || cap[2] || cap[1]
-      };
+  em(src, maskedSrc, prevChar = '') {
+    let match = this.rules.inline.emStart.exec(src);
+
+    if (match && (!match[1] || (match[1] && (prevChar === '' || this.rules.inline.punctuation.exec(prevChar))))) {
+      maskedSrc = maskedSrc.slice(-1 * src.length);
+      let emEnd;
+
+      if (match[0] === '*') { emEnd = this.rules.inline.emEndAst; } else { emEnd = this.rules.inline.emEndUnd; }
```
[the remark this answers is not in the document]
calculuschild: Done.

src/rules.js
```diff
@@ -168,19 +168,71 @@ const inline = {
...
   nolink: /^!?\[(?!\s*\])((?:\[[^\[\]]*\]|\\[\[\]]|[^\[\]])*)\](?:\[\])?/,
-  strong: /^__([^\s_])__(?!_)|^\*\*([^\s*])\*\*(?!\*)|^__([^\s][\s\S]*?[^\s])__(?!_)|^\*\*([^\s][\s\S]*?[^\s])\*\*(?!\*)/,
-  em: /^_([^\s_])_(?!_)|^_([^\s_<][\s\S]*?[^\s_])_(?!_|[^\s,punctuation])|^_([^\s_<][\s\S]*?[^\s])_(?!_|[^\s,punctuation])|^\*([^\s*<\[])\*(?!\*)|^\*([^\s<"][\s\S]*?[^\s\[\*])\*(?![\]`punctuation])|^\*([^\s*"<\[][\s\S]*[^\s])\*(?!\*)/,
+  reflinkSearch: 'reflink|nolink(?!\\()',
+  strStart: /^(?:(\*\*(?=[*punctuation]))|\*\*)(?![\s])|__/, // (1) returns if starts w/ punctuation
+  strEndAst: /[^punctuation\s]\*\*(?!\*)|[punctuation]\*\*(?!\*)(?:(?=[punctuation\s]|$))/, // last char can't be punct, or final * must also be followed by punct (or endline)
+  strEndUnd: /[^\s]__(?!_)(?:(?=[punctuation\s])|$)/, // last char can't be a space, and final _ must preceed punct or \s (or endline)
+  strong: /^\*\*(?:(?:(?!overlapSkip)(?:[^*]|\\\*)|overlapSkip)|\*(?:(?!overlapSkip)(?:[^*]|\\\*)|overlapSkip)*?\*)+?\*\*$|^__(?![\s])((?:(?:(?!overlapSkip)(?:[^_]|\\_)|overlapSkip)|_(?:(?!overlapSkip)(?:[^_]|\\_)|overlapSkip)*?_)+?)__$/,
```
UziTech, line 175: I wonder if it would be better to put all of the regexes together like

```
strong: {
  start: ...
  middle: ...
  endAst: ...
  endUnd: ...
}
```

just to make it a little more readable. I'm not sure what implications that might have in the rest of the code.

calculuschild: Sure. That makes sense.

calculuschild: Done. I had to redo the pedantic section as well to get names matching but behavior should be the same.

calculuschild:
What do you guys think? Is this baby ready to go?

brainchild0:
So 424 and 425 are considered out of scope of this change set? 

Or more simply: 

```
~/dst/dev/marked$ git show -s
commit 6b729ed8cdb98ea75d4031f6218a1f58b9f02d8a (HEAD, EmphasisFixes)
Merge: e27e6f9 ad720c1
Author: Trevor Buckner <calculuschild@gmail.com>
Date:   Thu Jul 9 19:37:22 2020 -0400

    Merge branch 'EmphasisFixes' of https://github.com/calculuschild/marked into EmphasisFixes
$ 
$ echo '**a **b** c**' |  bin/marked 
<p><strong>a **b</strong> c**</p>
```


(Should be `<p><strong>a <strong>b</strong> c</strong></p>`.)

UziTech:
@brainchild0 if you want to fix it I would be happy to review a PR 😁

brainchild0:
For me the starting point is understanding what makes this particular case more challenging than the ones that have been resolved in this merge. Intuitively they all seem roughly equally complex. Obviously, I make the remark abstractly, having no familiarity with the design.

Or another way to ask the question, starting from the top and moving down: The general rule   is to collect a stack of emphasis delimiters, which may be any of `*`, `_`, `**`, or `__`. As long as this rule may be somehow fully implemented, along with also dropping those same items from the stack, because of close (right flanking) delimiters, then all cases would be supported. What is missing, currently, that leaves out support for the remaining cases?

calculuschild:
> What is missing, currently, that leaves out support for the remaining cases?

The current implementation does not use a stack at all. It simply checks for existence of a left delimiter, then if found, find the first available matching right delimiter. Finally, ensure the text between the two is valid, meaning ignore any delimiters found inside links or code spans etc., and any other delimiters inside must occur in even pairs.

We already have regex for the left and right delimiters, so it would just be the extra effort of building up a stack in the tokenizer.

fredck:
There is a good chance that this PR introduced #1754.

Issue #1692, opened 2020-05-31
documentation to show ESM import  and usage


henriquez:
**Describe the bug**
I'm using rollup to bundle dependencies (for a Svelte framework web app) but having trouble importing and using markedjs.  It appears the library was recently ported to ESM modules, but I don't see any documentation on how to use as such.  The standard syntax to import ES6 modules give me errors:

`import { marked }  from 'marked.esm';`
returns
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
marked.esm (imported by src/parser/parser.js)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
marked.esm (guessing 'marked')

and 

`import { marked }  from 'marked';`
returns 
bundles src/main.js → public/build/bundle.js...
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
node_modules/marked/lib/marked.js
14:   typeof define === 'function' && define.amd ? define(factory) :
15:   (global = global || self, global.marked = factory());
16: }(this, (function () { 'use strict';
      ^
17: 
18:   function _defineProperties(target, props) {
[!] Error: 'marked' is not exported by node_modules/marked/lib/marked.js, imported by src/parser/parser.js
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
src/parser/parser.js (11:9)

**To Reproduce**
npm install marked

in a Svelte js file that's bundled by rollup:
import { marked }  from 'marked';   // or similar variations

**Expected behavior**
no import errors

KilianKilmister:
marked.esm is using a default export instead of a named export. so the proper syntax would be:

```js
import marked from 'marked' // note that this is a definition, so you can give it any name you want
```

But i think you are raising an important point: Why was a default export chosen over a named export? It removes many of the advantages of using esm. As the result is still just some convoluted mess of properties.

It would be much more userfriendly with well organized named exports as they work excellently with TS-language-features.

I'll file a seperate issue about that in a bit

henriquez:
I tried that syntax as well, but errors with "'default' is not exported by node_modules/marked/lib/marked.js" detail below:

`bundles src/main.js → public/build/bundle.js...
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
node_modules/marked/lib/marked.js
14:   typeof define === 'function' && define.amd ? define(factory) :
[...]

PR #1725, opened 2020-07-09
Added "exports"-field to package.json so import/require loads the appropriate file without requiering additional user-input

KilianKilmister:
**Marked version:** @latest

**Markdown flavor:** n/a

## Description

- adresssing #1692 
- added "export"-filed to package.json
- adjusted rollup.config.esm.js to output `.mjs`
- built and committed `lib/marked.esm.mjs`

I did a superficial test to ensure the paths are correct.
VSCode debugger showed that `import marked from 'marked'` will only load `lib/marked.esm.mjs` 

#### citing my comment

> cjs-require/esm-import can be automatically handled by specifying an `"exports"` field in the package.json.
> This way node will load the appropriate files without the user having to worry about in-module paths.
> 
> ## Short explanation
> ```js
> {
>   "exports": {
>     [exportPath]: {
>       "import": "./path/to/file.mjs",
>       "require": "./path/to/file.cjs"
>     }
>   }
> }
> ```
> 
> ### So for marked it would be:
> ```js
> {
>   "exports": {
>     ".": {
>       "import": "./lib/marked.esm.mjs", // `.mjs`-extention is needed
>       "require": "./src/marked.js"
>     }
>   }
> }
> ```
> 
> I'll file a PR for that shortly

## Contributor

## Committer

- [ ] Merge PR


KilianKilmister:
#1692

UziTech:
According to [Node](https://nodejs.org/api/esm.html#esm_package_entry_points):

> **Warning:** Introducing the `"exports"` field prevents consumers of a package from using any entry points that are not defined, including the `package.json` (e.g. `require('your-package/package.json')`. **This will likely be a breaking change.**

So we will need to do a major version bump for this.

package.json
```diff
@@ -34,6 +40,8 @@
     "@babel/core": "^7.9.6",
     "@babel/preset-env": "^7.9.6",
     "@markedjs/html-differ": "^3.0.2",
+    "@types/node": "^14.0.9",
+    "babel-eslint": "^10.1.0",
```
UziTech, line 44: why are these extra dependencies on here?

UziTech:
I also don't believe we should implement this until es modules are not experimental as the interface could still change.

It seems this is only supported in node v14+ so we should probably wait until v14 becomes LTS.

UziTech:
> Is importing a package.json a "common" occurance?

Not `package.json`, but I know there are some dependents that import just certain files like `require('marked/lib/marked.js')` or `require('marked/src/Renderer.js')`. We would need to define all valid entry points in the `"exports"` field not just `"."`.

KilianKilmister:
> > Is importing a package.json a "common" occurance?
> 
> Not `package.json`, but I know there are some dependents that import just certain files like `require('marked/lib/marked.js')` or `require('marked/src/Renderer.js')`. We would need to define all valid entry points in the `"exports"` field not just `"."`.

This would be the files in `src` and `lib`, right? We could list those folders and expose them as a whole

evanplaice:
> It seems this is only supported in node v14+ so we should probably wait until v14 becomes LTS.

This isn't true. ESM was introduced in Node 13.x and back ported to 12.x LTS. With 10.x being EOL'd there is no longer an officially supported version of Node that doesn't work with modules.

For the issue of direct path imports being unavailable after pkg.exports is introduced. Instead of defining an export for every file you can provide a facade instead that re-exports all of the modules. Like this (https://github.com/vanillaes/absurdum/blob/main/src/arrays/index.js) but using default exports because none of the existing components are written in ESM.

If you don't want to support 2 copies (ie CJS and ESM), you can always use RollupJS or ESBuild to convert from ESM to CJS. ESM -> CJS conversion is a LOT cleaner than CJS -> ESM.

UziTech:
If this PR was rebased I would be fine with including the exports field now :+1:

package.json
```diff
@@ -14,6 +14,12 @@
     "man/",
     "marked.min.js"
   ],
+  "exports": {
+    ".": {
+      "import": "./lib/marked.esm.mjs",
+      "require": "./src/marked.js"
+    }
```
benmccann, line 21: you should also export `package.json` as some bundlers will want to read it
```suggestion
    },
    "./package.json": "./package.json"
```

lib/marked.esm.mjs
```diff
@@ -0,0 +1,2206 @@
+/**
```
benmccann, line 1: I don't know what this project's processes are, but I'd imagine this file should only be updated when a release is cut? If this file is going to be added then the old `lib/marked.esm.js` should be removed

Issue #664, opened 2015-09-27
Inconsistent anchor-id linkage


KrishnaPG:
While trying to render markdown docs from github getting inconsistent anchor / link combinations (due to the header ids being forced to lowercase ). For an example, below is one such doc:
````
# Introduction

- [Introduction](Introduction),
  - [Reproducible research](#Reproducible-research)
  - [We need better tools!](#We-need-better-tools!)
  - [Introducing psychopy_ext](#Introducing-psychopy_ext)

## Reproducible research

So far we've discussed how to code experiments. But research is more than just making an experiment! You have to analyze data, possibly also compare them to simulated data, present them in conferences and publish in journals

## We need better tools!

I do not blame researchers for relying on all these ad-hoc solutions. While in theory it would be nice to code everything from A to Z, in practice we don't have the time to play these silly games. Why we don't have the time is a topic for a separate discussion, but why we have to play these silly games is the problem of software. Simply put, we the lack tools that would seamlessly enact good coding and sharing standards. We need **tools that act clever**.
````
The generated header ids are all [forced lower case](https://github.com/chjj/marked/blob/master/lib/marked.js#L800), while the generated `a href` links are retained "as-is".

-    Please add an option to control the casing for all generated header ids and internal hrefs (to either use "same case" or force lowercase). 
-     or, consistently force the same casing for all generated internal link `hrefs` and anchor ids,
-     or, when user explicitly specifies an internal link (as above) anywhere in the markup, ensure that generated ids are consistent with user expectations (without changing the case anywhere).

Forcing the lowercase only for header ids, while leaving the `<a href` "as-is" is causing un-navigatable docs to be rendered.

Feder1co5oave:
Auto-generated header IDs is a much discussed topic for marked, so any change to the current default behavior is going to upset some people.
The official suggestion is to never rely on the IDs generated by the default renderer and create one of your own that suits your needs.
For you, the following should do.
~~~javascript
renderer = new marked.Renderer;
renderer.heading = function(text, level, raw) {
  return '<h'
    + level
    + ' id="'
    + this.options.headerPrefix
    + raw.replace(/[^\w]+/g, '-')
    + '">'
    + text
    + '</h'
    + level
    + '>\n';
};
html = marked(src, { renderer: renderer });
~~~

KrishnaPG:
Thank you.  Your code snippet is helpful.

[...]

PR #1728, opened 2020-07-11
NFE: add dry run capabilities to default slugger

flouc001:
**Marked version: 1.1.0**

**Markdown flavor: n/a**

## Description

This PR adds a feature that allows the developer to use the default slugger without updating the internal accumulator. I have seen a use case for this when developers are implementing their own custom renderers. For example, see [this function](https://github.com/tc39/proposal-temporal/blob/5838233ecf733f70361880f237ed8c3fa8d5b837/docs/buildDocs.js#L45-L64) in the TC39 temporal-proposal repository. The changes made can be summarized as:
- Extends the default slugger with the ability to generate slug text without trigger internal communication.
- Separates slugger responsibilities into methods for re-use.

## Contributor

## Committer

- [x] Merge PR

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -85,10 +85,12 @@ slugger.slug('foo-1') // foo-1-2
 `slugger.slug` can also be called with the `dryrun` option for stateless operation:
 ```js
 slugger.slug('foo')                    // foo
-slugger.slug('foo', { dryrun: true })  // foo-1
 slugger.slug('foo')                    // foo-1
-slugger.slug('foo', { dryrun: true })  // foo-2
 slugger.slug('foo')                    // foo-2
+slugger.slug('foo', { dryrun: true })  // foo-3
+slugger.slug('foo', { dryrun: true })  // foo-3
+slugger.slug('foo')                    // foo-3
+slugger.slug('foo')                    // foo-4
 ...
 ```
 
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -82,6 +82,16 @@ slugger.slug('foo-1') // foo-1-2
 ...
 ```
 
+`slugger` also has a stateless method `slugText` which does not update the internal accumulator:
+```js
+slugger.slug('foo')      // foo
+slugger.slugText('foo')  // foo-1
+slugger.slug('foo')      // foo-1
+slugger.slugText('foo')  // foo-2
+slugger.slug('foo')      // foo-2
+...
+```
+
 `flags` has the following properties:
 
 ```js
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -82,13 +82,13 @@ slugger.slug('foo-1') // foo-1-2
 ...
 ```
 
-`slugger` also has a stateless method `slugText` which does not update the internal accumulator:
+`slugger.slug` can also be called with the `dryrun` option for stateless operation:
 ```js
-slugger.slug('foo')      // foo
-slugger.slugText('foo')  // foo-1
-slugger.slug('foo')      // foo-1
-slugger.slugText('foo')  // foo-2
-slugger.slug('foo')      // foo-2
+slugger.slug('foo')                    // foo
+slugger.slug('foo', { dryrun: true })  // foo-1
+slugger.slug('foo')                    // foo-1
+slugger.slug('foo', { dryrun: true })  // foo-2
+slugger.slug('foo')                    // foo-2
 ...
 ```
 
--- a/src/Slugger.js
+++ b/src/Slugger.js
@@ -6,11 +6,8 @@ module.exports = class Slugger {
     this.seen = {};
   }
 
-  /**
-   * Convert string to unique id
-   */
-  slug(value) {
-    let slug = value
+  serialize(value) {
+    return value
       .toLowerCase()
       .trim()
       // remove html tags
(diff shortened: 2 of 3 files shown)


src/Slugger.js
```diff
@@ -6,28 +6,50 @@ module.exports = class Slugger {
...
+    return this.getNextSafeSlug(slug);
+  }
+
+  /**
+   * Get slug text without incrementing accumulator
+   */
+  slugText(value) {
+    const slug = this.serialize(value);
+    return this.getNextSafeSlug(slug, true);
+  }
```
UziTech, line 54: Instead of having a new method could we just pass an option. That way in the future if we want to create more options we don't need to keep adding methods and changing parameters.

```suggestion
  slug(value, option = {}) {
    const slug = this.serialize(value);
    return this.getNextSafeSlug(slug, options.dryrun);
  }
```

test/unit/marked-spec.js
```diff
@@ -63,6 +63,12 @@ describe('Test slugger functionality', () => {
     const slugger = new marked.Slugger();
     expect(slugger.slug('<em>html</em>')).toBe('html');
   });
+
+  it('should not increment seen when just getting text', () => {
+    const slugger = new marked.Slugger();
+    slugger.slugText('<h1>This Section</h1>');
```
UziTech, line 69: we should test the output of the dryrun as well. Could you also add a test for a dryrun after a regular run?

src/Slugger.js
flouc001: Sure, this sounds like a good idea, I'll make the change and update the tests!

flouc001: What's your preferred way of documenting the option for intellisense? JSDoc?

test/unit/marked-spec.js
flouc001: Of course 👍

flouc001:
@UziTech I've made the changes you requested, the only thing I wasn't sure on was how you wanted the option documented in the code. I've added basic JSDoc but I can't see an example of it being used similarly anywhere else in the codebase so happy to take it out 👍

UziTech:
JSDoc works fine. Documentation in the code could definitely be better.

joshbruce:
Not that is has too and it would be good to know.

Does this affect special characters and asian character sets as that's been a topic: https://github.com/markedjs/marked/issues/664

Or, is it specifically giving the ability for a developer to intervene and verify themselves??

In either case, I will approve but not merge as this one feels a bit deeper than my understanding at present.

flouc001:
@joshbruce 
> Does this affect special characters and asian character sets as that's been a topic: #664
This code should not affect how the heading id is actually parsed so there will be no deviation from default behaviour on that front.
> Or, is it specifically giving the ability for a developer to intervene and verify themselves??
Exactly, this option is all about giving control to the developer who, for whatever reason, may want to generate heading ids without incrementing the Slugger's seen counter. At the moment there is no way to generate the heading id without causing the Slugger to mutate that internal counter.

docs/USING_PRO.md
```diff
@@ -82,6 +82,16 @@ slugger.slug('foo-1') // foo-1-2
 ...
 ```
 
+`slugger.slug` can also be called with the `dryrun` option for stateless operation:
+```js
+slugger.slug('foo')                    // foo
+slugger.slug('foo', { dryrun: true })  // foo-1
+slugger.slug('foo')                    // foo-1
+slugger.slug('foo', { dryrun: true })  // foo-2
+slugger.slug('foo')                    // foo-2
```
styfle, line 91: Perhaps change this to the following:

```suggestion
slugger.slug('foo')                    // foo
slugger.slug('foo')                    // foo-1
slugger.slug('foo')                    // foo-2
slugger.slug('foo', { dryrun: true })  // foo-2
slugger.slug('foo', { dryrun: true })  // foo-2
slugger.slug('foo')                    // foo-3
```

styfle: And maybe we should add that as a test ☝️

flouc001: Sure thing, I'll get this changed in the morning 👍

UziTech: Shouldn't the two dryruns return `"foo-3" ` since it should show what will happen on the next run?

styfle: Yep, you're right, I updated my example.

More reason to update this since its not immediately clear if its supposed to return the last slug or next slug.

Maybe we can think of something better than `dryrun`? Can't think of anything off the top of my head that would describe the expected behavior 🤔

test/unit/marked-spec.js
```diff
@@ -63,6 +63,18 @@ describe('Test slugger functionality', () => {
     const slugger = new marked.Slugger();
     expect(slugger.slug('<em>html</em>')).toBe('html');
   });
+
+  it('should not increment seen when using dryrun option', () => {
+    const slugger = new marked.Slugger();
+    slugger.slug('<h1>This Section</h1>', { dryrun: true });
```
UziTech, line 69:
```suggestion
    expect(slugger.slug('<h1>This Section</h1>', { dryrun: true })).toBe('this-section');
```

docs/USING_PRO.md
flouc001: I did previously think about calling it `stateless`?

flouc001:
@styfle I've added the test you recommended and committed all outstanding suggestions 👍

docs/USING_PRO.md
styfle: Let's keep `dryrun` 👍

PR #1735, opened 2020-07-20
Add Homebrew install to readme

SeekingMeaning:
**Marked version:** 1.1.0

**Markdown flavor:** n/a

## Description

## Contributor

## Committer

- [ ] Merge PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ Also read about:
 
 **In-browser:** `npm install marked`
 
+**Homebrew:** `brew install marked`
+
 ## Usage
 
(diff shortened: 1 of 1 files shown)


UziTech:
I'm not sure we want to support this way of installing marked. There are probably millions of different tools that could be used to install marked but we don't have time to make sure they all continue to work.

@styfle what do you think?

UziTech:
it looks like the code that controls which version of marked is installed this not controlled by us anyway.

https://github.com/Homebrew/homebrew-core/blob/master/Formula/marked.rb

styfle:
I wouldn't list it here since no one from the core team is maintaining it.

SeekingMeaning:
Sure, no problem, we'll try to keep the package updated from our side

PR #1740, opened 2020-08-06
Added .idea folder to .gitignore

xepozz:


xepozz:
Nice solution, thanks!

PR #1744, opened 2020-08-11
Org. owner to UziTech

joshbruce:
**@mention the contributor:**

## Recommendation to:

- [x] Change user group
- [ ] Add a badge
- [ ] Remove a badge

If it weren't for @UziTech, none of us would still be here - at least not begrudgingly. He has done wonders over the past 2+ years in improving the developer experience around here.

Not just in the tech and tools department - though that's pretty awesome - but also in the leading by example and helping ensure a non-toxic environment. We have seen so many commits and chip-ins from the community since he...inspired...me loosen the rock. Since then, it's been wonderful to see what talented, empowered people can do when contributing in the areas they want to.

(Sorry I didn't realize you weren't already an owner.)

Thanks...for everything. Personally, I've grown a lot thanks to Marked, and it wouldn't have happened without you.

## As the one mentioned, I would like to:

- [x] accept the recommendation; or,
- [ ] graciously decline; or,
- [ ] dispute the recommendation

within 30 days, if you have not indicated which option you are taking one of the following will happen:

1. If adding a badge, we will assume you are graciously declining.
2. If removing a badge, we will assume you do not want to dispute the recommendation; therefore, the badge will be removed.

Note: All committers must approve via review before merging, the disapproving committer can simply close the PR.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,3 +8,4 @@ Having said that, we take the security of Marked very seriously.
 Please disclose potential security issues by email to the project [committers](https://marked.js.org/#/AUTHORS.md) as well as the [listed owners within NPM](https://docs.npmjs.com/cli/owner).
 We will provide an initial assessment of security reports within 48 hours and should apply patches within 2 weeks
 (also, feel free to contribute a fix for the issue).
+


joshbruce:
@davisjam - Need you in on this one, brother as we don't have a fallback policy or 2/3 thing. Same with #1743

PR #1749, opened 2020-08-17
Static docs

styfle:
## Description

- Change documentation to use SSG instead of SPA via `npm run build:docs`
- Change page URLs to lowercase and use [`cleanUrls`](https://vercel.com/docs/configuration#project/cleanurls)
- Only uppercase markdown files are compiled to html pages
- Add unique `<title>` to each page
- Add syntax highlighting to code snippets
- Add license as its own page instead of a link
- Optimize svg logo, about 20% smaller
- Redirect from old URL to new URL format
 

## Contributor

- [x] no tests required for this PR, the [Vercel GitHub Integration](https://vercel.com/github) will show a red X if the build fails

## Committer

- [x] Merge PR


styfle:
@UziTech @joshbruce Do we want to use [`cleanUrls`](https://vercel.com/docs/configuration#project/cleanurls) so the URLs look a bit prettier?

For example, it would change `marked.js.org/using_advanced.html` to `marked.js.org/using_advanced`

UziTech:
I prefer clean urls but I don't know if there is a benefit one way or the other.

We do need to make sure that old links are properly redirected.

styfle:
I prefer clean urls too. We'll have to do the redirect on the client side since it needs to read the hash. I'll add that soon 👍

joshbruce:
I also prefer cleaner URLs from a UX perspective. Email link doesn't say https://marked.org/?page=12 but /play-with-marked or something - totally making stuff up as I do the inbox zero thing before a meeting.

package.json
```diff
@@ -67,6 +69,7 @@
     "lint": "eslint --fix bin/marked .",
     "build:reset": "git checkout upstream/master lib/marked.js lib/marked.esm.js marked.min.js",
     "build": "npm run rollup && npm run minify",
+    "build:docs": "node build-docs.js",
```
UziTech, line 72: Does this have to be called manually any time the docs are changed? Should we run this in the [Build step](https://github.com/markedjs/marked/blob/master/.github/workflows/tests.yml#L40) on merge?

styfle: We don't commit the docs output so it doesn't need to run unless you want to see the docs output.

Would you prefer I rename this script to `vercel-build` instead of `build:docs` so its clear its for Vercel?

UziTech: How does vercel know to run `npm build:docs` to get the output? and when does it run (on every PR, on every commit to master)?

styfle: The build command is configured here: https://vercel.com/markedjs/markedjs/settings

It runs every `git push` regardless of the branch. The `master` branch gets the production domain, https://marked.js.org, and PRs get the preview domains such as https://markedjs-git-static-docs.markedjs.vercel.app

UziTech:
Do we have documentation on how the documentation is built and where a markedjs member would go to edit it?

I like the idea of being able to use server side functions for the docs but I also like the simplicity of github pages and having all of the code/settings in a central place.

styfle:
@UziTech You would go to https://vercel.com to edit project settings and https://vercel.com/docs for documentation

PR #1755, opened 2020-09-04
fix underscore adjacent to asterisk

### LINKED ISSUE #1754: Touching italic/bold doesn't render right (v1.1.1 regression)
When italic and bold are touching, italic is ignored:

```
// Input
_te_**st**

// Expected output
<p><em>te</em><strong>st</strong></p>

// Actual output (wrong)
<p>_te_<strong>st</strong></p>
```

This was [working with v1.1.0](https://marked.js.org/demo/?outputType=html&text=_te_**st**&options=&version=1.1.0) and is [not working with v1.1.1](https://marked.js.org/demo/?outputType=html&text=_te_**st**&options=&version=1.1.1).

UziTech:
**Marked version:** 1.1.1

## Description

Fix underscore adjacent to asterisk

- Fixes #1754 

## Contributor

## Committer

- [x] Merge PR


src/rules.js
```diff
@@ -222,7 +222,7 @@ inline.strong.start = edit(inline.strong.start)
 
 inline.strong.middle = edit(inline.strong.middle)
   .replace(/punctuation/g, inline._punctuation)
-  .replace(/blockSkip/g, inline._blockSkip)
+  .replace(/overlapSkip/g, inline._overlapSkip)
```
UziTech, line 225: this seems to be a typo from #1686

styfle (approved):
Thanks!

src/rules.js
calculuschild: Good catch. Did that make any difference on the tests?

UziTech: no it didn't. I think in order for us to pass all of the commonmark specs for em and strong we will need a recursive way to skip internal strong and em. For example `**a **b** c**` should output `<strong>a <strong>b</strong> c</strong>` but it does not. We would need some way to realize that `b**` is the end of the second strong and not the first.

UziTech: To be honest I don't know why anyone would actually do that. I'm ok with saying that is just garbage in garbage out but it is spec'd so we should probably try to find a way to fix it.

UziTech:
@joshbruce @davisjam we just need one more approval to merge this

joshbruce (approved):

joshbruce:
@UziTech: Not in a position to finish the approved portion of checklist or tag others apparently despite trying.

calculuschild (approved):

PR #1762, opened 2020-09-24
fix: fix indented code in list item

UziTech:
**Marked version:** 1.1.1

## Description

fixes indented code in list item. Passes 100% tabs spec.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression

## Committer

In most cases, this should be a different person than the contributor.

- [ ] Draft GitHub release notes have been updated.
- [ ] CI is green (no forced merge required).
- [ ] Merge PR

PR #1767, opened 2020-09-28
chore: use semantic release

UziTech:
## Description

Create new releases with semantic release

## Contributor

## Committer


UziTech:
I already added an NPM_TOKEN secret to our organization so this should work after it is merged.

joshbruce:
Ah. Semantic Release is not the same as just semantic versioning. Given our other conversation want to make sure "squash" and "merge" is desired but not necessarily required.

UziTech:
[semantic release](https://github.com/semantic-release/semantic-release) is a software that releases software based on commit messages. It will handle everything from changelogs to publishing to npm and github.

The only thing that is required is that commit messages be formatted according to the [conventional commit guidelines](https://www.conventionalcommits.org/) so it knows when to release a major, minor, or patch version and what to enter into the changelog.

UziTech:
This is the same way we release new versions of [@markedjs/html-differ](https://github.com/markedjs/html-differ)

.github/workflows/tests.yml
```diff
@@ -37,26 +37,30 @@ jobs:
...
             git config --global user.email "<>"
             git config --global user.name "MarkedJS bot"
             git commit -am "🗜️ build [skip ci]"
-            git push "https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/markedjs/marked.git" HEAD:master
           fi
+      - name: Release 🎉
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx sematic-release
```
styfle, line 63: Where is this coming from? I don't see a package named [semantic-release](https://www.npmjs.com/package/sematic-release) in npm.

.github/workflows/tests.yml
```diff
@@ -37,26 +37,30 @@ jobs:
...
       - name: Build 🗜️
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           npm run build
           if ! git diff --quiet; then
             git config --global user.email "<>"
             git config --global user.name "MarkedJS bot"
             git commit -am "🗜️ build [skip ci]"
-            git push "https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/markedjs/marked.git" HEAD:master
```
styfle, line 58: Will semantic release push back the changes to the master branch?

package.json
```diff
@@ -31,29 +31,36 @@
...
-    "@markedjs/html-differ": "^3.0.2",
+    "@babel/core": "^7.11.6",
+    "@babel/preset-env": "^7.11.5",
+    "@markedjs/html-differ": "^3.0.3",
+    "@semantic-release/changelog": "^5.0.1",
+    "@semantic-release/commit-analyzer": "^8.0.1",
+    "@semantic-release/git": "^9.0.0",
+    "@semantic-release/github": "^7.1.1",
+    "@semantic-release/npm": "^7.0.6",
+    "@semantic-release/release-notes-generator": "^9.0.1",
```
styfle, line 42: That's a lot of extra dependencies 😧 

Do we need all of these? For example, `@semantic-release/changelog` seems unnecessary since we don't keep a changelog. I think `@semantic-release/release-notes-generator` is sufficient.

.github/workflows/tests.yml
UziTech: yes

UziTech: That is what @semantic-release/git does

.github/workflows/tests.yml
UziTech: typo

package.json
UziTech: I didn't even realize we didn't have a changelog.md

.github/workflows/tests.yml
styfle: Since `semantic-release` isn't defined in package.json, this is going to install the latest.

I think we should use a specific version of semantic-release and put it in package.json devDependencies like the rest.

UziTech: It is in package.json. This will run the version that is in package.json

styfle: Oh yes you're right, I see it on line 60 🤦‍♂️

styfle (approved):
Looks great, thanks!

package.json
aminya: Yeah. These dependencies are not needed! You can just install them in the CI if not included in the action

https://github.com/atom-ide-community/atom-ide-datatip/blob/fe61eec0db91901ef85f2f3eb42c27ae3820c5b1/.github/workflows/ci.yml#L75

aminya (changes requested):

.github/workflows/tests.yml
```diff
@@ -37,26 +37,30 @@ jobs:
...
           fi
+      - name: Release 🎉
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release
```
aminya, line 63:
```suggestion
        - name: Release 🎉
        uses: cycjimmy/semantic-release-action@v2
        with:
          extra_plugins: |
            @semantic-release/git@9.0.0
            @semantic-release/github@7.1.1
            @semantic-release/npm@7.0.6
            @semantic-release/release-notes-generator@9.0.1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

package.json
```diff
@@ -31,29 +31,35 @@
...
-    "@babel/preset-env": "^7.10.4",
-    "@markedjs/html-differ": "^3.0.2",
+    "@babel/core": "^7.11.6",
+    "@babel/preset-env": "^7.11.5",
+    "@markedjs/html-differ": "^3.0.3",
+    "@semantic-release/commit-analyzer": "^8.0.1",
+    "@semantic-release/git": "^9.0.0",
+    "@semantic-release/github": "^7.1.1",
+    "@semantic-release/npm": "^7.0.6",
+    "@semantic-release/release-notes-generator": "^9.0.1",
```
aminya, line 41:
```suggestion
```

release.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  plugins: [
+    '@semantic-release/commit-analyzer',
+    '@semantic-release/release-notes-generator',
+    '@semantic-release/npm',
+    '@semantic-release/github',
+    '@semantic-release/git'
+  ]
+};
```
aminya, line 9:
```suggestion
```

aminya (changes requested):
These dependencies are not needed to be installed locally! You just need to install them in the CI.

.github/workflows/tests.yml
UziTech: Then we would have to manually update the versions when new versions are released instead of relying on tools. I think the small amount of disk space they take up is worth the small amount of developer time it would take with this approach.

aminya: You don't need to specify the versions. Specifying the names is enough.

UziTech: Then it always installs the latest and can break things if a major version is released. I still think keeping them in package.json with version numbers is better.

UziTech: They are only dev dependencies if you really don't want them installed you can go through the extra work of removing them from package.json before running npm i on your machine.

UziTech: for that matter if you use pnpm they won't take up any extra space if another package uses them.

UziTech: If you are really concerned about ci dependencies you should petition npm to create a `ciDependencies` field for these types of dependencies.

aminya: I don't think these break anything. The semantic release supports configs as well. The configs wrap these functionalities in a preset. This is similar to `apm-config` that I use in atom-ide-community. I would be surprised if there is not any config for npm.

https://github.com/semantic-release/apm-config#plugins

styfle: I don't think we want to use unofficial semantic release tooling--it would increase the scope of vulnerabilities and it wouldn't be checked by github's automated vulnerability checks. And it basically gives cycjimmy access to our releases which is not a risk we need to take.

UziTech:
@styfle should I add dependabot to this PR to keep dependencies up to date?

styfle:
@UziTech Its already enabled in the github repo settings 
image

UziTech:
That is just security alerts. I'm talking about adding a [dependabot.yml](https://github.com/markedjs/html-differ/blob/master/.github/dependabot.yml) file to update dependencies when they are released.

styfle:
@UziTech Got it. I was thinking they integrated the full feature set in GitHub, but we can add the yaml file.

Does weekly checks sound good?

UziTech:
GitHub just made [codeql](https://securitylab.github.com/tools/codeql/atm) available for open source projects so I added that as well.

styfle (approved):
:shipit:

UziTech:
@joshbrue do you have any objections to this and requiring "Squash and merge"

UziTech:
@joshbruce once we merge this PR we can test out semantic release by merging the fix in #1762.

UziTech:
@joshbruce @davisjam can we get this merged so we can get releases automated?

davisjam (approved):
This is very cool.

UziTech:
@joshbruce if you don't have any objection to this I will merge this tomorrow.

PR #1768, opened 2020-09-28
Update Lexer.js

### LINKED ISSUE #1766: Using custom Tokenizer disables url rule
**Workaround**
For those coming from Google looking for a resolution, here's a simple solution:

```javascript

const lexer = new marked.Lexer({
  ...marked.defaults,
  tokenizer: new marked.Tokenizer()
})

console.log(lexer.tokenizer.rules.inline.url)

```

**Describe the bug**
When creating a lexer and passing in a tokenizer (`tokenzier: new marked.Tokenizer()`), the `url` rule gets assigned a value of `{ exec: [Function: noopTest] }`.

**To Reproduce**
See this example on Repl.it: https://repl.it/@trevorhreed/marked-multi-instance#index.js

Here's the minimal example: 

```javascript

const lexer1 = new marked.Lexer()
console.log(lexer1.tokenizer.rules.inline.url) // Output: /^((?:ftp|https?):\/\/|www\.)(?:[a-zA-Z0-9\-]+\.?)+[^\s<]*|^[A-Za-z0-9._+-]+(@)[a-zA-Z0-9-_]+(?:\.[a-zA-Z0-9-_]*[a-zA-Z0-9])+(?![-_])/i

const lexer2 = new marked.Lexer({
  tokenizer: new marked.Tokenizer()
})
console.log(lexer2.tokenizer.rules.inline.url) // Output: { exec: [Function: noopTest] }
```

**Expected behavior**
I expect the ruleset to be the same regardless of whether or not I pass in a tokenizer that I've instantiated myself or not (assuming all other options passed into the Lexer/Tokenizer are the same).

trevorhreed:
**Marked version: 1.2.0**
**Markdown flavor:** n/a

## Description

- Fixes #1766

## Contributor

*This PR may break existing functionality and may not be appropriate until a bump in major version*

## Committer

- [ ] Merge PR

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -50,7 +50,7 @@ module.exports = class Lexer {
   constructor(options) {
     this.tokens = [];
     this.tokens.links = Object.create(null);
-    this.options = options || defaults;
+    this.options = Object.assign(defaults, options);
     this.options.tokenizer = this.options.tokenizer || new Tokenizer();
     this.tokenizer = this.options.tokenizer;
     this.tokenizer.options = this.options;


UziTech:
this makes marked about 2x slower

trevorhreed:
How often is the constructor being called?

UziTech:
Once per call. But the bench marks go from ~6s to ~8s with this. I was exaggerating a little when I said 2x but it is unnecessarily slower which is why we recommend setting the options before running markdown through marked.

UziTech:
I think the fix in https://github.com/markedjs/marked/issues/1766#issuecomment-700279699 is a much better solution.

trevorhreed:
That solution works for me.

PR #1813, opened 2020-11-07
fix: Add type to def token

sawa-zen:
**Marked version:** 1.2.3

## Description

Added `def` token because it didn't have a type only.

## What was attempted

When I tried to switch the type of tokens in a TypeScript project, I noticed the problem that type doesn't always exist and found that type was missing only in the def token.

## Contributor

## Committer

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -320,6 +320,7 @@ module.exports = class Tokenizer {
       if (cap[3]) cap[3] = cap[3].substring(1, cap[3].length - 1);
       const tag = cap[1].toLowerCase().replace(/\s+/g, ' ');
       return {
+        type: 'def',
         tag,
         raw: cap[0],
         href: cap[2],


UziTech:
The `def` token is different than the other tokens since it isn't rendered at all and it is not inline in the token list. `def` tokens are help separate in `tokens.links` object.

This PR would change anything in the output so I don't understand why it would be relevant.

https://github.com/markedjs/marked/blob/c50e39e58a5127e56455ae768e6b080504339db7/src/Lexer.js#L203-L213

sawa-zen:
@UziTech 
I see. Personally I think it's unnatural that `type` doesn't exist as a return type for the tokenizer, even though `def` tokens aren't used for processing, but it's dangerous for the output to be changed for this fix.

This PR is closed.

deltachonglee81 (commented):
@styfle ![Uploading 98992121...]

PR #1832, opened 2020-11-17
fix: em and strong starting with special char

### LINKED ISSUE #1831: Bold "@" in block quote doesn't render correctly
**Marked version:**
1.1.1 and 1.2.4

**Describe the bug**
We use marked for comments in our app, and having a bold mention in the second line of a block quote doesn't work. 

**To Reproduce**
This input:
```
> Line 1\n> **@person**
```
Produces this HTML:
```html
<blockquote>
    <p>
        Line 1
        <br>
        **@person**
    </p>
</blockquote>
```
I could not repro the bug in the Marked Demo because it does not support newlines/br tags (or I couldn't figure out how to use them). 

**Expected behavior**
The `> **@person**` line should be wrapped in `<strong>` tags, but it instead shows the asterisks. This only repros when the bolded mention is after the first line, is at the beginning of the line, and starts with a `@`.
Expected output
```html
<blockquote>
    <p>
        Line 1
        <br>
        <strong>@person<strong>
    </p>
</blockquote>
```

UziTech:
**Marked version:** 1.2.4

## Description

Fix em that starts with a special character after another inline element

example:

```
Line 1\
*@person*
```

[demo](https://marked.js.org/demo/?text=Line%201%5C%0A*%40person*&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=1.2.4)

fixes #1831 

## Contributor

## Committer


src/Lexer.js
```diff
@@ -328,12 +328,13 @@ module.exports = class Lexer {
   /**
    * Lexing/Compiling
    */
-  inlineTokens(src, tokens = [], inLink = false, inRawBlock = false, prevChar = '') {
+  inlineTokens(src, tokens = [], inLink = false, inRawBlock = false) {
```
UziTech, line 331: @calculuschild I don't think we actually ever send prevChar to this function. Is there a reason this was added to the function signiture in #1686?

I changed it to get prevChar from the tokens below.

calculuschild: No, it was likely an accidental remnant from an earlier commit I didn't fully clean up. It doesn't need to be there.

src/Lexer.js
```diff
@@ -388,6 +389,14 @@ module.exports = class Lexer {
         continue;
       }
 
+      // get previous text char
+      if (tokens.length > 0 && tokens[tokens.length - 1].type === 'text') {
+        lastTokenRaw = tokens[tokens.length - 1].raw;
+        prevChar = lastTokenRaw.charAt(lastTokenRaw.length - 1);
+      } else {
+        prevChar = '';
+      }
```
calculuschild, line 398: I wonder if there's a way to do this without introducing this extra processing of the tokens every loop?

What about something dumb like this with a flag `keepPrevChar`:

```
inlineTokens(src, tokens = [], inLink = false, inRawBlock = false) {
...
  let keepPrevChar, prevChar;
...

while (src) {
  if(!keepPrevChar) {
    prevChar = '';
  }
  keepPrevChar = false;

....

  if (token = this.tokenizer.inlineText(src, inRawBlock, smartypants)) {
    prevChar = token.raw.slice(-1);
    keepPrevChar = true;
    tokens.push(token);
    continue;
  }
...
}
}
```

styfle: @calculuschild Seems like a good solution 👍

UziTech: I updated it :+1:

calculuschild (approved):

styfle (approved):

PR #1851, opened 2020-11-26
fix: fix link with angle brackets around href

### LINKED ISSUE #1639: Link URL With Spaces Does Not Render
When a link with spaces in the URL it does render the element as a link.

Examples

- `[My Link](http://somesite.com/my path)`
- `[My Link](http://somesite.com/my file name.pdf)`

When I'm adding the markdown syntax I am now making sure I encode that href url before it gets rendered by marked.

UziTech:
**Marked version:** v1.2.5

## Description

Fixes CommonMark tests [486](https://spec.commonmark.org/0.29/#example-486), [489](https://spec.commonmark.org/0.29/#example-489), [490](https://spec.commonmark.org/0.29/#example-490), [491](https://spec.commonmark.org/0.29/#example-491)

- Fixes #1639

## Contributor

## Committer

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -489,8 +489,6 @@ module.exports = class Tokenizer {
         if (link) {
           href = link[1];
           title = link[3];
-        } else {
-          title = '';
         }
       } else {
         title = cap[3] ? cap[3].slice(1, -1) : '';
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -457,17 +457,33 @@ module.exports = class Tokenizer {
   link(src) {
     const cap = this.rules.inline.link.exec(src);
     if (cap) {
-      const lastParenIndex = findClosingBracket(cap[2], '()');
-      if (lastParenIndex > -1) {
-        const start = cap[0].indexOf('!') === 0 ? 5 : 4;
-        const linkLen = start + cap[1].length + lastParenIndex;
-        cap[2] = cap[2].substring(0, lastParenIndex);
-        cap[0] = cap[0].substring(0, linkLen).trim();
-        cap[3] = '';
+      const trimmedUrl = cap[2].trim();
+      if (!this.options.pedantic && trimmedUrl.startsWith('<')) {
+        // commonmark requires matching angle brackets
+        if (!trimmedUrl.endsWith('>')) {
+          return;
+        }
+
+        // ending angle bracket cannot be escaped
+        const rtrimSlash = rtrim(trimmedUrl.slice(0, -1), '\\');
+        if (!((trimmedUrl.length - rtrimSlash.length) & 1)) {
+          return;
+        }
+      } else {
+        // find closing parenthesis
+        const lastParenIndex = findClosingBracket(cap[2], '()');
+        if (lastParenIndex > -1) {
+          const start = cap[0].indexOf('!') === 0 ? 5 : 4;
+          const linkLen = start + cap[1].length + lastParenIndex;
+          cap[2] = cap[2].substring(0, lastParenIndex);
+          cap[0] = cap[0].substring(0, linkLen).trim();
+          cap[3] = '';
+        }
       }
       let href = cap[2];
       let title = '';
       if (this.options.pedantic) {
+        // split pedantic href and title
         const link = /^([^'"]*[^\s])\s+(['"])(.*)\2/.exec(href);
 
         if (link) {
@@ -479,7 +495,16 @@ module.exports = class Tokenizer {
       } else {
         title = cap[3] ? cap[3].slice(1, -1) : '';
       }
-      href = href.trim().replace(/^<([\s\S]*)>$/, '$1');
+
+      href = href.trim();
+      if (href.startsWith('<')) {
+        if (this.options.pedantic && !trimmedUrl.endsWith('>')) {
+          // pedantic allows starting angle bracket without ending angle bracket
+          href = href.slice(1);
+        } else {
+          href = href.slice(1, -1);
+        }
+      }
       const token = outputLink(cap, {
         href: href ? href.replace(this.rules.inline._escapes, '$1') : href,
         title: title ? title.replace(this.rules.inline._escapes, '$1') : title
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -505,11 +505,10 @@ module.exports = class Tokenizer {
           href = href.slice(1, -1);
         }
       }
-      const token = outputLink(cap, {
+      return outputLink(cap, {
         href: href ? href.replace(this.rules.inline._escapes, '$1') : href,
         title: title ? title.replace(this.rules.inline._escapes, '$1') : title
       }, cap[0]);
-      return token;
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/Tokenizer.js
```diff
@@ -457,17 +457,33 @@ module.exports = class Tokenizer {
...
+      const trimmedUrl = cap[2].trim();
+      if (!this.options.pedantic && trimmedUrl.startsWith('<')) {
+        // commonmark requires matching angle brackets
+        if (!trimmedUrl.endsWith('>')) {
+          return;
+        }
+
+        // ending angle bracket cannot be escaped
+        const rtrimSlash = rtrim(trimmedUrl.slice(0, -1), '\\');
+        if (!((trimmedUrl.length - rtrimSlash.length) & 1)) {
```
styfle, line 469: Is this intended to be the bitwise AND?

UziTech: yes. It is checking if the number of slashes removed is odd.

styfle: I see. 

Minor nit pick: Ideally, the `!` could be removed and the XOR operator could be used instead of AND.

Or maybe just modulo/remainder 2 which is what I would expect.

UziTech: I will do some benchmark testing for the modulo operator

UziTech: I didn't see any difference between using modulo or xor so I went with modulo.

styfle (approved):

src/Tokenizer.js
```diff
@@ -479,7 +495,16 @@ module.exports = class Tokenizer {
       } else {
         title = cap[3] ? cap[3].slice(1, -1) : '';
       }
-      href = href.trim().replace(/^<([\s\S]*)>$/, '$1');
+
+      href = href.trim();
+      if (href.startsWith('<')) {
+        if (this.options.pedantic && !trimmedUrl.endsWith('>')) {
```
calculuschild, line 501: There are multiple checks for the pedantic option. Is there some way all of the pedantic logic can be handled together?

src/Tokenizer.js
```diff
@@ -479,7 +495,16 @@ module.exports = class Tokenizer {
...
+          // pedantic allows starting angle bracket without ending angle bracket
+          href = href.slice(1);
+        } else {
+          href = href.slice(1, -1);
+        }
+      }
```
calculuschild, line 507: I liked the original way of doing this with the regex. My impression is regex is generally faster than manually checking characters (but I may be wrong).

Could the original regex be updated to this? (just add a `?` so it allows a missing end `>`):
`href = href.trim().replace(/^<([\s\S]*)>?$/, '$1');`

src/Tokenizer.js
UziTech: That would require duplicating a lot of the logic since most of the logic is the same.

src/Tokenizer.js
```diff
@@ -457,17 +457,33 @@ module.exports = class Tokenizer {
...
+        // commonmark requires matching angle brackets
+        if (!trimmedUrl.endsWith('>')) {
+          return;
+        }
+
+        // ending angle bracket cannot be escaped
+        const rtrimSlash = rtrim(trimmedUrl.slice(0, -1), '\\');
+        if ((trimmedUrl.length - rtrimSlash.length) % 2 === 0) {
+          return;
+        }
```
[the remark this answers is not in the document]
UziTech: That regex wouldn't catch an escaped slash before the escaped bracket (e.g. `\\\>`) and the regex to catch that is vulnerable to ReDOS. That is why we have the rtrim function instead of using replace with a regex.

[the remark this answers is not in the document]
calculuschild: Wouldn't the `[^\]` handle the case of the escaped slash? If there is an extra `\` then it won't match.

For my own understanding, can you help me with the Regex vulnerability of that pattern? `[^a]aa$`

src/Tokenizer.js
UziTech: regex is faster if there isn't [catastrophic backtracking](https://regular-expressions.mobi/catastrophic.html?wlr=1) that can allow ReDOS attacks. The problem with just using regex is that the end bracket can be escaped and as mentioned above that is vulnerable to ReDOS.

calculuschild: Hmm... I agree with the concern, but I'm not sure if it applies to this pattern. How does this lead to catastrophic backtracking? There is no nested repeating element in this regex, and no branching paths. I tested this just for fun on regex101.com, and even with thousands of characters it completes the match (or non-match) in 8 steps. I don't think this pattern *can* backtrack because there is only one path it can take from the `^` to the `$`. There is not any alternate `|` and all of the capturing takes place in a single, non-repeated group.

src/Tokenizer.js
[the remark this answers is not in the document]
UziTech: reversing the string would be much slower on big strings. I know this isn't as pretty as regex replace but it is the fastest way I could think of while still being simple.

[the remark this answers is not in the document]
calculuschild: If this seems like the fastest way, I'm ok with it. Just wanted to be sure.

src/Tokenizer.js
calculuschild: Meh, I missed your comment about the end bracket not being escaped, and that leads to the same problem with the regex above. Alright, non-regex it is. 👍

calculuschild:
The lines [`else {title = '';}`](https://github.com/markedjs/marked/pull/1851/files#diff-c6ae4e51c94d9a0253c6ca0afc670670f29049d741da446e6b19a22e4871821bR492-R494) (it's not letting me insert a comment directly) are redundant.

src/Tokenizer.js
```diff
@@ -479,7 +495,16 @@ module.exports = class Tokenizer {
...
+        if (this.options.pedantic && !trimmedUrl.endsWith('>')) {
+          // pedantic allows starting angle bracket without ending angle bracket
+          href = href.slice(1);
+        } else {
+          href = href.slice(1, -1);
+        }
+      }
       const token = outputLink(cap, {
         href: href ? href.replace(this.rules.inline._escapes, '$1') : href,
         title: title ? title.replace(this.rules.inline._escapes, '$1') : title
```
calculuschild, line 510: Can we return the token contents directly instead of creating the `const token`?

src/Tokenizer.js
calculuschild: You can replace the `(trimmedUrl.length - rtrimSlash.length) % 2 === 0)` with `(!(trimmedUrl.length - rtrimSlash.length) % 2))` which seems to be a *bit* faster: https://jsbench.me/lwkigdfqus/1

UziTech:
I fixed `return token` and `else { title = ''; }`

src/Tokenizer.js
calculuschild: See my PR on your branch.  Broke out my old boolean logic notes and reordered some things around. Not sure if it's "better", but it's more legible to me.

calculuschild (approved):

PR #1852, opened 2020-11-27
chore(demo): add links to other demos

UziTech:
## Description

Add links to the CommonMark and Daring Fireball demos from our demo page so it is easy to compare.

image

## Contributor

## Committer

### DIFF
--- a/docs/demo/demo.css
+++ b/docs/demo/demo.css
@@ -31,6 +31,10 @@ header h1 {
 	z-index: 1000;
 }
 
+.other-demos {
+	margin-left: 3em;
+}
+
 .containers {
 	display: flex;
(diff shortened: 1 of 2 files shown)


calculuschild (approved):
Good idea. I like it.

styfle (approved):
Sounds good, I think we could also make these links pass the markdown text too

UziTech:
@styfle we can pass it to commonmark but I don't think there is a way to pass it to daring fireball. I thought about that but I didn't think it was that difficult to just copy/paste if that is what the user wanted to do.

UziTech:
I'm just going to merge this as is and we can find a way to pass the markdown later if we want to.

Issue #1695, opened 2020-05-31
Transition the Lexer from using a hardcoded Ruleset to it iterating over an extendable set of Rules


KilianKilmister:
Currently, when you want to alter the ruleset of the lexer uses, you will have to replace the entirety of the `lexer.blockTokens` and `lexer.inline` functions.

The code used for checking a single rule takes up only a few lines but by the way it's coded, you will need to write out all the unaltered rules again aswell.

Every rule-check looks pretty much alike, for example:

```js
 // fences
 if ((token = this.tokenizer.fences(src))) {
    src = src.substring(token.raw.length)
    tokens.push(token)
    continue
}
```

It would be easy to convert it to iterate over an array of rule-objects and could also be startingpoint for using it asyncronously.

The API would remain the same and performance would stay the same or potentially even improve.

I intend to file a PR for it in the near future, as i need to alter the lexer for my usecase.

### EDIT

As it turnes out, the core if this package is incredibly solid and it's pretty much impossible to modify the parser and lexer from outside the source

UziTech:
If it would have the same performance, or better, I would be ok with it. :+1:

KilianKilmister:
Recently I had to work a lot with large tree structures and last-secon-evaluation and i was able to produce some great results using async generators. They are incredibly pluggable and will seemlessly pass data down the chain. I think they might produce some nice results.

I just forked the repo and i'll see what i can do

calculuschild:
This is one I definitely would like to see for v2.0.0 if it's not any slower. Is the plan to simply place each step of the Lexer into a list and then iterate over the list instead of the pre-determined for-loop we have now? Doesn't sound too hard.... right? Would we want user-provided tokenizers to always be added to the end of the list (maybe right before "text"), or would it make sense to give them an index of where to insert their tokenizer?

@KilianKilmister Did you have any progress with this you would like to contribute to a PR?

UziTech:
It would be nice if the extension would be able to choose where new tokenizers and renderers go in the order. I would like extensions to still be able to just return an object that is sent to `marked.use`. I'm not sure how that would work but maybe something like:

```js
marked.use({
  tokenizer: {
    latex: {
      before: "codespan",
      func: (src) => {
        // latex tokenizer
      },
    }
  }
});
```

calculuschild:
[...]

PR #1864, opened 2020-12-07
fix: Total rework of Emphasis/Strong

calculuschild:
## Description

- Fixes #1860, Fixes #1811 
- Also fixes Commonmark/GFM examples: 
  - (em & strong) 361, 387, 388, 407, 412, 415, 416, 424, 425, 442, 445, 446, 453, 455, 456, 457, 465, 466, 467, 470
  - This puts us up to 100% compatibility with commonmark specs!

- Noticeable speedup, especially on the GFM benchmark (~8.7 sec -> ~8.1 sec, pretty consistent over 5 runs on my laptop)

## What was attempted

1) Simplify regex for em & strong, combined now into a single tokenizer
2) When masking the `src` string in Lexer, also mask out escaped `\*` and `\_` which further simplifies a lot of regex
3) Track total opening delimiter characters vs closing characters, and ensure they match
4) More closely follow CommonMark spec:
   * Favor `<em><strong>text</strong></em>` over `<strong><em>text</em></strong>`
   * Correct some of the "New" spec tests that had this ^ swapped incorrectly (and delete one that is redundant now)
   * Handle em/strong CommonMark [rules 9-10](https://spec.commonmark.org/0.29/#can-open-emphasis), that left and right delimiters cannot sum to a multiple of 3, unless each is a multiple of 3
   * Handle cases with lots of extra unmatched delimiters, e.g. `*text*********`

Note this involves significant changes in the Lexer and Tokenizer APIs, which should be noted in the update.

The new Regex should be pretty benign compared to the earlier stuff. It literally checks for sequences of the pattern `a***b`, that is, runs of `*` or `_` between a single character on each side.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR); or,
- [x] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] CI is green (no forced merge required).
- [ ] Squash and Merge PR following [conventional commit guidelines](https://www.conventionalcommits.org/).

PR #1872, opened 2020-12-13
Rework Lexer to use extendable array of tokenizer functions

calculuschild:
## Description

An attempt at #1695. Not sure about speed or elegance here, but wanted some feedback to see if this is even a reasonable route to take. If so, I would appreciate some troubleshooting to get this cleaner and to fix the issue with the broken test case.

Ideally, this would allow users to extend the Lexer by plugging in custom tokenizers at a chosen space in the lexer pipeline, and the `params` object exposes all of the required parameters to make functions with different signatures work.

**Notes:**
* I only included the block tokenizers here to start.
* This seems to break something with the `def` tokenizer. I don't fully understand what, but it makes Commonmark #187 fail.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -220,186 +220,155 @@ module.exports = class Lexer {
    */
 
   // newline
-  newline(params) {
+  newline(src, params) {
     let token;
-    if (token = this.tokenizer.space(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      if (token.type) {
-        params.tokens.push(token);
-      }
-      return true;
+    if (token = this.tokenizer.space(src)) {
+      return token;
     }
   }
 
   // code
-  code(params) {
+  code(src, params) {
     let token;
-    if (token = this.tokenizer.code(params.src, params.tokens)) {
-      params.src = params.src.substring(token.raw.length);
-      if (token.type) {
-        params.tokens.push(token);
-      } else {
-        params.lastToken = params.tokens[params.tokens.length - 1];
-        params.lastToken.raw += '\n' + token.raw;
-        params.lastToken.text += '\n' + token.text;
-      }
-      return true;
+    if (token = this.tokenizer.code(src, params.lastToken)) {
+      return token;
     }
   }
 
   // fences
-  fences(params) {
+  fences(src, params) {
     let token;
-    if (token = this.tokenizer.fences(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.fences(src)) {
+      return token;
     }
   }
 
   // table no leading pipe (gfm)
-  nptable(params) {
+  nptable(src, params) {
     let token;
-    if (token = this.tokenizer.nptable(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.nptable(src)) {
+      return token;
     }
   }
 
   // heading
-  heading(params) {
+  heading(src, params) {
     let token;
-    if (token = this.tokenizer.heading(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.heading(src)) {
+      return token;
     }
   }
 
   // hr
-  hr(params) {
+  hr(src, params) {
     let token;
-    if (token = this.tokenizer.hr(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.hr(src)) {
+      return token;
     }
   }
 
   // blockquote
-  blockquote(params) {
+  blockquote(src, params) {
     let token;
-    if (token = this.tokenizer.blockquote(params.src)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.blockquote(src)) {
       token.tokens = this.blockTokens(token.text, [], params.top);
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // list
-  list(params) {
+  list(src, params) {
     let token;
-    if (token = this.tokenizer.list(params.src)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.list(src)) {
       const l = token.items.length;
       let i;
       for (i = 0; i < l; i++) {
         token.items[i].tokens = this.blockTokens(token.items[i].text, [], false);
       }
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // html
-  html(params) {
+  html(src, params) {
     let token;
-    if (token = this.tokenizer.html(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.html(src)) {
+      return token;
     }
   }
 
   // def
-  def(params) {
+  def(src, params) {
     let token;
-    if (params.top && (token = this.tokenizer.def(params.src))) {
-      params.src = params.src.substring(token.raw.length);
+    if (params.top && (token = this.tokenizer.def(src))) {
       if (!this.tokens.links[token.tag]) {
         this.tokens.links[token.tag] = {
           href: token.href,
           title: token.title
         };
       }
-      return true;
+      return token;
     }
   }
 
   // table (gfm)
-  table(params) {
+  table(src, params) {
     let token;
-    if (token = this.tokenizer.table(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.table(src)) {
+      return token;
     }
   }
 
   // lheading
-  lheading(params) {
+  lheading(src, params) {
     let token;
-    if (token = this.tokenizer.lheading(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.lheading(src, params)) {
+      return token;
     }
   }
 
   // top-level paragraph
-  paragraph(params) {
+  paragraph(src, params) {
     let token;
-    if (params.top && (token = this.tokenizer.paragraph(params.src))) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (params.top && (token = this.tokenizer.paragraph(src))) {
+      return token;
     }
   }
 
   // text
-  text(params) {
+  text(src, params) {
     let token;
-    if (token = this.tokenizer.text(params.src, params.tokens)) {
-      params.src = params.src.substring(token.raw.length);
-      if (token.type) {
-        params.tokens.push(token);
-      } else {
-        params.lastToken = params.tokens[params.tokens.length - 1];
-        params.lastToken.raw += '\n' + token.raw;
-        params.lastToken.text += '\n' + token.text;
-      }
-      return true;
+    if (token = this.tokenizer.text(src, params.lastToken)) {
+      return token;
     }
   }
 
   blockTokens(src, tokens = [], top = true) {
+    let token;
     src = src.replace(/^ +$/gm, '');
 
     const blockParams = {
-      src: src,
-      tokens: tokens,
       top: top,
-      // token: null,
-      lastToken: null//,
+      lastToken: null
     };
 
-    while (blockParams.src) {
-      if (this.blockTokenizers.some(fn => fn.func.call(this, blockParams))) continue;
+    while (src) {
+      if (this.blockTokenizers.some(fn => token = fn.func.call(this, src, blockParams))) {
+        src = src.substring(token.raw.length);
+        if (token.type) {
+          if (token.type === 'continue') {
+            blockParams.lastToken.raw += '\n' + token.raw;
+            blockParams.lastToken.text += '\n' + token.text;
+          } else {
+            tokens.push(token);
+            blockParams.lastToken = token;
+          }
+        }
+        continue;
+      }
 
-      if (blockParams.src) {
-        const errMsg = 'Infinite loop on byte: ' + blockParams.src.charCodeAt(0);
+      if (src) {
+        const errMsg = 'Infinite loop on byte: ' + src.charCodeAt(0);
         if (this.options.silent) {
           console.error(errMsg);
           break;
@@ -409,7 +378,7 @@ module.exports = class Lexer {
       }
     }
 
-    return blockParams.tokens;
+    return tokens;
   }
 
   inline(tokens) {
@@ -478,135 +447,111 @@ module.exports = class Lexer {
   }
 
   // escape
-  escape(params) {
+  escape(src, params) {
     let token;
-    if (token = this.tokenizer.escape(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.escape(src)) {
+      return token;
     }
   }
 
   // tag
-  tag(params) {
+  tag(src, params) {
     let token;
-    if (token = this.tokenizer.tag(params.src, params.inLink, params.inRawBlock)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.tag(src, params.inLink, params.inRawBlock)) {
       params.inLink = token.inLink;
       params.inRawBlock = token.inRawBlock;
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // link
-  link(params) {
+  link(src, params) {
     let token;
-    if (token = this.tokenizer.link(params.src)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.link(src)) {
       if (token.type === 'link') {
         token.tokens = this.inlineTokens(token.text, [], true, params.inRawBlock);
       }
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // reflink, nolink
-  reflink(params) {
+  reflink(src, params) {
     let token;
-    if (token = this.tokenizer.reflink(params.src, this.tokens.links)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.reflink(src, this.tokens.links)) {
       if (token.type === 'link') {
         token.tokens = this.inlineTokens(token.text, [], true, params.inRawBlock);
       }
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // strong
-  strong(params) {
+  strong(src, params) {
     let token;
-    if (token = this.tokenizer.strong(params.src, params.maskedSrc, params.prevChar)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.strong(src, params.maskedSrc, params.prevChar)) {
       token.tokens = this.inlineTokens(token.text, [], params.inLink, params.inRawBlock);
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // em
-  em(params) {
+  em(src, params) {
     let token;
-    if (token = this.tokenizer.em(params.src, params.maskedSrc, params.prevChar)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.em(src, params.maskedSrc, params.prevChar)) {
       token.tokens = this.inlineTokens(token.text, [], params.inLink, params.inRawBlock);
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // code
-  codespan(params) {
+  codespan(src, params) {
     let token;
-    if (token = this.tokenizer.codespan(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.codespan(src)) {
+      return token;
     }
   }
 
   // br
-  br(params) {
+  br(src, params) {
     let token;
-    if (token = this.tokenizer.br(params.src)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.br(src)) {
+      return token;
     }
   }
 
   // del (gfm)
-  del(params) {
+  del(src, params) {
     let token;
-    if (token = this.tokenizer.del(params.src)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.del(src)) {
       token.tokens = this.inlineTokens(token.text, [], params.inLink, params.inRawBlock);
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
   // autolink
-  autolink(params) {
+  autolink(src, params) {
     let token;
-    if (token = this.tokenizer.autolink(params.src, params.mangle)) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (token = this.tokenizer.autolink(src, params.mangle)) {
+      return token;
     }
   }
 
   // url (gfm)
-  url(params) {
+  url(src, params) {
     let token;
-    if (!params.inLink && (token = this.tokenizer.url(params.src, params.mangle))) {
-      params.src = params.src.substring(token.raw.length);
-      params.tokens.push(token);
-      return true;
+    if (!params.inLink && (token = this.tokenizer.url(src, params.mangle))) {
+      return token;
     }
   }
 
   // text
-  inlineText(params) {
+  inlineText(src, params) {
     let token;
-    if (token = this.tokenizer.inlineText(params.src, params.inRawBlock, params.smartypants)) {
-      params.src = params.src.substring(token.raw.length);
+    if (token = this.tokenizer.inlineText(src, params.inRawBlock, params.smartypants)) {
       params.prevChar = token.raw.slice(-1);
       params.keepPrevChar = true;
-      params.tokens.push(token);
-      return true;
+      return token;
     }
   }
 
@@ -614,7 +559,7 @@ module.exports = class Lexer {
    * Lexing/Compiling
    */
   inlineTokens(src, tokens = [], inLink = false, inRawBlock = false) {
-    let match;
+    let match, token;
     let maskedSrc = src;
 
     // Mask out reflinks to avoid interference with em and strong
@@ -634,8 +579,6 @@ module.exports = class Lexer {
     }
 
     const inlineParams = {
-      src: src,
-      tokens: tokens,
       inLink: inLink,
       inRawBlock: inRawBlock,
       maskedSrc: maskedSrc,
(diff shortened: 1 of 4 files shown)


src/Lexer.js
```diff
@@ -73,6 +73,22 @@ module.exports = class Lexer {
       }
     }
     this.tokenizer.rules = rules;
+
+    this.blockTokenizers = [
+      this.newline,
```
UziTech, line 78: We need some way to get the name of the function so we can add new functions before and after certain ones.

Maybe something like:

```suggestion
      { name: "newline", func: this.newline},
```

calculuschild: Ok. What do you think about using a [Map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of an Array of one-element objects? It requires es6 but keeps insertion order for iterability. I have no idea about speed.

UziTech: Sounds good. We won't support ie11 by the time this gets merged anyway.

styfle: > We won't support ie11 by the time this gets merged anyway.

When are we expecting to merge this? Are we planning to do a major semver soon?

UziTech: Ya we have to bump the major version for #1864 anyway I figured we could add a few other breaking changes. The [v2](https://github.com/markedjs/marked/projects/7) project has a few things we could get in to v2. I'm not sure if we will get all of them in but I figured we could seriously look into them and if they are viable in a reasonable amount of time we could release v2 in a month or two.

calculuschild:
Anyone want to take a whack at getting CommonMark #187 working again with this? I'm hitting a wall.

src/Lexer.js
```diff
@@ -119,135 +135,191 @@ module.exports = class Lexer {
...
+       return true;
+     }
+   }//.bind(this);
+
+   // def
+   def(params) {
+     if (params.top && (params.token = params.tokenizer.def(params.src))) {
+       params.src = params.src.substring(params.token.raw.length);
+       if (params.tokens && (!params.tokens.links || !params.tokens.links[params.token.tag])) {
+         params.tokens.links[params.token.tag] = {
```
UziTech, line 239:
```suggestion
       if (!this.tokens.links[params.token.tag]) {
         this.tokens.links[params.token.tag] = {
```

src/Lexer.js
```diff
@@ -119,135 +135,191 @@ module.exports = class Lexer {
...
-      // lheading
-      if (token = this.tokenizer.lheading(src)) {
-        src = src.substring(token.raw.length);
-        tokens.push(token);
-        continue;
-      }
+  blockTokens(src, tokens = [], top = true) {
+    if(!tokens.links){
+      tokens.links = Object.create(null);
+    }
```
UziTech, line 295:
```suggestion
```

src/Lexer.js
```diff
@@ -119,135 +135,191 @@ module.exports = class Lexer {
...
-      if (token = this.tokenizer.text(src, tokens)) {
-        src = src.substring(token.raw.length);
-        if (token.type) {
-          tokens.push(token);
-        } else {
-          lastToken = tokens[tokens.length - 1];
-          lastToken.raw += '\n' + token.raw;
-          lastToken.text += '\n' + token.text;
+      for(fn of this.blockTokenizers) {
+        if(fn(blockParams)) {
```
UziTech, line 316:
```suggestion
        if(fn.call(this, blockParams)) {
```

UziTech (changes requested):
`links` should only be on the top level tokens array

calculuschild:
Alright, now it's just the Linter getting mad at my use of a label to break out of nested loops. Can we give a pass to the linter on this one or is there a good alternative?

src/Lexer.js
calculuschild: Ahhhhh.... CALL() is what I was looking for but didn't know what to search for. Now I know it's a thing. That solved it!

calculuschild:
Oof, now that it's all converted over the benchmarks are not looking too hot on this one....

UziTech:
I think a jump to an outer loop is fine to disable the linter for that line.

This definitely needs some work to get it faster before merging. I'll take a look at it soon.

calculuschild:
> This definitely needs some work to get it faster before merging. I'll take a look at it soon.

Is there any good Javascript profiling tool that might easily work with Marked?

Monkatraz:
> Alright, now it's just the Linter getting mad at my use of a label to break out of nested loops. Can we give a pass to the linter on this one or is there a good alternative?

I found this:
```js
if (this.inlineTokenizers.some(fn => fn.func.call(this, inlineParams))) continue;
```
On my system this has the same performance as the original loop. The performance of various methods is bizarre - if I create a bound (`func.bind(this, inlineParams)`) array and call those directly, it's actually slower.

calculuschild:
@Monkatraz Using `array.some` does seem to speed up a bit. Thanks for the suggestion!

We are getting there....

```
CommonMark  : 3878 (ms)
Markdown-it : 3726

Marked Commonmark es6
-----------------------------------
Current Master              : 4394


Previous best               : 4720
                               ↓
Array.some()                : 4569
                               ↓
Remove i, l, from params    : 4510
```

calculuschild:
Ok, so by moving the `src`, and `tokens` out of the params object (slight refactoring), we minimize the amount of object mutation which is somewhat slower than direct variable access. I think this might be right about at the performance of the current master; I can't see much difference in the benchmarks on my machine. Do you guys care to give it a test?

Also, lint doesn't like assigning variables inside the arrow function, but I believe arrow functions are slightly faster than traditional anonymous functions?

calculuschild:
Ooh, so `array.some()` has a `thisArg` parameter, so you don't need to bind the function to `this` after all. This seems quite a bit faster now.

UziTech:
testing on my machine shows this is about 10% slower than master. I'll try to go through it this weekend and see if we can speed anything up.

calculuschild:
Just had a thought: Ever since the Tokenizer.js was pulled out from the Lexer, the Lexer has just become a wrapper to direct the sequence of the various tokenizer functions. After stripping out all of the common code from the Lexer functions (consume src, push token onto the stack), the Lexer functions essentially just an intermediate empty function that calls another function. What if we refactor the Lexer.js to just call the Tokenizer directly? Instead of
```
blockTokens(src, tokens = [], top = true) {
   while(src) {   
      token = LexerStep(params);
      tokens.push(token);
   }

LexerStep(params){
  token = Tokenizer(src);
  return token
}
```

we do this:
```
blockTokens(src, tokens = [], top = true) {
   while(src) {   
      token = Tokenizer(params);
      tokens.push(token);
   }
```

That cuts out a full unnecessary function call for each token.

OR.... maybe it would make sense to merge the tokenizer and lexer back together but leave as functions instead of the inline tokens we had before? I know that undoes some work from before, but if the intention of that change was to make extending Marked.js easier, maybe we inadvertently added unneeded complexity that actually makes extension harder.

UziTech:
The Tokenizer class was supposed to be a way extend the Lexer the way they extend the Parser with the Renderer class.

It is more a separation of concerns. The Tokenizer function (the function that can be extended) should only be concerned with `Is this the next token in src and what are the properties`. The Lexer handles when to call those function like `Only check for paragraphs if this is the top`, or `Get inline tokens if reflink token is type "link"`

Similarly Renderer functions are only concerned with `Given these properties, what HTML should be output`. The Parser handles which properties to send to which renderer function.

calculuschild:
> The Lexer handles when to call those function

Exactly, which is what prompted my thought. In the current master, we have a loop that directly calls all the tokenizers in sequence until a match is found, handles the result, and then consumes the src and starts the loop over. What I ended up doing with this PR, which is probably where most of the slowdown comes from, was calling an array of intermediate functions in sequence which *then* calls the Tokenizer. Do you think there is some way we could refactor this PR to directly call the tokenizers from an array rather than this array of intermediate functions? 90% of the tokens are literally just "get token from Tokenizer.js, push token to tokens array" and don't need that intermediate function, but like you said, there are a few odd cases where the tokens have further processing (internal tokens, is this the top level, etc.) that are tripping me up.

calculuschild:
Ah wait, what if.... What if this array of lexer functions can include Tokenizer functions directly? Then for the 90% of tokens that need no further processing, we just directly put into our array of functions:
```
this.blockTokenizers = [
      { name: 'newline', func: this.tokenizer.space },
      { name: 'code', func: this.tokenizer.code },
     ...
```

And in the special cases where the lexer needs to step in and apply further processing, use a wrapper lexer function:
```
     ...
    { name: 'paragraph', func: this.paragraph },
    ];
```

This PR already makes the Lexer functions return tokens, but would skip the extra function call on 90% of the token types.

calculuschild:
Huh.... I don't know why, but calling the tokenizers directly seems to make things even slower. I have no idea what's going on.

Replacing `func.call(this, params)` with binding the functions ahead of time and just calling `func(params)` is suspiciously slower as well, like @Monkatraz pointed out, even though in JSBench executing a bound function should be much, much faster. https://jsbench.me/yekjtl7lje/1

Monkatraz:
It'll take forever to test - but one stupid idea is to pass `this` as an argument to the tokenizer functions. I can't imagine that not being the fastest possible approach, and it shouldn't have any overhead.

calculuschild:
> pass `this` as an argument to the tokenizer functions

Do you mean just use `func(this, params)` instead of `func.call(this, params)`?

Monkatraz:
> Do you mean just use `func(this, params)` instead of `func.call(this, params)`?

Yep, exactly. Having written `markdown-it` plugins, it basically does exactly this, so it's not unusual to do it that way.

calculuschild:
@Monkatraz  > Yep, exactly.

Well, it works... but I don't really see any speed increase. I don't know how much I trust the benchmark results on my laptop though. Pushed the commit in case it actually has an impact.

Monkatraz:
> Well, it works... but I don't really see any speed increase. I don't know how much I trust the benchmark results on my laptop though. Pushed the commit in case it actually has an impact.

I'm really not surprised, but I _think_ this means that you can't really get this any faster than it is currently. (at least with just messing with the arrays) I mean, you're just straight up calling the function with no manipulation applied to it.

I'm very confused on why bound functions are so much slower - I suspect either something is wrong with the benchmark or there is very low level shenanigans happening that will take forever to figure out.

Monkatraz:
I have no clue what's happening.

I've benched both versions - at first, the `this` pass version was slower, but then it was faster, and then it was slower again. The ES5 pedantic bench varied by a whole second, and I didn't even change which version I was benching. It just did that by itself.

I don't even know how you debug this - I know Firefox Developer Edition comes with a built in profiler utility, and it can get pretty detailed - maybe try rendering on the browser??? This is spooky.

calculuschild:
> I wonder if using .apply() is basically just like passing this as an argument internally

I think it is basically that, but with the added step of needing to also call `apply` (or `call`), so I would think passing this as an argument directly would be faster.

In any case, I wonder if it would make sense to upgrade the benchmarking script to use something like `benchmark.js` which supposedly is less prone to random noise in the timing data. Just getting consistent results on that would answer a lot of questions on this type of PR.

UziTech:
We will also need to figure out a way to prevent the `text` and `inlineText` tokenizers from consuming too much of the text.

For example, if I want to create a token for the letter `a` we would need to add a token that looks for the letter `a` at the beginning of `src` and also stops the `inlineText` tokenizer from consuming the letter `a`.

calculuschild:
What do they do now? Does the regex rule just consume everything until it reaches one of the "special" characters that could possibly signal a code block? Would it be possible to intelligently... read the preceding regex rules and extract the first character of each? and then we just replace the `text` rule with a list of those "special characters" to stop at? Or... sigh... require pushing the "start" character of your custom rule onto that list manually?

Ah... so the block `text` just eats everything. I guess Blocks have to start at the beginning of a line so there's not so much conflict; as long as your custom `a` block token starts a line you would be fine?

But the `inlineText` does need a special character to stop it. Oh. It is a nasty regex (especially GFM). I don't know how easy that would be to programmatically edit.

Making `inlineText` consume just one character each time would maybe solve that problem, but it would probably be very slow looping all the way through the lexer for each individual character.

calculuschild:
Hey @UziTech , are you aware of any JavaScript method of.... "precompiling" a set of code programmatically? Say, taking this array of functions and extracting their contents, inlining them in a master function (perhaps as part of the "use()" function), and then using that precompiled function rather than calling the original set of functions? I'm thinking similar to the way Babel transpiles different versions of JavaScript?

UziTech:
The only thing I know of that might do something like that is [Closure Compiler](https://github.com/google/closure-compiler-npm/)

UziTech:
I think we should add a way to call a function that is potentially the start of the token and cycle through those functions in `inlineText`.

You can see how I had to do it in the [marked-linkify-it extension](https://github.com/UziTech/marked-linkify-it/blob/main/src/index.js#L40-L46).

calculuschild:
@UziTech Please take a look at #1914 . I'm trying some experimental options that on my end seem like they solve the speed issue, but it does require the static lexer from #1909 or the overhead drops speed to about 5 times slower. I put it on a separate branch just in case it doesn't look feasible so as not to muck up this PR too much.

Monkatraz:
If the function recompilation idea doesn't work out, what about a generator function? I think a generator function could reduce on overhead by avoiding the usage of callbacks (like with the current `some` call). I'm thinking of something like:
```js
// imagine this is declared when the lexer is made or compiled or whatever
// it runs a tokenizer function (something returning a token if successful) and returns the token
// however, and most importantly, if no token was returned it simply yields. 
// hopefully being native, this lets the JS compiler optimize this?
const createTokenizerIterator = function* { ... }
const tokenizer = createTokenizeIterator()

while (src) {
  // if it is done before the end of the list, that means a token was returned
  // otherwise, done will be true and the value will be undefined
  while (!tokenizer.done) tokenizer.next()
  
  if (tokenizer.value) {
    src = src.substring(tokenizer.value.raw.length)
    tokens.push(tokenizer.value)
    // passing a value into `.next()` could allow you to reset the list position of the tokenizer to the start
    tokenizer.next(false)
    continue
  }
  // error out past here, like normal
}
```

Another potential optimization is to use some sort of linked list instead of an array of functions - that way you don't have to actually crawl through the list (in memory, that is) to find the function you want to use. Something like this:

```ts
interface TokenizerListItem {
  fn: TokenizerFunction,
  next: TokenizerListItem
}
```

You could see how this would be used with a generator that preserves its state - you set a variable inside the generator to the next item, and on the next execution call its function and move onto its next item.

I'm conflicted if this is actually faster, but my gut says it is. Storing objects in memory like this seems like it would have less overhead.

styfle:
> what about a generator function ... very short and sweet in code size

It's short and sweet for ES6 but the emitted ES5 will likely grow much larger.

Monkatraz:
> It's short and sweet for ES6 but the emitted ES5 will likely grow much larger.

I suspect more environments will support generator functions than not, so imo it's probably worth the trade, although only if it actually improves performance on ES6.

UziTech:
> It's short and sweet for ES6 but the emitted ES5 will likely grow much larger.

I think we are removing the ES5 version in v2 by the time this gets merged anyway.

calculuschild:
Now, we could go the other way... Instead of moving every step of the lexer out into an array of functions, what if we just leave the default functions as the are, but add checks for the existence of custom functions and execute them instead if they do. Something like this (seems to work well enough):

```
// Custom Newline Function (inserted via Marked.use() )
newLine(src) {
  let token;
  console.log("what");
  if(token = this.tokenizer.space(src)) {
    return token;
  }
}
```

And then in Lexer.js:
```
// newline
if(this.newLine && (token = this.newLine(src)) <-- if custom newline Tokenizer exists, execute instead
|| (token = this.tokenizer.space(src))) {
  if (token.type) {
    tokens.push(token);
  }
  continue;
}

if(this.newLine_after && (token = this.newLine_after(src))) { <-- if custom Tokenizer inserted after newline, execute
  if (token.type) {
    tokens.push(token);
  }
  continue;
}
```

It's clunky.... but it at least the original performance is mostly unaffected.

Unfortunately I'm at my wit's end here trying to get this to work with the original performance, so I'm going to need to take a break from this PR for a bit. @Monkatraz If you are interested in testing out the generator function I'll gladly let you give it a shot. 👍

Monkatraz:
Aight, if I suddenly gain the Awake At 3AM Urge to take a crack at this, I'll try it. Hopefully it works... lol.

calculuschild:
> Aight, if I suddenly gain the Awake At 3AM Urge to take a crack at this, I'll try it. Hopefully it works... lol.

Note that any overhead due to `const createTokenizerIterator = function* { ... }` in the constructor or whatever is going to occur with each individual call to `marked()` and turned out to be the reason why `bind` was slowing things down so much. You may want to build off of #1909 so the setup only happens once and the Lexer object persists between calls in Bench.js.

calculuschild:
Well, it's been a while now and I haven't had time to really look at this for months. I feel like we might need to nail down a better structure for how we expect extensions to be made, which would help direct how some of the code in this PR is laid out. Rant incoming. Sorry for the wall of text.

Right now, a user can take a Renderer or Tokenizer object and edit its sub-functions, but there isn't a clear way to add a NEW sub function to handle custom Markdown syntax. The user has to instead rely on intercepting and overriding, say, all the Text tokens before they get rendered and adding your custom handling in the tokenizer there. The user then has to copy/paste the existing tokenizer code in below their custom code to make sure the original functionality also works.

It would be great if there was a clear format for how these extensions are meant to be designed so as to cause minimal interference with other extensions and ease of maintenance.

Say, for simplicity, a user wants to add a custom Markdown symbol `:` at the beginning of the line. If a line starts with `:` the whole line needs to be underlined, generating a `<u> ... </u>` until the next newline character. Going in and editing the `text` tokens isn't obvious, because the user doesn't want to change the way text is rendered, but rather add an entire new token. But say he does that and hijacks the text tokenizer or text renderer to insert `<u>` when it detects `^:\s`, and then publishes that as an NPM package. How should the NPM package be consumed by Marked.js? I'm assuming we want packages to work in a way that allows `marked.use(customNPMpackage)` to be all that's necessary.

But, then, user2 comes along and makes another NPM packages that takes `{` and the beginning of a line and makes it surround the line with a span. Now we have two NPM packages that function by hijacking the same Tokenizer function, and seem like they will overwrite each other if you try to use both.

Not to mention, every time the Text tokenizer is updated here, extension maintainers need to copy and re-paste the code into their extension. I know the Text tokenizer isn't exactly complex, but some other Tokenizers depend on multiple helper functions that the maintainer now has to track down and duplicate in their extension.

So, with speed being kind of the priority of Marked.js, we need to decide which extensibility options we actually want to provide our users, because as we have seen in this PR, certain approaches tend to result in a lot of slowdown, and adding new features later only gets more complicated as Marked.js is pretty tightly interconnected between all its parts.

To start, I'm imagining something like this:

1) We provide the list of all tokens we currently support, and the order in which they are handled so users know where in the process their custom token needs to appear. This is already in the Docs but the order might not be correct anymore.

3) We provide an example Tokenizer function users can modify which takes input source text, and spits out a token object. Users currently can only modify the existing tokenizers.

5) We allow the users to specify *where in the processing order* their custom tokenizer function executes, which essentially sets their priority (i.e., the user has a custom "table" that needs to be checked before defaulting to the normal table token if it isn't a match). Users also specify if the token is block-level or inline.

4) We provide an example Renderer function users can take an input token and generate an HTML string as output. Users can currently only modify the existing renderers. The Parser.js can just handle all of these at the end before the `default` error message. I don't think processing order matters at this point, the resulting HTML is the same. Also the renderer here will handle all of the token processing, basically performing the function of both `parser.js` and `renderer.js`.

5) All of this related data is organized into an object that Marked.js consumes via `marked.use()` and references later on as it executes.

The extension for new custom Markdown looks like this maybe? When `marked.use` detects this format, we log it to some `customMarkdown` object somewhere rather than merging with and overwriting the existing functions. Then users can potentially add multiple custom extensions in this way without worrying about conflicts.

```
// Custom Markdown
const underline = {
  before : 'paragraph', // Leave blank to run after everything else...?
  level   : 'block',
  tokenizer : (src)=> {
    const rule = /^:.*$/;
    const match = rule.exec(src);
    return {
        type: 'underline',
        raw: match[0],
        text: match[0].trim()
      };
  },
  renderer : (token)=> {
    return `<u>${token.text}</u>`;
  }
};

marked.use({ underline });
```

Otherwise, the current method for hijacking existing tokenizers/renderers can remain the same.

This is probably missing some key points but if we can narrow down how we want extensions like this to work we can build around that rather than fumbling around here like I was doing in this PR.

UziTech:
That looks good. I would change a few things so we don't have a breaking changes with the current way `use` works.

We should add a new property `extensions` (or something else) that `underline` is added to.

`underline` would also have to have a `name` property or some way to know which renderer to call for the `underline` token type.

```js
marked.use({
  extensions: [
    underline,
  ]
});
```

UziTech:
That would work for the block extension but we would need to figure out how to prevent the `inlineText` tokenizer from consuming the start of an inline extension.

I think we would have to give some function that returns the next potential start of the inline token so inlineText can only consume up to that point then try all the inline tokenizers again.

For example if someone wanted to do an inline latex extension instead of overriding codespan:

```js
const latex = {
  before : 'codespan',
  level   : 'inline',
  start : (src) => src.indexOf("$"),
  tokenizer : (src)=> {
    const match = src.match(/^\$+([^\$\n]+?)\$+/);
    if (match) {
      return {
        type: 'codespan',
        raw: match[0],
        text: match[1].trim()
      };
    }
  },
};

marked.use({ latex });
```

Then we know to end the inlineText tokenizer at or before that `start` index.

calculuschild:
Let me make sure I understand your `start` function.

Currently `InlineText` consumes everything that we know can't be the start of an inline Token using a big RegEx, right? But now, we also make sure `InlineText` doesn't accidentally consume part of the custom token. So `start` returns the index of the next character to check for the custom token, and `InlineText` will work as normal, but only up to that index. Then we check again for the custom token, and if none is found, get a new index from `start`, and again consume up to that index, following the RegEx along the way.

So `InlineText` will consume in steps instead of all at once, stopping at the nearest index we get from any `start` functions in our `Extensions` object. Is that right? Yeah... And if no extensions are used, that should allow `InlineText` to run without any slowdowns for normal users.

calculuschild:
If this format is good, I think we can just redo this PR and simplify it down to just inserting a check between each Lexer step as mentioned above. None of this crazy arrays of tokenizers and binding/calling functions that would look cleaner, but unfortunately just slows it all down.

```
//<- Code Tokenizer

...

//Check for custom Tokenizer    <- duplicate and paste this code in between EVERY TOKENIZER in lexer.js.
if(this.fences_before && (token = this.fences_before(src))) { <-- if custom Tokenizer inserted before Fences, execute
  tokens.push(token);
  continue;
}

...

//Fences Tokenizer here ->
```

Although we probably need to account for multiple custom tokenizers inserted at the same location. So we would have to loop over the `extensions` object and execute each Tokenizer with `before : 'fences'`, in the order they were provided by the user.

Hm... Maybe we can make this `check for custom tokenizer` bit a function that intelligently updates what position it is at with each call so we can just insert a one-liner at each spot and the function knows if we are at `before : fences` or if we are now at `before : code` or wherever.

UziTech:
Pretty much. The way I got my [`marked-linkify-it`](https://github.com/UziTech/marked-linkify-it) extension working was to override `inlineText` with the following tokenizer:

```js
inlineText(src, ...args) {
  // find the next start index
  const match = linkify.match(src);
  if (match && match.length > 0) {
    // get `src` up to that index
    src = src.substring(0, match[0].index);
  }

	// run `inlineText` on the string up to that index
  return inlineText.call(this, src, ...args);
}
```

Here I find the next start index then truncate `src` and only run the `inlineText` tokenizer on that truncated string.

In my latex example the `start` function wouldn't necessarily return the actual start of the next custom token but just a potential start just to make the processing faster. Once `inlineText` found a next potential start of a token it would run `src` back through all of the inline tokens including the custom ones. We would have to make sure `start` returns a number greater than 0 since if that is the actual start the custom tokenizer should have caught it before getting to `inlineText`.

calculuschild:
How do we want to go about testing extensions? Do we have any other simple ones like your `marked-linkify-it` that we know of that we can plug in? I have some pretty ugly ones I hacked together for my own use but I'd need to rewrite them for this format.

I'm also not as versed in Javascript unit testing so adding an automated way to test extensions might need to fall to someone else. But for now I can manually plug things in and see if they work as this update is being built.

UziTech:
I don't know of any other extensions that change the tokenizers. I couldn't find any with a quick search of npm. Most extensions just change a renderer.

We could create some extensions for [Extended Markdown](https://www.markdownguide.org/extended-syntax/). Some of them, like [heading ids](https://www.markdownguide.org/extended-syntax/#heading-ids), should be pretty simple.

UziTech:
[Definition Lists](https://www.markdownguide.org/extended-syntax/#definition-lists) looks like a good example that could use this functionality.

calculuschild:
As I'm working on this, I may have run into an issue with custom block tokenizers. The problem is this: Say I'm using the example from above to underline any line starting with `:`, and I want this to behave as a block:

```
// Custom Markdown
const underline = {
  before : 'paragraph', // Leave blank to run after everything else...?
  level   : 'block',
  tokenizer : (src)=> {
    const rule = /^:.*$/;
    const match = rule.exec(src);
    if(match) {
      return {
          type: 'underline',
          raw: match[0],        //This is the text that you want your token to consume from the source
          text: match[0].trim() //You can add additional properties to your tokens to pass along to the renderer
      };
    }
  },
  renderer : (token)=> {
    return `<u>${token.text}</u>`;
  }
};

marked.use({ underline });
```

If I use an input of

```
No underline
: Underline
No Underline
```

The default `paragraph` regex in rules.js will always consume the whole block without ever checking the custom extension since it relies on explicit rules to interrupt the paragraph.

```
_paragraph: /^([^\n]+(?:\n(?!hr|heading|lheading|blockquote|fences|list|html| +\n)[^\n]+)*)/,
```

Does this mean any custom tokenizer will also need to override the default paragraph tokenizer? That could get very messy..... Is this something we can treat similarly to the `inlineText` tokenizer as you did above @UziTech ? I'm having trouble picturing how that should work.

UziTech:
I think that is a problem we are going to run into. We could do the same as the inline text. Basically the extension would give a start index and the other tokenizers would only be given the `src` up to that index. It might be slow but we aren't guaranteeing any speed with extensions.

It would go like this:

1. `underline` tokenizer would be given `No underline\n: Underline\nNo underline` and it wouldn't find a token.
2. `underline` start would give index `13` as the next potential start
3. `paragraph` tokenizer would be given `src` up to that index so `No underline\n` and it would find a token
4. `underline` tokenizer would be given `: Underline\nNo underline` and it would find a token
5. and so on

UziTech:
I guess the issue might be that index `13` is not an actual start and should have been part of the paragraph

UziTech:
Ya I'm not sure how that will work. I guess we could start by saying that they have to change paragraph if they want to do something that conflicts with it for now and maybe come up with a solution later. Block level tokens should have a blank line before and after anyways so we would still be able to finish this PR with that as a requirement. Really your example of the underline should be an inline token anyway if you don't want a blank line before and after.

calculuschild:
I made a new PR for this #2043. I successfully got the dumb `underline` extension working, but I did manually go into the paragraph rules and add `:` as another interrupter for now.

UziTech:
I really hate how CommonMark makes spec rules describing how CommonMark works not necessarily how markdown should be written.

calculuschild:
> Block level tokens should have a blank line before and after anyways

Yes, except there are so many exceptions to this (tables, hr, heading, fences, list, blockquote, html) that it seems like something we should be able to handle.

UziTech:
Realistically those exceptions should be considered garbage in/garbage out. There shouldn't be spec rules for badly written markdown.

Issue #458, opened 2014-07-29
async renderer support


taylorchu:
It is helpful if one needs to use http to retrieve rich data for rendering.

TiboQc:
+1 for this, I need some enhanced custom markdown that will fetch data from the server to get the info to render the final HTML.

e.g.
`The head of IT, [function:HEAD_OF_IT], will be in charge of...`
Should render:
`The head of IT, John Smith, will be in charge of...`
This needs to get the person from the server. That will always be accurate when displaying the page, even when the person that's assigned as Head of IT changes.

pod4g:
+1 for this
I define a special markdown syntax for our editor

the syntas like this:
```
\`\`\`sku
 1120
\`\`\`
```

janosh:
Async renderers would be very helpful! My use case is turning a markdown image responsive by fetching multiple `srcset`s as well as the image's width and height (to avoid layout shifts) from a CMS based on its `href`. Something like this:

```js
import marked from 'marked'
import yaml from 'js-yaml'

const renderer = {
  async image(href, title, text) {
    if (href?.includes(`mycms.tld`) && !href.endsWith(`.svg`)) {
      const id = href.split(`myCmsId`)[1]
      const query = `{
        asset(id: "${id}") {
          width
          height
        }
      }`
      const { asset } = await fetchFromCms(query)
      const { width, height } = asset
      title = title ? `title="${title}"` : ``
      return `
      <picture>
        <source media="(min-width: 1000px)" type="image/webp" srcset="${href}?w=1000"&fm=webp"" />
        <source media="(min-width: 800px)" type="image/webp" srcset="${href}?w=800"&fm=webp"" />
        <source media="(min-width: 600px)" type="image/webp" srcset="${href}?w=600"&fm=webp"" />
        <source media="(min-width: 400px)" type="image/webp" srcset="${href}?w=400"&fm=webp"" />
        <img src="${href}?w=1000" alt="${text}" ${title} loading="lazy" width="${width}" height="${height}" style="width: 100%; height: auto;" />
      </picture>`
    }

    return false // delegate to default marked image renderer
  },
}

marked.use({ renderer })
```

UziTech:
This is on the v2 project board but does not have a PR yet. If someone would like to start a PR this would go faster.

janosh:
@UziTech I'm happy to try it if pointed in the right direction.

UziTech:
[...]

PR #1878, opened 2020-12-15
move most options to Hooks

UziTech:
## Description

Move most options to Hooks.

This PR might not get merged until v3 but I wanted to throw it out there and see what people think.

The idea behind this is to add a way for extensions to hook into different parts of marked to do things like (sanitizing, escaping, highlighting, etc.) and allowing us to move a lot of the options to extensions.

Example:

```js
marked.use({
  hooks: {
    preprocess(markdown) {
      // Process markdown before marked
    }

    postprocess(html) {
      // Process HTML after marked is finished
    }

    sanitize(html) {
      // Sanitize HTML
      // replace sanitize and sanitizer options
    }

    smartypants(text) {
      // smarty pants
      // replace smartypants option
    }

    error(ex) {
      // Handle marked error
      // replaces silent option
    }

    escape(text) {
      // escape html entities for attributes
    }

    unescape(html) {
      // Unescape header text for id
    }

    languageClass(lang) {
      // set code class
      // replaces langPrefix option
    }

    encode(text) {
      // encode html entities for code blocks
    }

    headerId(text, slugger) {
      // create header id
      // replace headerIds option
    }

    cleanUrl(href) {
      // clean href
      // replace baseUrl option
    }

    mangle(text) {
      // mangle email address
      // replace mangle option
    }

    highlight(code, lang, callback) {
      // highlight code
      // replace highlight option
    }
  }
})
```

### TODO

- [ ] docs
- [ ] create hooks tests
- [ ] create extensions
- [ ] map old options to new hooks and alert user to use extensions


calculuschild:
I'm all for more extendability. This seems fine to me, just skimming through, but maybe you could walk me through what this does differently than the existing `.use` and `.options`?

UziTech:
For one it adds more opportunities suggested for extensibility in https://github.com/markedjs/marked/discussions/1232#discussioncomment-168005 (namely `preprocess` and `postprocess`) it also allows more flexibility for these hooks.

The current highlight and sanitizer are overwritten if multiple extensions provide one. `hooks` are able to fallback to the next option if they return `false`. (sidenote: I was actually thinking about changing `hooks`, `tokenizers`, and `renderers` to take a `next` parameter that they can call instead of returning `false` to make it more like middleware)

Also it provides a better way than using options like `baseUrl` and `headerPrefix`.

calculuschild:
A middleware style would be interesting.... Would that be more performant too?

janosh:
Seems like marked is undergoing a lot of internal changes atm. Would now be good time to tackle async renderers (#458) or would it be better to wait until this is finished?

UziTech:
It would be a great time to tackle async renderers. This PR is mostly a proof of concept. Sort of to flesh out the idea and see what errors we run into.

huaichaow:
want the lifecycle hooks badly.

currently working on an extension that needs cleanup work:

https://github.com/huaichaow/marked-toc-extension/blob/main/src/index.ts

@UziTech can we prioritize this PR?

UziTech:
@huaichaow ya I think the preprocess and postprocess hooks would be the most beneficial I'll start with just those. I'll try to get a PR for that this week.

UziTech:
@huaichaow I'm not completely done with it but #2730 is what I am thinking. Feel free to review it and leave comments.

Issue #1694, opened 2020-05-31
Serving a default export defeats the purpouse of offering both CommonJS and esm


KilianKilmister:
## Reasoning

Default esm exports can be useful for their simplicity. Your module exports only a single function or a simple object? A default export can save the enduser a few characters.

```js
// with default export
import whateverNameIWant from 'some-module'
// vs named export
import { someObject as whateverNameIWant } from 'some-module'
```

But as soon as a module wants to serve multiple things, esm really starts to shine, as the enduser can export exactly what he needs and avoid deconstruction-hell and keep their code considerably more readable and maintainable.

```js
// Need only a few things out of a large library?
import {this, that as someName} from 'large-library'
// Need a large portion of it or simply want to keep it simple?
// use a namespace import:
import * as largeLibrary from 'large-library'
```

Additionally, TypeScript-Language-Features can interpret esm-modules and will automatically map out a module and a modern editor will use that and not only provide suggestions and autocompletion, but will ship them together with in-file doc-comments and automatic typings.
Esm can also be used by a typescript-compiler to generate `.d.ts` files that will mirror the actual project instead of how manual type definitions can be arbitrarily structured, much to the pain of the enduser. (I can provide the necessary configurations if needed)

## Solution

In my opinion it would probably be best to transition the project to natively use esm and use a transpiler to create the CommonJS (the exact opposite of how it is right now). I've done it multiple times before and it is rather painless to do (made even easier by the project using modern ES), adapting to developing using  esm isn't very hard and highly rewarding. It is also backwards-compatible so users don't have to change any of their code.

There is an issue for current esm users, tho as offering both default and named exports is considered bad practice (although permitted by the spec). But there is not much way around it so it would probably be best to have an overlap period with the default entering deprecation.

## User experience using an esm version of marked 
For simple usage one would just have to import the `marked()` function. And as it would be used and act the exact same way as always.

But for usecases where some heavier modification is needed, a user can easily import and extend things like the `Parser` class directly.

### Concluding...

If this proposal is accepted, i'll gladly assist in the transition. I'm already relatively familiar with the source-code as my usecase requires some heavy modification.

UziTech:
We were looking into moving to ESM a while ago but our benchmarks showed cjs was about 10% faster then ESM in Node 12.

Also moving from default to named exports would be a breaking change and I don't think it would be worth it since marked is such a small library the savings from tree shaking would be minuscule.

KilianKilmister:
[...]

PR #1909, opened 2021-01-23
Static lexer

calculuschild:
## Description

PR #1872 is running into some optimization issues, and part of that is due to the way the Benchmark is run. With each individual call to marked(markdown), we create a brand new Lexer, Tokenizer, Parser, Renderer, etc, and doing this for each of the thousands of calls in the Benchmark makes any setup code take up a significant portion of the execution time, i.e., `bind()`ing functions in a constructor, which should be faster than `call()` because it only needs to happen once, is happening with *every Marked execution*.

I added a `static` property `staticLexer` to the Lexer.js, and instead of creating a new Lexer in each call to `static Lexer.lex()`, it instead returns the existing static `staticLexer` property, and just applies any changed options. Did a similar thing with the Tokenizer.js. I don't see any reason to be building up entire separate trees of objects every time someone wants to compile a line of markdown; we only use one at a time, so lets "singleton" it?

Only issue is Lint doesn't like it because `static property` isn't a formal part of JS yet, even though Babel can handle it for us, so we might need to use something like `https://github.com/babel/babel/tree/main/eslint/babel-eslint-parser`

It *seems* to give a bit of a speed boost to the current Master, but it's really hard to tell with the current Benchmarking setup since results vary wildly. I'd appreciate a second opinion.

## Contributor

## Committer

### DIFF
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -378,15 +378,6 @@ var Tokenizer_1 = class Tokenizer {
     this.options = options || defaults$1;
   }
 
-  static staticTokenizer
-
-  static getStaticTokenizer() {
-    if(!this.staticTokenizer) {
-      this.staticTokenizer = new Tokenizer();
-    }
-    return this.staticTokenizer;
-  }
-
   space(src) {
     const cap = this.rules.block.newline.exec(src);
     if (cap) {
@@ -1431,7 +1422,11 @@ var Lexer_1 = class Lexer {
     this.tokens = [];
     this.tokens.links = Object.create(null);
     this.options = options || defaults$2;
-    this.tokenizer = this.options.tokenizer || Tokenizer_1.getStaticTokenizer();
+    // this.options.tokenizer = this.options.tokenizer || new Tokenizer();
+    // this.tokenizer = this.options.tokenizer;
+    if(!this.tokenizer) {
+      this.tokenizer = this.options.tokenizer || new Tokenizer_1();
+    }
     this.tokenizer.options = this.options;
 
     const rules = {
@@ -1452,8 +1447,6 @@ var Lexer_1 = class Lexer {
     }
     this.tokenizer.rules = rules;
   }
-
-  static staticLexer
 
   /**
    * Expose Rules
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -378,6 +378,15 @@ var Tokenizer_1 = class Tokenizer {
     this.options = options || defaults$1;
   }
 
+  static staticTokenizer
+
+  static getStaticTokenizer() {
+    if(!this.staticTokenizer) {
+      this.staticTokenizer = new Tokenizer();
+    }
+    return this.staticTokenizer;
+  }
+
   space(src) {
     const cap = this.rules.block.newline.exec(src);
     if (cap) {
@@ -1422,8 +1431,7 @@ var Lexer_1 = class Lexer {
     this.tokens = [];
     this.tokens.links = Object.create(null);
     this.options = options || defaults$2;
-    this.options.tokenizer = this.options.tokenizer || new Tokenizer_1();
-    this.tokenizer = this.options.tokenizer;
+    this.tokenizer = this.options.tokenizer || Tokenizer_1.getStaticTokenizer();
     this.tokenizer.options = this.options;
 
     const rules = {
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -9,6 +9,8 @@
  * The code in this file is generated from files in ./src/
  */
 
+var commonjsGlobal = typeof globalThis !== 'undefined' ? globalThis : typeof window !== 'undefined' ? window : typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : {};
+
 function createCommonjsModule(fn) {
   var module = { exports: {} };
 	return fn(module, module.exports), module.exports;
@@ -1415,6 +1417,10 @@ function mangle(text) {
  */
 var Lexer_1 = class Lexer {
   constructor(options) {
+    this.setOptions(options);
+  }
+
+  setOptions(options) {
     this.tokens = [];
     this.tokens.links = Object.create(null);
     this.options = options || defaults$2;
@@ -1441,6 +1447,8 @@ var Lexer_1 = class Lexer {
     this.tokenizer.rules = rules;
   }
 
+  static staticLexer = new commonjsGlobal();
+
   /**
    * Expose Rules
    */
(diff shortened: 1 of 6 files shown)


UziTech:
It doesn't seem to speed up the current benchmarks but I see how it could if #1872 was merged and `bind` was used.

src/Lexer.js
```diff
@@ -75,6 +78,8 @@ module.exports = class Lexer {
     this.tokenizer.rules = rules;
   }
 
+  static staticLexer
```
UziTech, line 81: I don't think this static property is necessary since `Lexer.staticLexer` will be `undefined` without this line until it is set anyways.

lib/marked.esm.js
```diff
@@ -1415,11 +1424,14 @@ function mangle(text) {
...
+    this.setOptions(options);
+  }
+
+  setOptions(options) {
     this.tokens = [];
     this.tokens.links = Object.create(null);
     this.options = options || defaults$2;
-    this.options.tokenizer = this.options.tokenizer || new Tokenizer_1();
-    this.tokenizer = this.options.tokenizer;
+    this.tokenizer = this.options.tokenizer || Tokenizer_1.getStaticTokenizer();
```
calculuschild, line 1434: Just cleaning up these two lines. I didn't see a reason for them to be separated, but would appreciate a double-check.

UziTech: I feel like those lines nullify the need for the static tokenizer. With those two lines a new Tokenizer is only created once since it is set in options and reused every time marked is run after that.

calculuschild: Ah that makes sense. So, it didn't necessarily do anything useful when Lexer was remade from scratch each time, but with a static Lexer, `this.options.tokenizer` also stores the Tokenizer between calls to `marked()`. 

Although, it still seems weird to me to use this intermediate step. What do you think about this, so we set the tokenizer in `this.tokenizer` and we don't have to reassign it from `this.options` each time?:

```
if(!this.tokenizer) {
    this.tokenizer = this.options.tokenizer || new Tokenizer();
}
```

UziTech: > So, it didn't necessarily do anything useful

Since `this.options` is the `defaults` object when no options are given then `defaults.tokenizer` is set here so the tokenizer is reused even on `new Lexer()`

UziTech: I don't think the static tokenizer is need because of this. It might still be helpful to save the lexer though.

UziTech:
I wonder if this will actually be any faster since we will still have to run `setOptions` every time anyways to reset everything. I'm thinking we should just pass `this` into the functions instead of binding if it is faster.

calculuschild:
The current version of #1872 is passing this as a parameter but it doesn't seem to make a difference in my machine. I have some other ideas that seemed to speed up parsing, but repeated calls got quite slow with the setup. Maybe I will post in a separate PR so you can judge.

lib/marked.esm.js
calculuschild: Removed Static Tokenizer.

src/Lexer.js
calculuschild: Fixed.

calculuschild:
Made the fixes requested plus a small tweak so `this.tokenizer` doesn't have to be re-written from `this.options.tokenizer` each time. Again, I *think* it's a small speedup but I've been wrong before.

Also removed unneeded Babel extension, and it now passes the Linter.

Even if performance isn't improved here, I think it will streamline any future updates that include heavier setup steps in the constructor.

src/Lexer.js
```diff
@@ -49,11 +49,16 @@ function mangle(text) {
...
+    this.setOptions(options);
+  }
+
+  setOptions(options) {
     this.tokens = [];
     this.tokens.links = Object.create(null);
     this.options = options || defaults;
-    this.options.tokenizer = this.options.tokenizer || new Tokenizer();
-    this.tokenizer = this.options.tokenizer;
+    if (!this.tokenizer) {
```
UziTech, line 59: If they pass a different tokenizer in the options wouldn't we want to use that one here?

calculuschild: That still happens, if my logic is correct.

Instead of this with every loop:
1) Check for `this.options.tokenizer`
2) If it doesn't exist, assign it the default value `new Tokenizer`
3) Assign the result to `this.tokenizer`

Do this:
1) Check for `this.tokenizer`
2) If it doesn't exist, assign it a value, preferably `options.tokenizer`, but if not, default to `new Tokenizer`

The end result is the same: `this.tokenizer == options.tokenizer || new Token`, but we have saved an object write in the second case.

Instead of "saving" the optional Tokenizer to `this.options.tokenizer` to be reapplied every loop, we "save" it where it actually belongs, in `this.tokenizer`, and only have to apply it the first time around.

calculuschild: In fact... I wonder if we even need this separate "setOptions" function run every time. Once the tokenizer is defined and its options are set the first time, normal use would generally not be swapping options on the fly, right? Wouldn't 99% of use cases just set options one time and then run `marked(string)` with those same options every time?

The only real reason I see the need to assign options on the fly is for unit testing in development, because it needs to update options as it steps through the spec tests for GFM, Pedantic, etc. in a random order.

UziTech: What happens in this situation?

```js
marked("# test");

marked.use({
  tokenizer: {
    heading: () {
      // no heading
    }
  }
});

marked("# test");
```

with this PR `this.tokenizer` wouldn't be updated and the new heading tokenizer is ignored.

UziTech:
I see what you are doing here but I don't see any benefits. It doesn't seem any faster especially since we still need to set the options.

If we want it to be faster we should make marked a class that you can use like `const marked = new Marked(options)` then we can reuse everything on every run.

Currently the way marked sets the options they are global so currently with this PR the only way to update the tokenizer would be to restart the node process.

calculuschild:
> with this PR the only way to update the tokenizer would be to restart the node process.

Ok, so we *do* need to support hot-swapping options in production. This is good to know.

> ...I don't see any benefits.

This is the key point I was trying to make in the OP. In its current state, this doesn't do a whole lot, especially if we have to re-apply the same options over and over whether they are changed or not. *However*, if we want to make future changes that add a lot of overhead to the constructor, such as `.bind()`, we can avoid that slowdown by doing something like this. The benefit isn't for the current master, but to enable future fixes and optimizations.

>  It doesn't seem any faster especially since we still need to set the options.

This is a good point, and indicates to me why this PR is a good place to optimize. If we are needlessly resetting the same options over and over with each call to `marked(src)`, then that is an area we can speed up. And now that I know we do need to allow changing options, I'm thinking we should check whether the options have changed, and only update them if needed. Or, we can go the route you mentioned with making Marked an instantiable class, and just fully rebuild a new instance for each new set of options.

UziTech:
Maybe we can add a way to do `new Marked(options)` in v2 and tell people to start using it essentially just returning `marked` and remove `marked.setOptions` and implement #1872 in v3.

We could change marked to:

```js
const {Marked} = require("marked");
const marked = new Marked(options);

marked("# markdown");
```

Or add this way of doing it and keep the old way but just tell people that if they want speed they need to do it this way? It seems like this would have all of theses benefits (plus some [ESM benefits](https://github.com/markedjs/marked/issues/1694)) and not alienate users that don't want to change but don't care about speed.

calculuschild:
Sure, I would not be opposed to that. If it opens up better options for #1872 or similar changes down the line I'm all for it.

FYI I just tried the approach of using `JSON.stringify(options) === JSON.stringify(this.options)` and only applying the options if they have changed. It worked, but the added overhead of strinfigying the objects was more than the benefit from not applying options.

PR #1914, opened 2021-01-27
[WIP] Tokenizer Array using Stringified functions

calculuschild:
## Description

At risk of flooding the repo with garbage PRs:

Experimental modification of #1872. This one uses a static Lexer (via #1909) which enables some hacky shenanigans due to the reduced setup overhead when.

The array of Tokenizer functions is serialized via `function.stringify()` and then combined into a single large function that runs just as fast as the original inlined code. Speed issues seem to be totally solved, testing on my own machine.

BUT, I understand this is a weird way to do things and potentially vulnerable to who knows what. The format for custom functions would have to be carefully designed to prevent conflicts in the stringifying/inlining step, e.g. if they define a variable that already exists in one of the other functions, combining them will lead to "variable already exists" errors.

On the other hand, if we could figure out a robust way of doing this, we could further inline other functions that we know execute in a predefined sequence (i.e. the functions in Tokenizer.js) for further speed gains. The entire `blockTokens()` step could be "recompiled" into a single large monolithic function rather than the hundreds of function calls every second.

I think this has potential, but needs a good sanity check from someone else.

### DIFF
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -1,6 +1,6 @@
 /**
  * marked - a markdown parser
- * Copyright (c) 2011-2020, Christopher Jeffrey. (MIT Licensed)
+ * Copyright (c) 2011-2021, Christopher Jeffrey. (MIT Licensed)
  * https://github.com/markedjs/marked
  */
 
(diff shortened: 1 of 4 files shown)


Monkatraz:
An unfortunate issue with this sort of handler is with browser [CSP `script-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). Using Marked in the browser with a restrictive CSP (which is good for security when you have a lot of UGC, which Marked would be very relevant in) means it will fail to actually create the compiled function, as `eval` and `Function` are disabled unless you pass `unsafe-eval` or load Marked in a trusted manner with `strict-dynamic`.

Monkatraz:
FYI: https://github.com/no-context/moo/issues/141 Moo (regex-compiling lexer) runs into an almost identical issue.

UziTech:
> and potentially vulnerable to who knows what.

Ya I'm going to say I would rather sacrifice speed than security. I'm thinking there are better ways to speed things up like https://github.com/markedjs/marked/pull/1872#issuecomment-768784255

PR #1926, opened 2021-02-04
fix: Join adjacent inlineText tokens

### LINKED ISSUE #1906: Paragraph text is split into separate tokens when using backslash
**Marked version:** 1.2.7

**Describe the bug**
The `inline.text` rules seem to be splitting inline text when it encounters a backslash, resulting in two separate text tokens inside a paragraph token.

**To Reproduce**
Observe on the [demo.](https://marked.js.org/demo/?outputType=lexer&text=First%5Csecond%0A%0AFirst%20second&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=1.2.7)

`First second` results in: 

```
[
{type:"paragraph", raw:"First second", text:"First second", tokens:[
  {type:"text", raw:"First second", text:"First second"}
]}
]
```

But `First\second` results in

```
[
{type:"paragraph", raw:"First\\second", text:"First\\second", tokens:[
  {type:"text", raw:"First", text:"First"}
  {type:"text", raw:"\\second", text:"\\second"}
]}
]
```

**Expected behavior**
While the resulting HTML is the same, this causes issues when using, for instance, a custom `text` renderer. The expected string is instead sliced into multiple parts, which prevents handling of the complete string as a whole.

Either the two consecutive text tokens should be joined as they are in the "block" text, or the `inline.text` rules need to be fixed to not stop at the first backslash.

calculuschild:
**Marked version:** 1.2.9

## Description

- Fixes #1906. Joins together adjacent inline `text` tokens similar to how block text tokens are merged. HTML output doesn't change, but this makes it easier to make a custom renderer since text tokens aren't broken up.

## Contributor

## Committer

### DIFF
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -1816,11 +1816,10 @@ var Lexer_1 = class Lexer {
         prevChar = token.raw.slice(-1);
         keepPrevChar = true;
         const lastToken = tokens[tokens.length - 1];
-        if(lastToken && lastToken.type === 'text') {
+        if (lastToken && lastToken.type === 'text') {
           lastToken.raw += token.raw;
           lastToken.text += token.text;
-        }
-        else {
+        } else {
           tokens.push(token);
         }
         continue;
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -1815,7 +1815,14 @@ var Lexer_1 = class Lexer {
         src = src.substring(token.raw.length);
         prevChar = token.raw.slice(-1);
         keepPrevChar = true;
-        tokens.push(token);
+        const lastToken = tokens[tokens.length - 1];
+        if(lastToken && lastToken.type === 'text') {
+          lastToken.raw += token.raw;
+          lastToken.text += token.text;
+        }
+        else {
+          tokens.push(token);
+        }
         continue;
       }
 
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -391,18 +391,9 @@ var Tokenizer_1 = class Tokenizer {
     }
   }
 
-  code(src, tokens) {
+  code(src) {
     const cap = this.rules.block.code.exec(src);
     if (cap) {
-      const lastToken = tokens[tokens.length - 1];
-      // An indented code block cannot interrupt a paragraph.
-      if (lastToken && lastToken.type === 'paragraph') {
-        return {
-          raw: cap[0],
-          text: cap[0].trimRight()
-        };
-      }
-
       const text = cap[0].replace(/^ {1,4}/gm, '');
       return {
         type: 'code',
@@ -719,17 +710,9 @@ var Tokenizer_1 = class Tokenizer {
     }
   }
 
-  text(src, tokens) {
+  text(src) {
     const cap = this.rules.block.text.exec(src);
     if (cap) {
-      const lastToken = tokens[tokens.length - 1];
-      if (lastToken && lastToken.type === 'text') {
-        return {
-          raw: cap[0],
-          text: cap[0]
-        };
-      }
-
       return {
         type: 'text',
         raw: cap[0],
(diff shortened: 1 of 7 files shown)


calculuschild:
CI is not passing but all the spec tests and benchmarks pass just fine on my machine. Not sure what's going on. Any help?

calculuschild:
Ah I see there's a separate Lexer unit test suite. It expects adjacent but separate text tokens in a couple of cases. I assume that isn't a requirement anymore....

UziTech:
I think I fixed the tests in https://github.com/calculuschild/marked/pull/2. I also changed the block text token to do the merge in the lexer instead of the tokenizer and merge text tokens returned by other tokenizers.

calculuschild:
> I also changed the block text token to do the merge in the lexer instead of the tokenizer and merge text tokens returned by other tokenizers.

@UziTech I was going to ask about this as well so I'm glad you had the same idea. Seemed to make more sense in the Lexer. Do we also want to do the same for the block `code` tokenizer? It also does this same merge thing.

UziTech:
Ya it is probably a good time to change the tokenizer signatures if we need to since this and #1864 should be released as v2 soon.

lib/marked.esm.js
```diff
@@ -1751,10 +1741,16 @@ var Lexer_1 = class Lexer {
       // reflink, nolink
       if (token = this.tokenizer.reflink(src, this.tokens.links)) {
         src = src.substring(token.raw.length);
+        const lastToken = tokens[tokens.length - 1];
```
UziTech, line 1744:
```suggestion
        lastToken = tokens[tokens.length - 1];
```

lib/marked.esm.js
```diff
@@ -1734,7 +1718,13 @@ var Lexer_1 = class Lexer {
         src = src.substring(token.raw.length);
         inLink = token.inLink;
         inRawBlock = token.inRawBlock;
-        tokens.push(token);
+        const lastToken = tokens[tokens.length - 1];
```
UziTech, line 1721:
```suggestion
        lastToken = tokens[tokens.length - 1];
```

UziTech (changes requested):
These changes are from my code

UziTech:
Can you update the tokenizer signatures for `text` and `code` in the [docs](https://marked.js.org/using_pro#tokenizer)

calculuschild:
Yay! All the Block Tokenizers have the same signature now! 🎉

UziTech (approved):

src/Lexer.js
```diff
@@ -136,14 +136,15 @@ module.exports = class Lexer {
...
-      if (token = this.tokenizer.code(src, tokens)) {
+      if (token = this.tokenizer.code(src)) {
         src = src.substring(token.raw.length);
-        if (token.type) {
-          tokens.push(token);
-        } else {
-          lastToken = tokens[tokens.length - 1];
+        lastToken = tokens[tokens.length - 1];
+        // An indented code block cannot interrupt a paragraph.
+        if (lastToken && lastToken.type === 'paragraph') {
```
UziTech, line 143: I wonder if we should check for a paragraph before we call the `code` tokenizer? That might save some work that doesn't need to be done.

calculuschild: Probably worth trying. It's a tradeoff of always checking `LastToken` and sometimes calling `codeTokenizer` vs always calling `codeTokenizer` and sometimes checking `LastToken`. I'm not sure how often this situation comes up that one would be better than the other.

UziTech: > If LastToken is a paragraph though what do we do? Just continue? or call the "text" tokenizer out of sequence?

Good point. I suppose we would still need to call code tokenizer to see if we should skip the other tokens. Maybe it is still better to check code tokenizer first.

styfle (approved):

PR #1933, opened 2021-02-09
Update codeql-analysis.yml

XhmikosR:
Remove the HEAD workaround as it's no longer suggested upstream. Also, reindent and and remove comments boilerplate.

Non-whitespace diff: https://github.com/markedjs/marked/pull/1933/files?w=1

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -26,8 +26,5 @@ jobs:
         with:
           languages: "javascript"
 
-      - name: Autobuild
-        uses: github/codeql-action/autobuild@v1
-
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v1
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,71 +1,34 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
 name: "CodeQL"
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
+      - "!dependabot/**"
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [master]
+    branches:
+      - master
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -1,71 +1,34 @@
...
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
 name: "CodeQL"
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
+      - "!dependabot/**"
```
UziTech, line 7: Seems redundant to say we only want this on `master` and not `dependabot`.
```suggestion
```

XhmikosR: It's not redundant because we know dependabot always opens a PR.

UziTech: but if we don't want to run code ql on `dependabot` branches why don't we just not include this line? then it will only run code ql on master.

XhmikosR: Actually, you are right. Not my day today, I need to drink more coffee :)

XhmikosR: OK now that I think about it, it is needed for PRs. But it's your call. Right now this PR matches the current behavior on master which is to run CodeQL even on dependabot PRs.

UziTech: I don't see any problem with running it on dependabot PRs.

.github/workflows/codeql-analysis.yml
```diff
@@ -1,71 +1,33 @@
...
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: "javascript"
 
-    #- run: |
-    #   make bootstrap
-    #   make release
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v1
```
UziTech, line 31: I don't think we need this auto build

```suggestion
```

XhmikosR:
I'm going to close the PR. Feel free to make any changes yourself to match the behavior you want.

UziTech (approved):
:100:

PR #1938, opened 2021-02-10
Update docs to clarify Node.js and browser support

alasdairhurst:
**Marked version:** 2.0.0

**Markdown flavor:** n/a

## Description

Updates docs to describe Node.js support clearly, and clarify the strategy for dropping old versions of Node.js in relation to semver.

## Expectation

More clear docs for consumers

## Result

n/a

## What was attempted

n/a

## Contributor

## Committer

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ Also read about:
 * [Options](https://marked.js.org/#/USING_ADVANCED.md)
 * [Extensibility](https://marked.js.org/#/USING_PRO.md)
 
+## Compatibility
+
+**Node.js:** Only [current and LTS](https://nodejs.org/en/about/releases/) Node.js versions are supported. End of life Node.js versions may become incompatible with Marked at any point in time.
+
+**Browser:** Not IE11 :)
+
 ## Installation
 
 **CLI:** `npm install -g marked`
--- a/docs/PUBLISHING.md
+++ b/docs/PUBLISHING.md
@@ -10,6 +10,6 @@ Marked uses [semantic-release](https://github.com/semantic-release/semantic-rele
 
 We follow [semantic versioning](https://semver.org) where the following sequence is true `[major].[minor].[patch]`:
 
-1. **Major:** There is at least one change to the public API or a break from the [CommonMark](https://spec.commonmark.org/current/) or [GFM](https://github.github.com/gfm/) spec.
+1. **Major:** There is at least one change to the public API or a break from the [CommonMark](https://spec.commonmark.org/current/) or [GFM](https://github.github.com/gfm/) spec. Only [current and LTS](https://nodejs.org/en/about/releases/) Node.js versions are supported at any point in time.  A drop in support for a Node.js version may not result in a semver major bump to Marked.
 2. **Minor:** There is at least one new feature added to the public API.
 3. **Patch:** Changes that move Marked closer to spec compliance or change a public API that does not break backwards compatibility.


README.md
```diff
@@ -29,6 +29,12 @@ Also read about:
 * [Options](https://marked.js.org/#/USING_ADVANCED.md)
 * [Extensibility](https://marked.js.org/#/USING_PRO.md)
 
+## Compatibility
+
+**Node.js:** Only [current and LTS](https://nodejs.org/en/about/releases/) Node.js versions are supported. End of life Node.js versions may become incompatible with Marked at any point in time.
+
+**Browser:** Not IE11 :)
```
alasdairhurst, line 36: @UziTech Makes sense to also document the supported browsers but I can't seem to find them after a bit of light digging. Rollup doesn't seem to be using any specific browserlist for transpiling.

alasdairhurst:
I noticed the version in component.json was out of sync. Not sure what this is used for or how it should be updated, but i'm guessing this should be matching the package.json version?
https://github.com/markedjs/marked/blob/master/component.json

UziTech:
> I noticed the version in component.json was out of sync

I'm not even sure what that is for. Looks like it was for something a long time ago. It should probably just be deleted.

README.md
UziTech: it compiles to es5 but some functions (like `startsWith`) might need polyfills for browsers that don't include them.

alasdairhurst:
>I'm not even sure what that is for. Looks like it was for something a long time ago. It should probably just be deleted.

Had a little dig a bit further and it's this old tool
https://github.com/componentjs/component/blob/master/Readme.md
Deprecated and dead for quite a few years now

package.json
```diff
@@ -83,6 +83,6 @@
     "preversion": "npm run build && (git diff --quiet || git commit -am build)"
   },
   "engines": {
-    "node": ">= 8.16.2"
+    "node": "^10 || ^12 || >=14"
```
UziTech, line 86: I wonder if it would be better to remove the `engines` field as no one seems committed to keeping it up to date.

alasdairhurst: I'd prefer to keep it so that downstream users would be at least notified by NPM if they are using an unsupported version. Maybe we can automate it in a script as part of the build process?. I'll have a think about the lightest approach. Potentially parsing https://github.com/nodejs/Release/blob/master/schedule.json, and have a play tomorrow.

Martii: > notified by NPM if they are using an unsupported version

In the [package.json spec currently](https://docs.npmjs.com/cli/v6/configuring-npm/package-json/#engines) it is advisory at this time. Pay particular notice to `engine-strict` and not `engineStrict`. Knowing if a package could be advised on install that it is incompatible would be nice, if *npm* gets those and related issues in those issues fixed. We use the package.json advisory since we're "top level" but I also mention to other devs to read CHANGELOGs assuming a project has one.

See also:
* https://github.com/npm/cli/issues/692 *(hover for issue label status)*
* https://github.com/npm/cli/issues/786 *(hover for issue label status)*

I do read deprecation warnings too in case *node* changes their mind.

styfle: We should definitely keep `engines`. Its not just a signal to consumers but also anyone who wants to contribute to the project.

Even though npm doesn't respect `engines`, yarn does.

README.md
styfle: Let's just link to `package.json` so that can be the canonical place to determine which Node.js versions are supported.

We should also add this to the docs website, probably after the "Supported Markdown specifications" section.

package.json
styfle, line 86:
```suggestion
    "node": ">=10"
```

Let's simplify this semver range.

We shouldn't explicitly block 11 and 13 unless there are known bugs.

UziTech: > Its not just a signal to consumers but also anyone who wants to contribute to the project.

The problems with engines as I see them:

- No one looks at it to make sure their PR complies with it.
- No one reviewing PRs makes sure they are complying with that field.
- No one updates it when versions of node pass their end of life.

Therefore it is usually out of date and not very useful to anyone.

If there is a way to automate updating it and testing for active node versions, or if someone is willing to commit to keeping it up to date and review PRs to make sure they comply, I am all for having. If there is no way to automate it or no one step forward to maintain it, it just seems like a burden for us developers and sometimes useful to users.

styfle: Vercel deployments use it, so its is implicitly tested in the sense that Node.js 14 is selected based on `engines` when building our docs and api.

Maybe we just need a test to compare `engines` to GH CI to make sure we're testing what is defined there.

UziTech: > Vercel deployments use it

But they would use the same version if we didn't have it.

> Maybe we just need a test to compare `engines` to GH CI to make sure we're testing what is defined there.

How do we do that? Currently we only test `lts` and `latest` node versions so something could break node v10 and we wouldn't know it.

alasdairhurst: Ah, when you said you only support lts and latest, I thought you meant *all* current LTS versions of node (10 and 12) as well as latest (14) rather than just the most recent LTS

UziTech: I feel like the `engines` field, at best, is a crutch. If someone wants to know which versions we support we should put that in the docs but relying on the `engines` field to be correct seems ridiculous if it is not necessarily going to be up to date.

"When looking for the time it is better to have not found a clock at all then to have found one that is incorrect." -- Someone Smart

styfle: Let's correct the clock then 👍 

> But they would use the same version if we didn't have it.

No, they would use the version assigned in the project settings. See [the docs](https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version).

alasdairhurst: I think @styfle had something close to the right idea.
We could do the following if you want to support all LTS + latest:

1. Create a script to fetch the current latest node release versions & filter for LTS and current
2. Run these latest versions in a semver.satisfies call against the package.json `engines` field.
3. Run this script as part of CI. If a new version of node comes out then we'll be instantly notified and can update `engines` in that PR who's branch failed.

UziTech: > Ah, when you said you only support lts and latest, I thought you meant all current LTS versions of node

Realistically we support whatever version someone is willing to maintain. If someone wants to submit a PR that fixes v2 for node v8 I would happily accept it, as long as it doesn't make things slower or break things for newer node versions.

UziTech: > Whenever a new Project is created, the latest Node.js LTS version available on Vercel at that time is selected for it.

Great so we don't need the engines field according to the docs.

> Let's correct the clock then 👍

But who is going to make sure the clock stays correct?  and what is the correct version? as anyone tested it on node v10.0.0? if not than `>=10` could be incorrect.

alasdairhurst: Maybe we can make use of Renovate https://docs.renovatebot.com/node/

UziTech: > Maybe we can make use of Renovate

I would be ok with Renovate. I use it on most of my other projects to auto-merge dependency updates.

> Does it hurt to add it to CI?

Ya if we can get it automated than I would be ok with keeping it.

> unless you mean the original 10

Tecnically `v10.0.0` is in the range `>=10` so someone looking in engines or using yarn would be correct in assuming `v10.0.0` was a supported node version.

styfle: If we aren't testing against Node 10 (looks like we're not) then we probably shouldn't say we support it.

Also since 10 is reaching EOL soon lets just say `>=12`.

My main point is that our CI needs needs to match what we expect. If we drop support for a version of Node.js, that would be a major semver.

UziTech: > Also since 10 is reaching EOL soon lets just say `>=12`.

So we should support v12.0.0 and v12.0.1 ...? otherwise if the `engines` field is set to `>=12` than it is about as useful as a broken clock.

> If we drop support for a version of Node.js, that would be a major semver.

I agree, but as I said above we really support whatever someone is willing to maintain.

UziTech: > My main point is that our CI needs needs to match what we expect.

I think it does. I expect marked to work with `lts` and `latest` versions of node. If we could do something like

```json
  engines: {
    "node": "lts || latest"
  }
```

I would be all for it.

styfle: > So we should support v12.0.0 and v12.0.1

Yes, of course. Since Node.js follows semver we can assume code that works on 12.0.0 will work on 12.0.1.

> "node": "lts || latest"

I think this is backwards.

You're trying to make `engines` match our current CI but I think we need to make our CI match `engines` by being explicit (test both 12.x and 14.x which are both LTS and probably 15.x which is stable).

https://nodejs.org/en/about/releases/

UziTech: > Yes, of course. Since Node.js follows semver we can assume code that works on 12.0.0 will work on 12.0.1.

Not true, For example: `fs.rm(dir)` would work in node v14.14.0 but not in v14.13.0. The `fs.rm` function was added in v14.14.0

So just because something works for latest version in a major version does not mean it will work for every version in that major version. Semver only guarantees the opposite. If something works in v14.0.0 it should work in every later version with the same major version number. 

And every fix could be a breaking change for someone's [workflow](https://xkcd.com/1172/)

UziTech: > You're trying to make `engines` match our current CI

Yes, I am trying to make `engines` match the versions known to work with marked (because they are the only ones we test). The problem with testing specific versions or supporting specific versions is that they get out of date really fast and everyone forgets about them in smaller projects like marked.

UziTech: I understand how useful `engines` can be **if it is accurate**. I just don't believe anyone is willing to keep it accurate enough to be useful in marked.

README.md
realityking: It might be worth to a set a rule which browsers are supported and to enforce it using https://www.npmjs.com/package/eslint-plugin-compat

package.json
realityking: I'd suggest:
- setting the range in `engines` to start with the first LTS version of the major. So something like `"^10.13 || ^12.13  || ^14.15  || >=15"`. Features are not back ported anymore after a release reaches LTS so you have a consistent basis.
- If you want to be extra safe, add 10.13.0, 12.13.0 and 14.15.0 to the build matrix. That way you'll notice when a new features breaks an older version and can bump the `engines` field whenever the build matrix is changed - this is also easy to review.

As an aside, strong vote from my end to bump the major whenever adjusting the list of supported Node or Browser version - it makes things a lot easier for downstream consumers.

UziTech: @realityking The issue I have with setting the `engines` field is it is very difficult to keep it accurate for a project as small as marked where there are often weeks or even months between changes.

This is a community project, no person or company has any commercial interest in marked. So keeping the `engines` field up to date seems like a waste of time that provides no benefit because it will rarely be accurate and because as I stated in https://github.com/markedjs/marked/pull/1938#discussion_r575383861 we support any version of node that someone is willing to maintain.

calculuschild:
Is this something we can address as part of #2157 before the V3 update? This PR has been open for quite a long time, and seems like this would be the perfect time to either make this change or officially close it out since it relates to semver issues.

UziTech:
We did update `engines` for v3 in #2157 so the rest of this PR is just docs changes.

calculuschild:
Right. Just saying since we did change `engines` we might as well finish updating the related docs.

UziTech:
Yes @alasdairhurst if you would rebase this we could update the docs.

alasdairhurst:
@UziTech @calculuschild  Rebased.

calculuschild (approved):

docs/PUBLISHING.md
```diff
@@ -10,6 +10,6 @@ Marked uses [semantic-release](https://github.com/semantic-release/semantic-rele
 
 We follow [semantic versioning](https://semver.org) where the following sequence is true `[major].[minor].[patch]`:
 
-1. **Major:** There is at least one change to the public API or a break from the [CommonMark](https://spec.commonmark.org/current/) or [GFM](https://github.github.com/gfm/) spec.
+1. **Major:** There is at least one change to the public API or a break from the [CommonMark](https://spec.commonmark.org/current/) or [GFM](https://github.github.com/gfm/) spec. Only [current and LTS](https://nodejs.org/en/about/releases/) Node.js versions are supported at any point in time.  A drop in support for a Node.js version may not result in a semver major bump to Marked.
```
styfle, line 13: I don't think we plan to do this.

We should do semver major when dropping a version of Node.js and bump the engines field

https://github.com/markedjs/marked/blob/e781b18b0faf8d23691c88769638027a33fcaaec/package.json#L87

alasdairhurst: Would be great if that happened but I wasn't able to convince everyone on the team :)

Unless something changed since then?

UziTech: @styfle Currently we test Node v12, Node lts (v14), and Node latest (v16). When Node LTS changes to v16 we will be testing Node v12, Node lts (v16), and Node latest (v17), so Node v14 will no longer be tested. If we had a team of people working on this every day I think we could task someone with updating that, but realistically there are three-ish people working on marked, whenever we have extra time.

I think it is good to set the expectations low. Ideally we won't break Node versions that aren't tested and we we will only update the engines field on semver major releases but I still like saying, realistically, we could break older node versions in patches since only lts and latest are guaranteed to be tested and the best fix might be for you to update Node.

Issue #2000, opened 2021-04-08
Set `type` member for all nodes


sapristi:
**What pain point are you perceiving?.**

Not all nodes have a `type` member set, namely the `def` (see https://github.com/markedjs/marked/blob/d0f9ffdb2479ec04a1806b5a3bafee5761a5ff72/lib/marked.js#L747

and also in one case for the `space` node: https://github.com/markedjs/marked/blob/d0f9ffdb2479ec04a1806b5a3bafee5761a5ff72/lib/marked.js#L485

Having a defined `type` member would help discriminating them when iterating the markdown tree.
It also seems to be an issue for the https://www.npmjs.com/package/@types/marked packages, where the `Def` interface does't have a `type` member, and thus it is not possible to discriminate node types in typescript (I guess it could be possible to add a type member in the interface, but it would be troublesome if this project later chose a different string).

**Describe the solution you'd like**

Add a `type` member to all nodes of the markdown tree.

UziTech:
If you would like to create a PR to add the type that would be much appreciated.

You should just need to add a `type` property to the `def` tokenizer's return object

https://github.com/markedjs/marked/blob/d0f9ffdb2479ec04a1806b5a3bafee5761a5ff72/src/Tokenizer.js#L344-L356

sapristi:
Ok I will do one. What about the case where the `space` node doesn't define a `type` member ?

PR #2001, opened 2021-04-09
Add a `type` property to the `def` tokenizer's return object

sapristi:
Related to https://github.com/markedjs/marked/issues/2000

**Marked version:**

2.0.1

**Markdown flavor:** all

## Description

- Fixes https://github.com/markedjs/marked/issues/2000

## Expectation

Markdown nodes all have a `type` property to ease working with the markdown tree.

## Result

`def` nodes do not have a `type` property. 

## What was attempted

Using `walkTokens` in typescript, I have the need to discriminate between nodes; the https://www.npmjs.com/package/@types/marked package doesn't define a `type` member for the `Def` interface, seemingly because there is not one here either.

## Contributor

## Committer

### DIFF
--- a/lib/marked.js
+++ b/lib/marked.js
@@ -751,6 +751,7 @@
         if (cap[3]) cap[3] = cap[3].substring(1, cap[3].length - 1);
         var tag = cap[1].toLowerCase().replace(/\s+/g, ' ');
         return {
+          type: 'def',
           tag: tag,
           raw: cap[0],
           href: cap[2],


UziTech:
@styfle @calculuschild would this be a breaking change?

UziTech (approved):
LGTM

calculuschild:
I don't see anything that would be breaking. I think *adding* information is usually ok; only if someone is depending on no type to detect def tokens but I doubt that is a common occurrence. LGTM.

calculuschild (approved):

styfle (approved):
Looks good 👍 

Agreed, I don't think adding a new prop is considered breaking

sapristi:
Thanks !

sapristi:
Thanks !

Issue #2033, opened 2021-04-29
Consider changing Parser's switch cases to methods, for extensibility


mitranim:
**What pain point are you perceiving?**

My use case seems to require special logic which cannot be done at the `Renderer` level, cannot be done by editing the AST via `walkTokens` (unless #2032 is implemented), but could be done at the `Parser` level if the API was easier to extend.

**Describe the solution you'd like**

In `Parser`, convert the bodies of switch cases to methods:

```js
class Parser {
  parseToken(token, renderer) {
    switch (token.type) {
      case 'link': return this.link(token, renderer)
      case 'text': return this.text(token, renderer)

      // ...more cases for other known types; then:

      default: return this.default(token, renderer)
    }
  }

  link(token, renderer) {
    return renderer.link(token.href, token.title, this.parseInline(token.tokens, renderer))
  }

  text(token, renderer) {
    return renderer.text(token.text)
  }

  // ... more methods for other known types; then:

  default() {throw appropriate error}
}
```

We could even drop the switches and look up methods dynamically:

```js
class Parser {
  parseToken(token, renderer) {
    if (this[token.type]) return this[token.type](token, renderer)
    return this.default(token, renderer)
  }
}
```

This would also make it trivial to implement #2032 in user code, by overriding `default` to support new token types.

**Actual use case**

To avoid an [XY problem](https://xyproblem.info) (proposing a poorly-chosen solution instead of stating the actual goal), here's what I actually need to do.

Depending on the source data:

* Sometimes links are treated as their child `.tokens`, ignoring `href`.
* Sometimes links are preserved as-is.
* Sometimes a text token is converted to multiple tokens including 0 to N links, _unless_ it's already inside a link.

With this proposal, this logic could be implemented by subclassing `Parser` and overriding just the right methods:

```js
class MdParser extends md.Parser {
  link() {/* just the stuff */}
  text() {/* just the stuff */}
}
```

mitranim:
Addendum. For dynamic method lookup, we can avoid collisions with `Object.prototype` properties by extending `null` instead:

```js
function Null() {}
Null.prototype = null

class Parser extends Null {}

console.log(Object.getOwnPropertyNames(Parser.prototype))
// ['constructor']
```

This gives us a squeaky-clean `Parser`.

UziTech:
We have a proposal to add tokenizers and renderers to marked [here](https://github.com/markedjs/marked/pull/1872#issuecomment-827933753). Maybe that would solve this problem.

[...]

PR #2038, opened 2021-05-03
[WIP] extensible Parser: separate methods for token types

mitranim:
This branch is **not yet ready for merge**; see known regressions below.

Logic for different token types has been moved from switch cases to methods. The main parse loop now simply looks up the method by token type, falling back on `default()`.

Token types are now an open set, rather than a closed set. New token types can be supported by subclassing Parser and adding corresponding methods, or by overriding `default()`.

Consolidated the parsing logic for "top" and "inline" tokens. Removed `parser.parseInline`. The top-level function `parseInline` uses `lexer.lexInline` to generate an appropriate AST; additional restrictions at the parser level appear redundant.

More flexible approach to "contextual" parsing logic, such as using a different renderer or indicating "loose" mode. Now every parse method takes an additional parameter: a "context", which contains a renderer and possibly other settings. By default, such context objects are allocated lazily and no more than once per Parser. The method signatures make it possible to use context inheritance, where child contexts inherit from parent contexts, overriding only some of their properties. This is not used by default, but can be useful for advanced cases such as one described in #2033.

Benchmarks indicate no significant performance regressions or improvements. Differences appear to be within noise levels.

Known regressions that must be fixed before merging the MR:

  * No support for coalescing adjacent `text` nodes.
  * List "loose" mode, with paragraph wrapping, is now a placeholder with incorrect logic.
  * Fewer tests pass, presumably as a result of ↑.

Addresses [#2033](https://github.com/markedjs/marked/issues/2033).

---

Requesting feedback on:

  * Good ways to recover the "text coalesce" functionality, where adjacent text nodes would be joined by `\n`.
  * Good ways to recover the feature where in loose-mode lists text would sometimes be wrapped into paragraphs.
  * Possibly anything else I've overlooked.

The reason for the above regression is that the main parse loop has been split, is no longer "aware" of token types, and has no special cases. I really want to keep it this way.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -2,21 +2,26 @@ const Renderer = require('./Renderer.js');
 const TextRenderer = require('./TextRenderer.js');
 const Slugger = require('./Slugger.js');
 const { defaults } = require('./defaults.js');
-const {
-  unescape
-} = require('./helpers.js');
+const { unescape, isFun } = require('./helpers.js');
+
+function Null() {}
+Null.prototype = null;
 
 /**
  * Parsing & Compiling
  */
-module.exports = class Parser {
+module.exports = class Parser extends Null {
   constructor(options) {
+    super();
     this.options = options || defaults;
     this.options.renderer = this.options.renderer || new Renderer();
     this.renderer = this.options.renderer;
     this.renderer.options = this.options;
     this.textRenderer = new TextRenderer();
     this.slugger = new Slugger();
+    this.context = undefined;
+    this.textContext = undefined;
+    this.looseContext = undefined;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/Parser.js
```diff
@@ -28,236 +33,199 @@ module.exports = class Parser {
...
+  html(token, ctx) {
+    return ctx.renderer.html(token.text);
+  }
+
+  paragraph(token, ctx) {
+    return ctx.renderer.paragraph(this.parse(token.tokens, ctx));
+  }
+
+  // Regression; missing features:
+  //   * Join adjacent text tokens with '\n'.
```
UziTech, line 184: Could we add `tokens` and `currentIndex` to the context object so the `text` function could combine tokens the same way it does in the switch statement?

mitranim: Unless I misunderstood, sounds like it involves mutating the context. Parsing/compiling any token, even `text`, is potentially recursive, with nested parse loops. To avoid trampling each other's `tokens` and `currentIndex` in the context, parse loops would have to either make new context objects (slow), or stash some contextual values in the stack frame of the current parse loop, and restore them before returning from the loop (shouldn't affect performance, but looks dirty). This also creates a "communication channel" between a parse loop and the methods it's calling, though a shared mutable object. This could work, but a cleaner approach would be much appreciated.

I admit this is a self-inflicted problem. The "old" loop had the power to handle this case and worked just fine. But I also feel that making the parse loop really dumb, where it just calls one method per token, "buys" us nice extensibility through method overrides, and is at least worth exploring.

I don't entirely understand the reasoning behind this special case. Why would there be adjacent text tokens? Is this currently generated anywhere inside Marked, or supported for outside customization? Could this be handled better at the Lexer level? Could it be handled better through `text.tokens`?

UziTech: > "buys" us nice extensibility through method overrides

If the goal is just to have extensibility it seems like it would be easier to just have it check for function in the `default` case.

> but a cleaner approach would be much appreciated

Clean code isn't really what marked is aiming for. Clean code often comes with slow performance. We try to make it as clean as possible without sacrificing performance at all. The switch statement may be ugly but it works and is as fast as it can be. Perhaps the goal of a "dumb" parse loop is the wrong goal.

UziTech:
I think #2043 fixes the issue this PR would solve

Issue #1693, opened 2020-05-31
Can't add additional custom tokenizer or renderers?


alystair:
**What pain point are you perceiving?.**
I'm reviewing Marked documentation, attempting to create a custom setup where, it transforms new lines starting with 'notice: ' into a specifically formatted DIV.  By my understanding I need to first add a custom named tokenizer and then a renderer based on it? Or am I going about this the wrong way?
 
**Describe the solution you'd like**
I'd like to easily add this functionality without creating a separate post-processing tool external to Marked...

KilianKilmister:
You are propably able to get by creating a [walk-token]. They basically act as an intermediate processing step before the data is handed to the renderer. Their ability is somewhat limited, but it should be plenty for things like this.

[walk-token]:<https://marked.js.org/#/USING_PRO.md#walk-tokens>

alystair:
I tried the following but it doesn't work. Does tokenizer only substitute existing tokens? Can you not add your own custom ones?

```js
const tokenizer = {
	tipNotice(src) {
		const match = src.match(/\nnotice: (.*)\n/);
		if (match) {
			return {
				type:'tipNotice',
				raw:match[0],
				text:match[1].trim()
			};
		}
		return false;
	}
};
const renderer = {
	tipNotice(text) {
		return `<div class="tip notice">${text}</div>`;
	}
};
marked.use({ tokenizer, renderer });
marked(content,{ headerIds:true });
```

KilianKilmister:
@alystair Both the Parser and the Lexer implementation currently have their structure hardcoded. So you can so there currently is no straight way of altering what they hand to the Tokenizer and Renderer, as those are "stupid" as they are build tor speedy processing.

And as it turnes out, it is pretty much impossible to make invasive modifications short of altering the source-code or meta programming a weil.

Whoever wrote the core of this package can really pad himself on the back, it is as solid as granit. 

#1695 i filed an issue about that yesterday

alystair:
Argh... in that case I'll have to replace several built-ins like paragraph, and review the source code to expand on the current functionality. Fun.

KilianKilmister:
A little hint: copy the repo and directly modify the sourcecode i spent a full day trying to alter the parser and lexer from inside my module and i have nothing to show for it, as marked is stable as hell.

But the code is modern and well formatted. Navigating it is a breze

UziTech:
This is something that could be useful. If someone wants to create a PR I would be happy to review it.

[...]

Issue #339, opened 2014-01-29
Support image resizing


SimonCropp:
is there any support for changing the size of an image. or do we have to drop back to html?

timlevett:
+1

eugeniop:
+1

ponyesteves:
+1

isiahmeadows:
In my particular case, where I already had to modify the renderer to properly highlight a code block ([known issue](https://github.com/chjj/marked/pull/418)), I just added another method to do just this.

```js
// Super simple sanitizer
function sanitize(str) {
    return str.replace(/&<"/g, function (m) {
        if (m === "&") return "&amp;"
        if (m === "<") return "&lt;"
        return "&quot;"
    })
}

/**
 * Make Marked support specifying image size in pixels in this format:
 *
 * ![alt](href = x WIDTH "title")
 * ![alt](href = HEIGHT x "title")
 * ![alt](href = HEIGHT x WIDTH "title")
 *
 * Note: whitespace from the equals sign to the title/end of image is all
 * optional. Each of the above examples are equivalent to these below,
 * respectively:
 *
 * ![alt](href =xWIDTH "title")
 * ![alt](href =HEIGHTx "title")
 * ![alt](href =HEIGHTxWIDTH "title")
 *
 * Example usage:
 *
 * ![my image](https://example.com/my-image.png =400x600 "My image")
 */
renderer.image = function (href, title, alt) {
    var exec = /\s=\s*(\d*)\s*x\s*(\d*)\s*$/.exec(href)
    var res = '<img href="' + sanitize(href) + '" alt="' + sanitize(alt)
    if (exec && exec[1]) res += '" height="' + exec[1]
    if (exec && exec[2]) res += '" width="' + exec[2]
    return res + '">'
}
```

It takes advantage of the fact Marked parses images this way:

```md
![img](http://foo.com/img.png =100x300)
![img](http://foo.com/img.png =100x300 "foo")
```

```html
<p><img src="http://foo.com/img.png =100x300" alt="img"></p>
<p><img src="http://foo.com/img.png =100x300" alt="img" title="foo"></p>
```

Do note that you can't flip the two:

```md
![img](http://foo.com/img.png "foo" =100x300)
```

```html
<p><img src="http://foo.com/img.png &quot;foo&quot; =100x300" alt="img"></p>
```

---

Although, most definitely +1 @SimonCropp on a "useful renderer overrides" list.

[...]

PR #2043, opened 2021-05-08
feat: Custom Tokenizer/Renderer extensions

calculuschild:
## Description

New attempt at custom Tokens, Renderers for Marked.js. Faster than #1872.

May be a possible fix for: #1373, #1693, #1695, #2061,  

Users can add a custom Tokenizer and Renderer which will be executed without needing to overwrite existing components, with the extension in the following format:

```
const myExtension = {
  extensions: [{
    name: 'underline',
    start(src) { return src.match(/:/)?.index; },  // First characters of your token so Marked.js knows to stop and check for a match
    level: 'block',        // Is this a block-level or inline-level tokenizer?
    tokenizer(src, tokens) {
      const rule = /^:([^\n]*)(?:\n|$)/;  // Regex for the complete token
      const match = rule.exec(src);
      if (match) {
        return {                          // Token to generate
          type: 'underline',                // Should match "name" above
          raw: match[0],                    // The text that you want your token to consume from the source
          text: match[1].trim()             // Any custom properties you want the Renderer to access
        };
      }
    },
    renderer (token) {
      return `<u>${token.text}</u>\n`;
    }]
};
```

The extension(s) can then be loaded like so:

```
marked.use(myExtension, extension2, extension3);

\\ EQUIVALENT TO:

marked.use(myExtension);
marked.use(extension2);
marked.use(extension3);
```

##### Master from Feb 2021
```
es5 marked completed in 4260ms and passed 82.74%
es6 marked completed in 4289ms and passed 82.74%
es5 marked (gfm) completed in 4518ms and passed 82.13%
es6 marked (gfm) completed in 4504ms and passed 82.13%
es5 marked (pedantic) completed in 4596ms and passed 61.48%
es6 marked (pedantic) completed in 4783ms and passed 61.48%
commonmark completed in 3617ms and passed 100.00%
markdown-it completed in 3646ms and passed 89.21%
```

##### Master (current)
```
es5 marked completed in 4504ms and passed 86.90%
es6 marked completed in 4622ms and passed 86.90%
es5 marked (gfm) completed in 4776ms and passed 86.29%
es6 marked (gfm) completed in 5029ms and passed 86.29%
es5 marked (pedantic) completed in 4674ms and passed 71.03%
es6 marked (pedantic) completed in 5007ms and passed 71.03%
commonmark completed in 3678ms and passed 100.00%
markdown-it completed in 3672ms and passed 89.21%
```

##### This PR
```
es5 marked completed in 4761ms and passed 86.90%
es6 marked completed in 4814ms and passed 86.90%
es5 marked (gfm) completed in 5115ms and passed 86.29%
es6 marked (gfm) completed in 5105ms and passed 86.29%
es5 marked (pedantic) completed in 4972ms and passed 71.03%
es6 marked (pedantic) completed in 5069ms and passed 71.03%
commonmark completed in 3619ms and passed 100.00%
markdown-it completed in 3735ms and passed 89.21%
```

##### This PR with all extensions running at the top of the Lexer rather than using "before" to spread them out:
```
es5 marked completed in 4591ms and passed 86.90%
es6 marked completed in 4644ms and passed 86.90%
es5 marked (gfm) completed in 4891ms and passed 86.29%
es6 marked (gfm) completed in 4804ms and passed 86.29%
es5 marked (pedantic) completed in 4770ms and passed 71.03%
es6 marked (pedantic) completed in 4799ms and passed 71.03%
commonmark completed in 3617ms and passed 100.00%
markdown-it completed in 3587ms and passed 89.21%
```

## Contributor

## Committer


UziTech:
We should definitely only run through the extensions once and create a map like:

```js
this.extensions = {
  paragraph: [
    underline
  ]
}
```

Then looking them up like:

```js
runTokenzierExtension(src, tokens, before) {
  let tokensLength = 0;
  this.extensions[before].forEach(function(extension, index) {
    if (token = extension.tokenizer(src)) {
      tokens.push(token);
      tokensLength += token.raw.length;
    }
  });
  return tokensLength;
}
```

src/Parser.js
```diff
@@ -179,7 +181,20 @@ module.exports = class Parser {
           out += top ? this.renderer.paragraph(body) : body;
           continue;
         }
+
         default: {
+          // Run any renderer extensions
+          tokenParsed = false;
+          if (this.options.extensions) {
+            Object.values(this.options.extensions).forEach(function(extension, index) {
```
UziTech, line 189: we should also create a map with `name` as the key to prevent looping here as well.

calculuschild:
Used a similar method to your `inlineText` tokenizer on the `paragraph` tokenizer. Now if there are extensions, it looks for the "start" regex and cuts src down to that index before continuing on.

src/Parser.js
calculuschild: Ok. Shouldn't be too hard.

calculuschild:
@UziTech Is there a way to clear the extensions/options between unit tests?

calculuschild:
Added an "inline" version of my underline token and the extension works.

I was looking at the [header-ids](https://www.markdownguide.org/extended-syntax/#heading-ids) extension but I'm not sure how to best approach that. I see a few options:

1) Overwrite the Heading renderer (can already do this)
2) Overwrite the Heading tokenizer (can already do this)
3) Insert a custom Block HeadingID extension which runs before the normal Heading tokenizer (simple enough with this PR)
4) Have a custom Inline ID extension that we could apply more generally to whatever the parent token is (also add an id to an image, to a codespan, etc.) (not possible with this PR yet)

So this brings up some questions:
* Do we want to change the syntax for overwriting existing renderers/tokenizers to somehow follow this new style? (i.e., an optional parameter `overwrite : 'heading'` instead of `before : 'heading'`.)
* I would be interested in getting something like 4) above to work. There are existing markdown extensions that add some snippet of text inside of another block element to give it a particular style or html property, such as [image sizing](https://github.com/markedjs/marked/issues/339), or Pandoc's [fenced divs](https://pandoc.org/MANUAL.html#divs-and-spans) which is an example of re-using the `{#id}` syntax from the `header-ids` extension in different block types, and it would be neat if it didn't require fully overwriting each block element you want to apply the snippet to.

UziTech:
> Is there a way to clear the extensions/options between unit tests?

You should add `extensions: null` to the defaults it should be reset between tests.

UziTech:
> Do we want to change the syntax for overwriting existing renderers/tokenizers to somehow follow this new style? (i.e., an optional parameter overwrite : 'heading' instead of before : 'heading'.)

Good question. I'm not sure if it is better to have multiple ways to do the same thing. I think it would depend on whether it would slow down marked without extensions too much.

calculuschild:
> it would depend on whether it would slow down marked without extensions too much.

I was imagining just a change in the Marked.js page, to handle a different extension format but we would parse it into the same marked.defaults.renderer or whatever so speed should be similar when executing.

I would probably be in favor of just having one way to do things, and I also like the idea of having consistency between how extensions are formatted.

UziTech:
Sounds good. We could add this way of extensions and deprecate the old way and remove it in v3 or v4

src/Parser.js
calculuschild: This is done now.

calculuschild:
> We should definitely only run through the extensions once and create a map like:

@UziTech I applied all the changes discussed above. Processing in marked.js got a bit ugly (and I'm probably missing special cases/error handling, which I am not experienced with) but the code elsewhere is much simplified now without the looping and results in an `extensions` object of the following format:

```
extensions {
  [beforeA] : [tokenizer1, tokenizer2],
  [beforeB] : [tokenizer3],
  'last' : [tokenizer4],
  [nameA] : renderer1,
  [nameB] : renderer2,
  startBlock : /regex1|regex2|regex3/,
  startInline : /regex4|regex5/
```

Overwriting extensions now require existence of the property `overwrite : true`, and they will be merged into `defaults` as before.

Custom extensions with Tokenizers will be added to the `before` keyword `last` by default, and will run after all of the default tokenizers, unless the user supplies a valid `before` property. A valid `start` value is needed for inline tokenizers to work properly, and must have `start` for block tokenizers only if they want to be able to interrupt `paragraph`s, a la `tables`, `headers`, etc.

Custom extensions with Renderers must have a `name` property.

`start` regexes are now merged into one to avoid looping and to ensure the first match index is chosen.

test/unit/marked-spec.js
```diff
@@ -137,8 +137,89 @@ describe('parseInline', () => {
...
+            raw: match[0], // This is the text that you want your token to consume from the source
+            text: match[1].trim() // You can add additional properties to your tokens to pass along to the renderer
+          };
+        }
+      },
+      renderer: (token) => {
+        return `<u>${token.text}</u>`;
+      }
+    };
+    marked.use(underline);
```
UziTech, line 215: The extension should be added as an extension option.

```js
marked.use({
	extension: underline
	extension: [underline] // or as an array
});
```

That way it won't be a breaking change for `marked.use`,

UziTech: > The only breaking change was adding an overwrite flag to existing extensions

So this is a breaking change currently.

src/Lexer.js
```diff
@@ -229,8 +250,21 @@ module.exports = class Lexer {
         continue;
       }
 
+      if (tokensLength = this.runTokenzierExtension(src, tokens, 'paragraph')) {
+        src = src.substring(tokensLength);
+        continue;
+      }
```
UziTech, line 256: do we need this check before each tokenizer?

calculuschild: Yes we do. I just didn't apply it everywhere yet in case we want to keep changing things. Should I go ahead and copy it around?

calculuschild: Is it worth it to cache whether `this.options.extensions` exists in the constructor maybe? Is it faster to check `if(haveExtensions)` than `if(this.options.extensions)` ?

calculuschild: I copied this code everywhere now. Running benchmarks...

calculuschild:
Added current benchmarks to the OP. Not sure where the slowdown is coming from yet.

calculuschild:
Now that I think about it.... Do we NEED a `before` parameter? What if we just have all the custom extensions run one after another very first thing to cut down on the number of calls we need sprinkled throughout the Lexer? I can't actually think of a case where a user would want to add a custom tokenizers that specifically runs *after* the others.

src/Lexer.js
calculuschild: Moving the check does help with the slowdown some. ~ `4814ms` to `4718ms` on es6 commonmark spec.

UziTech:
Ya that sounds fine. If we need to allow running between other tokenizers we can add that later.

calculuschild:
Alright, all the fixes applied except for passing in parent tokens.

calculuschild:
> I would be interested in getting something like 4) above to work....
> ...I can try to cook up a version of the Header IDs extension that uses these as a test case.

Haha just kidding.

I'm quickly realizing that this requires being able to modify the HTML output that hasn't even been generated yet (i.e., somehow influence the output of the parent renderer which doesn't even execute until all the children are processed). Is it even possible?

So the example I was trying to do was a custom `inline` token of the form `{#myID}` that applies `id="myID"` to the HTML of the parent element, no matter what it is:
```
# My Header {#myID}
->
<h1 id="myID">My Header</h1>

\\ OR

My paragraph {#myID}
->
<p id="myID">My paragraph</p>
```

We would probably have to overwrite EVERY renderer that we want it to apply to, which is a nightmare for a user crazy enough to try it, but they could at least look at children in `token.tokens` and check for the presence of that child token. I can't think of any other way around it beyond just having the user make their own HTML post-processor to scan for their `{#myID}` and insert that ID tag in the right spot.

Unless there's an obvious way of doing this that I'm not seeing. Otherwise I'm happy with the capabilities of this PR as it is.

UziTech:
> Unless there's an obvious way of doing this that I'm not seeing.

I think the obvious way to do this would be to override the heading renderer and parse for this id and use that id instead of the slug id. That seems easier than putting it in the token anyways.

calculuschild:
> the obvious way to do this would be to override the heading renderer

Right, that's what I was trying to say. There's not a straightforward way to create a custom `{#id}` token that applies the ID generally to any parent token; you have to also overwrite the renderer for each type of parent token you want it to work for.

UziTech:
I just tested it out and this works pretty well already for the heading ids:

```js

marked.use({
  renderer: {
    heading(text, level, raw, slugger) {
      const headingIdRegex = /(?: +|^)\{#([a-z][\w-]*)\}(?: +|$)/i;
      const hasId = text.match(headingIdRegex);
      if (!hasId) {
        // fallback to original heading renderer
        return false;
      }
      return `<h${level} id="${hasId[1]}">${text.replace(headingIdRegex, '')}</h${level}>\n`;
    }
  }
});

console.log(marked("# has id {#custom-id}\n\n# no id"));
```

I think I am going to create a package for that extension.

UziTech:
Ya I think number 1 is the way to go for that.

For this PR in general I think it should be fine to just add custom tokenizers to the top of the list instead of having a `before` attribute. The current way to override a tokenizer/renderer is essentially adding it before that tokenizer/renderer since it will run the original tokenizer/renderer (or the last added one) if it returns false.

The only reason that doesn't work as a replacement for this PR is because each renderer is sent specific parameters instead of the whole token.

calculuschild:
Alright. So I think maybe just some more unit tests? Anything else we should do for this PR?

src/Lexer.js
```diff
@@ -230,7 +243,15 @@ module.exports = class Lexer {
       }
 
       // top-level paragraph
-      if (top && (token = this.tokenizer.paragraph(src))) {
+      // prevent paragraph consuming extensions by clipping 'src' to extension start
```
UziTech, line 246: The only problem with blocking the paragraph tokenizer for block tokens this way is that the start will have to be the actual start of a different block element otherwise it will break a single paragraph into multiple paragraphs.

Blocking inlineText for inline tokens this way doesn't really matter because multiple inline text tokens render the same as a single inline text token.

For Example:

We want to create an underline block token when a line is surrounded by `:` characters

The potential start is `/^:/`

```markdown
first line
:underlined line:
last line
```

```html
<p>first line</p>
<u>underlined line</u>
<p>last line</p>
```

The problem is when it is given a potential start that is not an actual start

```markdown
first line
:no underlined line
last line
```

```html
<p>first line</p>
<p>:no underlined line
last line</p>
```

where it should output

```
<p>first line
:no underlined line
last line</p>
```

src/Parser.js
```diff
@@ -179,7 +180,16 @@ module.exports = class Parser {
...
         }
+
         default: {
+          // Run any renderer extensions
+          tokenParsed = false;
+          if (this.options.extensions?.[token.type]) {
+            out += this.options.extensions[token.type](token);
+            tokenParsed = true;
+          }
+          if (tokenParsed) continue;
```
UziTech, line 191:
```suggestion
          // Run any renderer extensions
          if (this.options.extensions?.[token.type]) {
            out += this.options.extensions[token.type](token);
            continue;
          }
```

src/Lexer.js
calculuschild: Hm good catch. Can we just use the full token regex then instead of a start character? That's how `rules.js` does it. We have the paragraph regex search for an entire table, fence, etc, not just the start characters.

calculuschild: Alternatively, we can make the `Paragraph` step in the Lexer behave the same as `inlineText` where adjacent tokens are merged:

```
lastToken = tokens[tokens.length - 1];
        if (lastToken && lastToken.type === 'text') {
          lastToken.raw += token.raw;
          lastToken.text += token.text;
        } else {
          tokens.push(token);
        }
```

This solves the problem for me except it causes spec test `new smartypants_code` to fail, which actually might not be a correct test. That test shows two adjacent lines of text rendering as separate paragraphs for some reason:

[new smartypants_code odd behavior](https://marked.js.org/demo/?text=%23%23%23%23%23%20The%20lexer%20shows%20the%20following%20text%20as%20two%20paragraphs%3B%20The%20text%20%60script%60%20in%20%60%3Cscript-custom%3E%60%20is%20throwing%20off%20the%20parser.%0A%0A%0AEnsure%20that%20text%20such%20as%20custom%20tags%20that%20happen%20to%0Abegin%20with%20the%20same%20letters%20as%20the%20above%20tags%20don%27t%0Amatch%20and%20thus%20benefit%20from%20Smartypants-ing.%0A%3Cscript-custom%3E--foo%3C%2Fscript-custom%3E%0A%60--foo%60%20%3Ccodebar%20--foo%20codebar%3E%0A%0A%0A%23%23%23%23%23%20Change%20it%20to%20%60scrit%60%20and%20we%20get%20a%20single%20paragraph%2C%20which%20should%20be%20the%20correct%20behavior.%0A%0AEnsure%20that%20text%20such%20as%20custom%20tags%20that%20happen%20to%0Abegin%20with%20the%20same%20letters%20as%20the%20above%20tags%20don%27t%0Amatch%20and%20thus%20benefit%20from%20Smartypants-ing.%0A%3Cscrit-custom%3E--foo%3C%2Fscript-custom%3E%0A%60--foo%60%20%3Ccodebar%20--foo%20codebar%3E%0A&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20true%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)

src/Parser.js
calculuschild: Good catch. Right, we shouldn't have multiple Renderers with the same `name`, so this is fine. That also might be something to make sure we clarify in the documentation.

This is fixed now.

src/Lexer.js
UziTech: I would think requiring a blank link before a block token extension should be good enough for now.

src/marked.js
```diff
@@ -139,44 +139,84 @@ marked.defaults = defaults;
...
+
+    //= =-- Parse "overwrite" extensions --==//
+    if (ext.renderer && !ext.name) {
+      const renderer = marked.defaults.renderer || new Renderer();
+      for (const prop in ext.renderer) {
+        const prevRenderer = renderer[prop];
+        renderer[prop] = (...args) => {
+          let ret = ext.renderer[prop].apply(renderer, args);
+          if (ret === false) {
+            ret = prevRenderer.apply(renderer, args);
```
UziTech, line 182: if `ext.renderer` is a new renderer than `prevRenderer` will be `undefined` and this will fail.

We should make sure there is a unit test for a new renderer returning false. It should probably just render `''`.

UziTech: same for the tokenizer

src/Lexer.js
calculuschild: I don't know... It's such a common thing in the spec to be able to interrupt paragraphs. I feel like it's a pretty important feature, and with the solution above it's not hard to fix (and even corrects broken behavior in that unit test).

src/marked.js
```diff
@@ -139,44 +139,84 @@ marked.defaults = defaults;
...
+        extensions[ext.level] = [ext.tokenizer];
+      }
+      if (ext.start) { // Regex to check for start of token
+        if (ext.level === 'block') {
+          extensions.startBlock = extensions.startBlock
+            ? new RegExp(extensions.startBlock.source + '|' + ext.start.source)
+            : ext.start;
+        } else if (ext.level === 'inline') {
+          extensions.startInline = extensions.startInline
+            ? new RegExp(extensions.startInline.source + '|' + ext.start.source)
```
UziTech, line 168: We will have to make sure this isn't a source of vulnerabilities. I'm kind of thinking `extensions.startInline` should be an array of functions/regexps. It might be a little slower if there are many extensions but it seems like it would be more secure and we aren't really guaranteeing speed when using extensions.

src/marked.js
```diff
@@ -139,44 +139,84 @@ marked.defaults = defaults;
...
+  }
+  const opts = merge({}, ...extension);
+  opts.tokenizer = null;
+  opts.renderer = null;
+  const extensions = {};
+
+  extension.forEach((ext) => {
+    //= =-- Parse "addon" extensions --==//
+    if (ext.renderer && ext.name) { // Renderers must have 'name' property
+      extensions[ext.name] = ext.renderer;
```
UziTech, line 153: Maybe this should be on a `renderer` property so if `ext.name` is `block` it won't overwrite the block level tokenizers.

```suggestion
      extensions.renderers[ext.name] = ext.renderer;
```

src/marked.js
UziTech: and it would allow using means other than regexp to find the potential start

calculuschild: Any guidelines on implementing this so it's secure? I'm not sure how separately calling regexes from an array is more secure than `or`ing them together or if there's a certain approach that would be more secure.

How much do we need to worry about security on user-made extensions like these in general? Any time the dev is constructing their own regexes like this, there's risk of exposing their own app to redos or whatever, and at some point that's out of our control.

calculuschild: Also how are you envisioning this working with non-regex? Can you help me with examples so I can implement this correctly?

UziTech: > I'm not sure how separately calling regexes from an array is more secure than `or`ing them together or if there's a certain approach that would be more secure.

There are regexps that are fine separately but redos when together

For Example: `/(a*)*/` and `/b/` do not backtrack when run separately but `/(a*)*b/` does backtrack.

I don't know if there are regexps that are insecure when appended together with `|` but I also don't know that there aren't any.

> Any time the dev is constructing their own regexes like this, there's risk of exposing their own app to redos or whatever

The devs aren't "constructing their own regexes like this", that is part of the problem. If a dev uses the following code:

```js
const markedExtension1 = require('marked-extension-1')
const markedExtension2 = require('marked-extension-2')
marked.use([markedExtension1, markedExtension2])
```

The user might not even know that a regexp is being constructed when using those extensions. Also, as shown above, the extension authors might not be creating a vulnerable regexp but the act of combining them creates one by accident.

UziTech: > Also how are you envisioning this working with non-regex?

Any function that takes the `src` and returns an index of the first possible start. Something like:

```
start: (src) => src.indexOf(":")
```

src/Lexer.js
UziTech: html with the `sanitize` option currently creates a paragraph token instead of html so it would break

```markdown
test
<div></div>
```

[demo](https://marked.js.org/demo/?text=test%0A%3Cdiv%3E%3C%2Fdiv%3E&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20true%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20true%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)

UziTech: > It's such a common thing in the spec to be able to interrupt paragraphs.

This is one of those areas where the CommonMark spec uses the spec to describe how CommonMark works not how it should work.

According to the original markdown [spec](https://daringfireball.net/projects/markdown/basics):

> A paragraph is simply one or more consecutive lines of text, separated by one or more blank lines. (A blank line is any line that looks like a blank line — a line containing nothing but spaces or tabs is considered blank.)

src/marked.js
calculuschild: Ok that makes sense. I'll see if I can get that working. Should we just require `start` to be a function then to simplify our implementation? Then users can choose to use regex or string methods internally and we don't have to be type-checking their input?

UziTech: Ya that sounds good

src/Lexer.js
calculuschild: > CommonMark spec uses the spec to describe how CommonMark works not how it should work.

Right, I get that. But for better or for worse, being able to interrupt paragraphs is the behavior a lot of users are probably used to and expect because Commonmark has taken over as more or less the standard. Even Github continued the tradition when they introduced tables that also interrupt paragraphs

So the question is do we let users create "Commonmark" extensions, or do we restrict them to "original markdown" style?

(I personally really want to make some extensions that use this style, so I'm biased...)

In any case, I tweaked the `paragraph` clipping method so now it only merges paragraphs if the previous one was affected by the clipping-to-'start' process. It works with your sanitize tag example now.  Take a look....? 😅

UziTech: That still wouldn't help if `<` was the potential start of an extension. It would think that `test` was clipped and combine it with `<div></div>`.

Any extensions are, by definition, not "CommonMark" extensions because they are not in the CommonMark spec. GitHubs extensions would be considered "GFM" extensions. I don't see why any block extension can't have a blank line above it. Can you give me some examples of extensions that wouldn't work if we required a blank line?

UziTech: I think the only way to make it work with my sanitize tag example would be to add a `space` token before the html `paragraph` token if the last token was a `paragraph`.

I'm not sure if that would be a breaking change. I doubt anyone relies on there not being a `space` token there.

UziTech: or that could be another example of an error that would be fixed by combining `paragraph` tokens? There is no spec for sanitized html in markdown.

calculuschild: > ...examples of extensions that wouldn't work if we required a blank line?

Say a user wants to make a new type of `table` that uses a different syntax than a normal GFM table. So they insert a custom `fancyTables` Tokenizer. And it *works*, until they start replacing their old tables with the custom Markdown, and they realize that half of them don't work. Because a blank line is missing. "But GFM tables don't need a blank line...? What is going on here?"

It's not that the extension "doesn't work", but that for the user it doesn't work "like the rest of my Markdown". In my own use case, for example, my business documents already follow a particular format that takes advantage of being able to place paragraphs directly above or below tables, blocks of code, etc. as captions or introductory text, which for us is very handy for legibility when skimming through the source (Markdown is designed to be somewhat human readable on its own). If adding extensions means abandoning our existing template, it's a bit of a nuisance.

So yeah, not that the extension doesn't work, but that the formatting is broken. I hope that makes sense.

If it comes down to it, we can push without that feature, but I would still like to see it \*someday\*...?

UziTech:
We still need a way for extensions to tokenize their `token.text` property (or any other property) and let the lexer know if it should be block tokens (like a blockquote extension) or inline tokens.

And a way for the renderer to render the child tokens.

calculuschild:
> We still need a way for extensions to tokenize their `token.text` property (or any other property)

In one of my unit tests I just did this:

https://github.com/markedjs/marked/blob/7bf58eb5e218fcf75964cd02d952914b85625093/test/unit/marked-spec.js#L233-L235

src/marked.js
calculuschild: Done.

src/marked.js
calculuschild: This should be blocked by the if statement above, right? If we have a new renderer without a `name` property, then we don't even enter this section.

https://github.com/markedjs/marked/blob/7bf58eb5e218fcf75964cd02d952914b85625093/src/marked.js#L175

https://github.com/markedjs/marked/blob/7bf58eb5e218fcf75964cd02d952914b85625093/src/marked.js#L189

calculuschild: Added a unit test with a new renderer that returns false. The parser will use `''` if the return value from the renderer is falsy.

UziTech:
The only problem is each extension won't know what options or other extensions to use.

calculuschild:
I'm not sure I follow.  Wouldn't each extension just by default use all the currently loaded options and extensions? By the time `parse.inline()` executes inside the custom renderer, `marked.use()` has already loaded up all extensions and options, right?

UziTech:
Not for options given when calling marked `marked(markdown, options)` those options aren't loaded into defaults.

Also we would want the tokens in a `.tokens` parameter so `walkTokens` works. Which makes me think the extensions should have a function so `walkTokens` has a way to know where to go. e.g. for a `list` extention `walkTokens` should walk down `items` property.

calculuschild:
That unit test above has two extensions, one inline and one block and they both call `parseInline` and interact just fine?

https://github.com/markedjs/marked/blob/7bf58eb5e218fcf75964cd02d952914b85625093/test/unit/marked-spec.js#L217-L267

UziTech:
Also I would like to stop marked changing defaults some time in the future and make marked create instances. That would solve quite a few of the open issues.

calculuschild:
> Not for options given when calling marked `marked(markdown, options)` those options aren't loaded into defaults.
> 
> Also we would want the tokens in a `.tokens` parameter so `walkTokens` works.

I might need you to take the reins on that. This is getting pretty tangled into the other parts of Marked that I'm not as familiar with and I imagine you have something in mind for this? Probably easier than me trying to guess at a good solution that's going to break more stuff.

UziTech:
> and I imagine you have something in mind for this?

Unfortunately I don't. You are not the first person to try to tackle adding extensions to marked (although I think you did get the farthest). Me breaking the tokenizer out of the lexer was the first step to creating better extensions.

> Is it as simple as having the tokenizer run marked.lexer(text)?

Maybe we could send the tokenizer `lexer.blockTokens`, `lexer.inline`, and `lexer.inlineTokens`. They would just need to know when to use which function.

UziTech:
> Is there a good reason marked(markdown, options) behaves differently than marked.use(options); marked(markdown)?

Yes it is so the options given can be temporary and don't clobber defaults for other parts of the program. Again instances would solve this but currently that would be a major breaking change.

calculuschild:
> Maybe we could send the tokenizer lexer.blockTokens, lexer.inline, and lexer.inlineTokens. They would just need to know when to use which function.

Could we tell users to use `marked.lexer.blockTokens()` directly so we don't have to find a way to pass functions around like that? Would the behavior be any different?

UziTech:
> Would the behavior be any different?

I don't think it would have the options passed in to `marked.lexer(markdown, options)`

calculuschild:
Ok, so is this just blocked until instances are implemented?

UziTech:
> is this just blocked until instances are implemented?

No if we pass in `blockTokens` and other functions to the custom tokenizer those should have the options. Or we could change the call to the custom tokenizer so `this` is the current instance of the lexer.

```diff
- if (token = extTokenizer(src, tokens)) {
+ if (token = extTokenizer.call(this, src, tokens)) {
```

then they could use `this.blockTokens` and other functions.

calculuschild:
I like the second option.

UziTech:
For the `walkTokens` we could just make the extension have a `walkTokens` function that knows how to walk the child tokens if there are any.

src/marked.js
calculuschild: Done.

calculuschild:
> we could change the call to the custom tokenizer so this is the current instance of the lexer.

Tried doing this and realized it means we can't use arrow functions in the extension because `this` won't get picked up properly.

```
const arrowExtension = {
      name: 'test',
      level: 'block',
      tokenizer: (src, tokens) => {
        console.log(this);                 // this = window
      }
}

const funcExtension = {
      name: 'test',
      level: 'block',
      tokenizer: function(src, tokens) {
        console.log(this);                 // this = lexer
      }
}
```

Which would you prefer? Tell users we require non-arrow functions so we can do `call`, or just pass in the lexer functions as you first suggested? Or just pass the whole lexer in?

UziTech:
With the object method syntax it is more terse than arrow functions anyway. Since we do this for renderer and tokenizer I think it would be ok to assign `this`.

```js
const funcExtension = {
      name: 'test',
      level: 'block',
      tokenizer(src, tokens) {
        console.log(this);                 // this = lexer
      }
}
```

calculuschild:
> object method syntax 

Ok, this is done now.

> make the extension have a walkTokens function

Did this too, but named it `walkableTokens` since it returns an array of tokens that should be walked, to distinguish it from the existing `walkTokens` extensions which define the walkTokens callback function. Any `.tokens` will be walked by default.

```
if (this.options.extensions?.walkableTokens?.[token.type]) { // Walk any extensions
  marked.walkTokens(this.options.extensions.walkableTokens[token.type], callback);
}
if (token.tokens) {
  marked.walkTokens(token.tokens, callback);
}
```

I could use some help coming up with a `walkTokens` unit test though.

UziTech (changes requested):
What does a package do if it wants to provide multiple extensions?

src/Lexer.js
```diff
@@ -123,9 +123,22 @@ module.exports = class Lexer {
...
+        tokensParsed = false;
+        this.options.extensions.block.forEach((extTokenizer) => {
+          if (token = extTokenizer.call(this, src, tokens)) {
+            src = src.substring(token.raw.length);
+            tokens.push(token);
+            tokensParsed = true;
+          }
+        });
+        if (tokensParsed) { continue; }
+      }
```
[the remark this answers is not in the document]
calculuschild: Ah good catch. I can change that.

src/marked.js
```diff
@@ -204,6 +251,9 @@ marked.walkTokens = function(tokens, callback) {
         break;
       }
       default: {
+        if (this.options.extensions?.walkableTokens?.[token.type]) { // Walk any extensions
+          marked.walkTokens(this.options.extensions.walkableTokens[token.type], callback);
+        }
```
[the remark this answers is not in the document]
calculuschild: > Is walkableTokens supposed to be the list of tokens or the names of the properties for tokens?

Currently It's just a list of tokens, but I'm seeing now that you can't define a list of tokens in the extension since the tokens are not generated at that point. I will change it to a list of properties of the token.

> Can you add a test with walkable tokens?

I could use some help coming up with a walkTokens unit test. I never really use that feature and I'm not sure what good use cases/bugs we need to catch.

> If the walkable tokens are in the tokens property will they be walked twice?

Currently yes. I can fix that so it only walks `tokens` once.

UziTech:
> What does a package do if it wants to provide multiple extensions?

I think this is a good reason to provide an `extensions` option instead of adding all of the properties of an extension to the top level object.

Essentially instead of:

```js
{
  name: "underline",
  tokenizer() {...},
  renderer() {},
}
```

it would be:

```js
{
  extensions: [
    {
      name: "underline",
      tokenizer() {/* new underline tokenizer */},
      renderer() {/* new underline renderer */},
    },
    {
      name: "overline",
      tokenizer() {/* new overline tokenizer */},
      renderer() {/* new overline renderer */},
    },
  ],
  tokenizer: {
    heading() {/* overwrite heading tokenizer */}
  }
}
```

This also provides the benefit of not having to explain how the `tokenizer` option can be an object, or a function if `name` is provided as a new extension in the docs.

src/marked.js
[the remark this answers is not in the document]
calculuschild: Alright, these are all fixed now.

calculuschild:
> What does a package do if it wants to provide multiple extensions?

I was assuming we could just use this which currently works:

```
package returns:
[
  {
    name: "underline",
    tokenizer() {/* new underline tokenizer */},
    renderer() {/* new underline renderer */},
  },
  {
    name: "overline",
    tokenizer() {/* new overline tokenizer */},
    renderer() {/* new overline renderer */},
  },
  {
    tokenizer: {
    heading() {/* overwrite heading tokenizer */}
  }
]

marked.use(package);
```

But I can try changing it around if we need to.

UziTech:
> I was assuming we could just use this which currently works:

Then they would need to call `marked.use` multiple times when loading multiple packages. I thought you wanted it to be an array so they could call `marked.use` once with an array of packages.

```js
const extension1 = require("marked-extension-1");
const extension1 = require("marked-extension-1");

**Comment by calculuschild** [comment]
> I thought you wanted it to be an array so they could call `marked.use` once with an array of packages.

Yeah I did... Anyway I already made the change to use `{extensions: ArrayOfExtensions}`. Check it out.

File: src/marked.js [hunk 11]
```diff
@@ -139,73 +139,132 @@ marked.defaults = defaults;
...
-    const walkTokens = marked.defaults.walkTokens;
-    opts.walkTokens = (token) => {
-      extension.walkTokens(token);
-      if (walkTokens) {
-        walkTokens(token);
-      }
-    };
+  });
+
+  if (Object.keys(extensions).length) {
```
UziTech, line 228: This will always be `true` since `renderers` and `walkableTokens` are added automatically. We could probably set a boolean in the `if (pack.extensions)` block.

src/marked.js
```diff
@@ -139,73 +139,132 @@ marked.defaults = defaults;
...
+
   marked.setOptions(opts);
 };
 
 /**
  * Run callback for every token
  */
 
-marked.walkTokens = function(tokens, callback) {
+marked.walkTokens = function(tokens, callback, opt) {
```
UziTech, line 239: It is pretty standard to have the callback as the last parameter. I think we should use `marked.defaults` instead of requiring passed in options.

src/marked.js
```diff
@@ -139,73 +139,132 @@ marked.defaults = defaults;
...
       }
       case 'list': {
-        marked.walkTokens(token.items, callback);
+        marked.walkTokens(token.items, callback, opt);
         break;
       }
       default: {
+        if (opt?.extensions?.walkableTokens?.[token.type]) { // Walk any extensions
+          opt.extensions.walkableTokens[token.type].forEach(function(walkableTokens) {
+            if (walkableTokens !== 'tokens') {
```
UziTech, line 261: I think we would want to walk them in the order they give so we should skip the `tokens` below instead of here.

src/marked.js
calculuschild: Fixed.

src/marked.js
calculuschild: Fixed

src/marked.js
calculuschild: Fixed.

src/Lexer.js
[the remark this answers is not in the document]
calculuschild: Fixed now.

UziTech (changes requested):

src/marked.js
```diff
@@ -139,44 +139,98 @@ marked.defaults = defaults;
...
-        if (ret === false) {
-          ret = prevRenderer.apply(renderer, args);
+  if (!Array.isArray(extension)) { // Wrap in array if not already to unify processing
+    extension = [extension];
+  }
+  const opts = merge({}, ...extension);
+  opts.tokenizer = null;
+  opts.renderer = null;
+  opts.extensions = null;
+  const extensions = { renderers: {}, walkableTokens: {} };
```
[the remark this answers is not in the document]
UziTech: I added a test that uses multiple extensions in an array and calling `marked.use` multiple times and it looks like different things are getting overwritten depending on how multiple extensions are loaded.

src/Lexer.js
```diff
@@ -230,9 +242,27 @@ module.exports = class Lexer {
       }
 
       // top-level paragraph
-      if (top && (token = this.tokenizer.paragraph(src))) {
+      // prevent paragraph consuming extensions by clipping 'src' to extension start
+      cutSrc = src;
+      if (this.options.extensions?.startBlock) {
+        let startIndex = Infinity;
+        this.options.extensions.startBlock.forEach(function(getStartIndex) {
+          startIndex = Math.max(0, Math.min(getStartIndex(src), startIndex));
```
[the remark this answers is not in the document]
UziTech: Also it looks like we should trim the first character before passing in the `src`.

if `src` === `:extension 1\n:extension2` and the start function returns `src.indexOf(':')` we want it to return the index for the second `:` not the first one at index 0.

We would have to add 1 to the output and check that it is > 0.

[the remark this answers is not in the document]
UziTech: I'm also wondering if we should make this more resilient and disregard the return if it is not a number. Currently I think `Math.min` would coerce the return value to a number which could produce surprising results if say `true` was returned. (which `Math.min` would coerce to a `1`)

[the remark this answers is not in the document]
calculuschild: Working on this...

src/marked.js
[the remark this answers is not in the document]
calculuschild: > I added a test that uses multiple extensions in an array

I'm working on this, but I see a possible error in your block extension. Just want to make sure I'm testing the extension as it is intended. This:

```
renderer(token) {
  return `<${token.type}>${token.text}</${token.type}>\n`;
}
```
should be this:
```
renderer(token) {
  return `<${token.type}>${this.parseInline(token.tokens)}</${token.type}>\n`;
}
```
Otherwise the inline tokens are generated but never rendered and we get `<block-extension1>:extension1</block-extension1>` instead of `<block-extension1>used extension1 walked</block-extension1>`. Also we probably need to turn off the `headerIds` option because they are being generated in the test which doesn't match the expected output.

Is this what you intended? I want to make sure I'm fixing the right bug.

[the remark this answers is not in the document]
UziTech: Yes. nice catch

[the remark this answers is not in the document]
calculuschild: As for the extensions that overwrite default tokenizers, I imagine the latest one to be loaded should overwrite the first one. i.e. this:
https://github.com/markedjs/marked/blob/64cdfa2e88131404e077c9e56bbc7425630263ac/test/unit/marked-spec.js#L395-L414
will only apply to `# extension2`  because when extension 2 is loaded, `name` is overwritten and won't match `# extension1`.

So expected output for 
```
# extension1

# extension2

# no extension
```
becomes:
```
<h1>extension1</h1>   // overwritten by extension2
<h1>used extension2 walked</h1>
<h1>no extension</h1>
```

... Or should `# extension1` and `# no extension` even be translated at all? Since the heading tokenizer is completely overwritten to only match `# extension 2`?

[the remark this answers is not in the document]
UziTech: The heading tokenizer in extension1 shouldn't be overwritten. If extension 2 heading tokenizer returns `false` than it should drop back to extensions 1 heading tokenizer, and if extension 1 heading tokenizer returns false it should drop back to the original heading tokenizer.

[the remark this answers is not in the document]
calculuschild: Oh what! I didn't realize it left previous tokenizers intact. Ok that is good to know.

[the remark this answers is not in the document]
calculuschild: Wait where is the "fallback" logic in the code? Can you point me to it?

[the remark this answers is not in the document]
UziTech: https://github.com/markedjs/marked/blob/29a301bf8ce5b6bf3524eb49efb0e1e669982303/src/marked.js#L157-L170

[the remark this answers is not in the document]
calculuschild: Ah wait I see I see... I could not tell exactly what that bit of code was doing but now I see it replaces the default tokenizer with a function that first uses the extension and then if it fails, runs again on the previous tokenizer, which itself could be another function depending how many extensions have been loaded that overwrite the same tokenizer.

Oh my gosh this changes everything. I totally missed that this had a fallback baked in. Oh my gosh. This solves so many problems with my own projects. Oh my gosh.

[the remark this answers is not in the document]
calculuschild: Done!

calculuschild:
Could you help me figure out the merge conflicts here?

UziTech:
Those merge conflicts are in the built files. You should be able to rebase on master then run `npm run build` to remove the merge conflicts.

src/marked.js
```diff
@@ -139,45 +139,99 @@ marked.defaults = defaults;
...
+  const opts = merge({}, ...extension);
+  const extensions = marked.defaults.extensions || { renderers: {}, walkableTokens: {} };
+  let hasExtensions;
+
+  extension.forEach((pack) => {
+    // ==-- Parse "addon" extensions --== //
+    if (pack.extensions) {
+      hasExtensions = true;
+      pack.extensions.forEach((ext) => {
+        if (ext.renderer && ext.name) { // Renderers must have 'name' property
```
UziTech, line 155: I think all extensions should have a `name` property so it should probably throw an error if an extension doesn't.

src/marked.js
```diff
@@ -139,45 +139,99 @@ marked.defaults = defaults;
...
+          extensions.renderers[ext.name] = ext.renderer;
```
UziTech, line 156: Do we want to fallback to extension renderer with same name if exists on `false`. I don't know if we need to but thought I'd put the idea out there. Same with extension tokenizer.

src/marked.js
UziTech: Also what happens if `name` already exists or is one of the original tokenizer/renderer names?

src/marked.js
```diff
@@ -139,45 +139,99 @@ marked.defaults = defaults;
...
-        let ret = extension.tokenizer[prop].apply(tokenizer, args);
-        if (ret === false) {
-          ret = prevTokenizer.apply(tokenizer, args);
+
+    // ==-- Parse "overwrite" extensions --== //
+    if (pack.renderer) {
+      const renderer = marked.defaults.renderer || new Renderer();
+      for (const prop in pack.renderer) {
+        const prevRenderer = renderer[prop];
+        // Replace renderer with func to run extension, but fall back if fail
```
UziTech, line 191: It falls back if the renderer returns `false` not if it throws an error.

```suggestion
        // Replace renderer with func to run extension, but fall back if false
```

src/marked.js
```diff
@@ -204,7 +258,12 @@ marked.walkTokens = function(tokens, callback) {
         break;
       }
       default: {
-        if (token.tokens) {
+        if (marked.defaults?.extensions?.walkableTokens?.[token.type]) { // Walk any extensions
+          marked.defaults?.extensions.walkableTokens[token.type].forEach(function(walkableTokens) {
+            marked.walkTokens(token[walkableTokens], callback);
+          });
+        }
+        if (token.tokens && !marked.defaults?.extensions?.walkableTokens[token.type]?.tokens) {
```
UziTech, line 266: Should we still walk `tokens` if it isn't in `walkableTokens`? Could that property mean something else for that token or should we make that property always walkable?

UziTech (changes requested):

src/marked.js
calculuschild: I don't know. What behavior do we want?

src/marked.js
calculuschild: Maybe. In case multiple packages happen to choose the same extension name?

src/marked.js
calculuschild: I think if it's the same as an existing one, the Tokenizer extension would work fine but the renderer would never trigger because the default renderer should consume any tokens with that name first.

src/marked.js
calculuschild: My instinct was always walk `tokens`. We can say so in the documentation, and then users don't have to bother with making a walkabletokens if their extension is simple enough.

UziTech: ya I think we should walk `tokens` if they don't provide `walkableTokens` but if they do create `walkableTokens` I think it would be more intuitive if we just walk those properties even if they have a `tokens` property that isn't in `walkableTokens`. If they have `walkableTokens` and they don't add `tokens` I would guess it would be because they don't want it walked.

src/marked.js
UziTech: Ya ! could see having an extension that provides a special case for another extension.

src/Lexer.js
```diff
@@ -230,9 +242,30 @@ module.exports = class Lexer {
...
       // top-level paragraph
-      if (top && (token = this.tokenizer.paragraph(src))) {
+      // prevent paragraph consuming extensions by clipping 'src' to extension start
+      cutSrc = src;
+      if (this.options.extensions?.startBlock) {
+        let startIndex = Infinity;
+        const tempSrc = src.slice(1);
+        let tempStart;
+        this.options.extensions.startBlock.forEach(function(getStartIndex) {
+          tempStart = getStartIndex(tempSrc);
```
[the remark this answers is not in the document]
UziTech: furthermore I know that Array.forEach is a lot slower than a for loop. I don't think that matters a whole lot since we are not guaranteeing speed with extensions but it is something to think about.

src/marked.js
UziTech: > the Tokenizer extension would work fine

I think currently it would run the extension tokenizer and the original if an extension contained the same name as a current tokenizer.

> It probably makes sense to allow the user to write an overwrite extension using the new structure so they don't have to juggle remembering multiple different styles.

I agree

src/marked.js
UziTech: Can we rename `walkableTokens` to `childTokens` or something like that?

calculuschild: Yeah we can do that.

src/Lexer.js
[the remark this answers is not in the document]
calculuschild: > Array.forEach is a lot slower than a for loop

Ok. I'm immediately worried since I don't expect this array to ever get that huge but I might take a look through all the loops I added and switch them out later.

src/marked.js
```diff
@@ -105,7 +105,7 @@ function marked(src, opt, callback) {
   try {
     const tokens = Lexer.lex(src, opt);
     if (opt.walkTokens) {
-      marked.walkTokens(tokens, opt.walkTokens);
+      marked.walkTokens(tokens, opt.walkTokens, opt);
```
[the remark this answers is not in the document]
calculuschild: Done.

src/Lexer.js
```diff
@@ -459,7 +505,18 @@ module.exports = class Lexer {
...
+      cutSrc = src;
+      if (this.options.extensions?.startInline) {
+        let startIndex = Infinity;
+        this.options.extensions.startInline.forEach(function(getStartIndex) {
+          startIndex = Math.max(0, Math.min(getStartIndex(src), startIndex));
+        });
+        if (startIndex < Infinity && startIndex > 0) {
+          cutSrc = src.substring(0, startIndex);
+        }
+      }
```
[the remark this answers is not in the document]
calculuschild: Done.

src/Lexer.js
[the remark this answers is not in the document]
calculuschild: Done.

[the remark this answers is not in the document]
UziTech: > I don't expect this array to ever get that huge

true

src/marked.js
calculuschild: > I think currently it would run the extension tokenizer and the original

Yes.

> It probably makes sense to allow the user to write an overwrite extension using the new structure

In that case do we also need to require that they provide the correct `level: 'block' or 'inline`'? What's the preferred output if we get invalid syntax like that? Or if there's no `name` given on a normal extension? Do we give an error or just ignore the extension?

src/marked.js
calculuschild: Ok renamed to childTokens and updated a unit test to show `token.tokens` is only walked by default if `childTokens` isn't provided.

src/marked.js
UziTech: Definitely throw an error. It will help with debugging. We should be as descriptive as we can. Like ` extension name required` and `extension level must be 'block' or 'inline'`

calculuschild:
Check it out. I finagled a way to use a curly bracket tag as discussed above to modify the HTML output of the parent token a la: `# Header {something}` -> `<h1 'something'>Header</h1>` without having to overwrite the renderer or tokenizer for the parent. This makes it apply to any parent token no matter what type! Similar to this [kramdown syntax](https://kramdown.gettalong.org/quickref.html#block-attributes).

Can you see anything unsafe about allowing that type of manipulation?

UziTech:
> Can you see anything unsafe about allowing that type of manipulation?

It can only be manipulated by extensions not malicious markdown so as long as marked users don't allow untrusted extensions there shouldn't be a problem.

src/marked.js
calculuschild: I set the renderers to fall back if there is another extension with the same name.

The tokenizers however are just piled into an array `extensions.block` or `extensions.inline` and executed in array order. This means if a tokenizer returns false, it will just try the next tokenizer in the array anyway. I did change the append order from `extensions.block.push(newTokenizer)` to `extensions.block.unshift(newTokenizer)` so extensions later in the load order are given first priority.

So instead of falling back directly, tokenizers will just rely on load order, which seems logical to me. Assume `extension1_false` has the same `name` as `extension1` but always returns false:
```
marked.use([extension1, extension2, extension1_false]); \\Latest load gets priority

// In Lexer.js:
if (token = extension1_false.call(this, src, tokens)) {... 
\\ returns false

if (token = extension2.call(this, src, tokens)) {... 
\\ doesn't match

if (token = extension1.call(this, src, tokens)) {...
\\ Matches
```

src/marked.js
```diff
@@ -139,45 +139,111 @@ marked.defaults = defaults;
...
+  const extensions = marked.defaults.extensions || { renderers: {}, childTokens: {} };
+  let hasExtensions;
+
+  extension.forEach((pack) => {
+    // ==-- Parse "addon" extensions --== //
+    if (pack.extensions) {
+      hasExtensions = true;
+      pack.extensions.forEach((ext) => {
+        if (ext.renderer && ext.name) { // Renderers must have 'name' property
+          const prevRenderer = extensions.renderers?.[ext.name] || null;
```
UziTech, line 156: ` || null` seems redundant

```suggestion
          const prevRenderer = extensions.renderers?.[ext.name];
```

src/marked.js
```diff
@@ -139,45 +139,111 @@ marked.defaults = defaults;
...
+        };
+      }
+      opts.tokenizer = tokenizer;
+    }
+
+    // ==-- Parse WalkTokens extensions --== //
+    if (pack.walkTokens) {
+      const walkTokens = marked.defaults.walkTokens;
+      opts.walkTokens = (token) => {
+        pack.walkTokens(token);
```
UziTech, line 234: We should probably allow `walkTokens` functions to see the options through `this.options` just to be similar to other extension functions.

```suggestion
        pack.walkTokens.call(this, token);
```

calculuschild: done.

src/marked.js
calculuschild: done.

calculuschild:
> It probably makes sense to allow the user to write an overwrite extension using the new structure so they don't have to juggle remembering multiple different styles.

Still working on this one but I think I have everything else up to this point.

src/marked.js
calculuschild: Now throws errors for missing `name` on all extensions or invalid `level` on tokenizer extensions.

calculuschild:
Alright, users can now add extensions with the same name as an existing function and it will fall back to the original if it returns false.

I couldn't directly replace the default functions with the "fallback" logic, since `parser.js` and `lexer.js` call their tokenizers and renderers with parameters that aren't compatible with the extensions, which are looking for just `token` or `src`. So instead:

- Tokenizers 'fall back' anyway as in https://github.com/markedjs/marked/pull/2043#discussion_r637349004

- For Renderers, I moved the extensions to the top of the `parser.js` loop so they can do a similar thing and have priority over the default renderers. If the extension shares the name of a default renderer but returns false, it will continue on so he default renderer can execute.

Made a unit test showing both cases. The tokenizer returns false in one but has a working renderer, and the renderer returns false with a working tokenizer in the other, and falls back to the default behavior in the false cases.

src/marked.js
```diff
@@ -207,7 +279,12 @@ marked.walkTokens = function(tokens, callback) {
         break;
       }
       default: {
-        if (token.tokens) {
+        if (marked.defaults?.extensions?.childTokens?.[token.type]) { // Walk any extensions
+          marked.defaults?.extensions.childTokens[token.type].forEach(function(childTokens) {
+            marked.walkTokens(token[childTokens], callback);
+          });
+        }
+        if (token.tokens && !marked.defaults?.extensions?.childTokens[token.type]) {
```
UziTech, line 287:
```suggestion
        } else if (token.tokens) {
```

UziTech:
This looks amazing :100:. It took a while but I think this is almost done. I'm going to take about a week to come up with some other tests before I approve this but I think it should all work. :+1:

calculuschild:
I think some of my unit tests might be poorly designed or overly complicated or redundant as well. I'm not sure how important it is to clean those up.

UziTech:
I added a few tests for custom tokenizers and renderers returning false. I think this is good to go once documentation is updated. :+1:

calculuschild:
Is there going to be confusion between `marked.use()` and `marked.setOptions()`? We have both in the docs and at a glance I'm not sure it's obvious to most users what the difference is. Would we ever want to deprecate `setOptions()`?

I just want to make sure what I'm writing up for the docs is accurate.

UziTech:
I don't know if changing `marked('text', options)` to be the same as `marked.use(options)` would be the best option. That would make `marked('text', options)` a lot slower since it would have to build the options on every call.

I think it is ok to have them do different things. We should probably document it better.

calculuschild:
Alright, I added in some documentation and rearranged/rephrased a bit of the other sections to match. Feel free to add anything else that I missed or have wrong.

calculuschild:
@UziTech What do you think about passing in like a `previousToken` and `nextToken` to the `walktokens`? Since `walkTokens` is kind of a "post-processor" on the token tree as a whole, it could be handy to be able to manipulate tokens depending on what the neighboring tokens are.

docs/USING_PRO.md
```diff
@@ -1,24 +1,71 @@
...
+  smartLists: true,
+  smartypants: false,
+  xhtml: false
+});
+```
+
+You can also supply an array of multiple `extension` objects.
+
+```
+marked.use([myExtension, extension2, extension3]);
```
UziTech, line 27: Do you think we should allow multiple parameters (i.e. `marked.use(extension1, extension2, etc)`) instead of an array? might be a little cleaner for the user and not any harder for marked to implement.

docs/USING_PRO.md
```diff
@@ -56,20 +103,31 @@ console.log(marked('# heading+'));
...
-- list(*string* body, *boolean* ordered, *number* start)
-- listitem(*string* text, *boolean* task, *boolean* checked)
-- checkbox(*boolean* checked)
-- paragraph(*string* text)
-- table(*string* header, *string* body)
-- tablerow(*string* content)
-- tablecell(*string* content, *object* flags)
+### Block-level renderer methods
+
+- <code>**code**</strong>(*string* code, *string* infostring, *boolean* escaped)</code>
```
UziTech, line 108:
```suggestion
- <code>**code**(*string* code, *string* infostring, *boolean* escaped)</code>
```

UziTech (changes requested):

UziTech:
> What do you think about passing in like a `previousToken` and `nextToken` to the `walktokens`?

I have thought about adding all of those things. The problem is: where do you stop? it could also be helpful to know what the parent token is ... and the grandparent token ... and all of the siblings.

I think the best thing to do is say `walkTokens` is simply a convenience function for walking each token and altering the tokens just based on that token. If someone wants to do more they can get the tokens and walk them however they want manually.

docs/USING_PRO.md
calculuschild: Sure, if that seems like a better way to do it.

calculuschild:
> they can get the tokens and walk them however they want manually.

How would a user go about doing that? Just accessing the Lexer?

> The problem is: where do you stop?

Right... Ok, what if we do this instead: after a quick google, apparently we could get the index of the current token being walked with `for [index, val] of tokens.entries()`. Then we could pass in `index` and `tokens` which would give access to all siblings, and you know your position within the siblings. Children are already accessible. And parents/grandparents could be accessed by just walking *them* and checking if they contain the child token of interest, then manipulating that child. I think that gives access to anything in the token tree. You would just have to write your walktokens function to look *sideways* or *down*.

UziTech:
> How would a user go about doing that? Just accessing the Lexer?

yes

> parents/grandparents could be accessed by just walking them and checking if they contain the child token of interest

How is that easier than just accessing tokens from the lexer and walking them manually?

If you want extensions to be able to get tokens we should do something like [hooks](#1878) and just have a hook that gives the extension all tokens between lex and parse and let them do whatever they want. Same with a hook for pre and post processing.

For the [gfm-heading-id](https://github.com/markedjs/marked-gfm-heading-id#reset) extension it would be nice to be be able to reset the slugger in a preprocess hook instead of needing to call it manually.

calculuschild:
> How is that easier than just accessing tokens from the lexer and walking them manually?

Because then the extension creator essentially has to re-write his own custom `walkTokens` logic and break apart the Marked pipeline to manually call the lexer, then run his custom walkTokens, then call the parser, instead of just using `marked('my Markdown')`. This is a headache if anyone else actually wants to use that extension, instead of just plugging into `marked.use()` and it "just works".

The preprocessor/postprocessor hooks would make it work with `marked.use()` so you don't have to break apart the pipeline, but it still requires users to build their own `walkTokens` from scratch which seems redundant given that we already have `walkTokens`.

Ok, so another approach: What if inside of `walkTokens` we inject two properties into the token before calling the extension functions. A `previousToken` and a `parentToken`. Then each token can navigate up (`token.parentToken.parentToken.parentToken`), down (`token.tokens.tokens`), or backward to previous siblings (`token.previousToken.previousToken`). You could even get away with *just* the `previousToken`. And you just need to handle it like CSS where your selectors can only access the next element or child elements in the list, but in reverse (you can only access previous or child items, since next siblings wouldn't have been given their `previousToken` yet.)

So if you want to change a token between two paragraphs:

```
tokens = [
token1 {type: 'paragraph' ...},
token2 {type: 'table', previousToken: token1 ...}
token3 {type: 'paragraph', previousToken: token2 ...}
];

//WalkTokens
walkTokens(token) {
  if(token?.previousToken?.previousToken.type === 'paragraph' && token.type === 'paragraph') {
    token.previousToken.text += `I'm between two paragraphs!`;
  }
}
```

Essentially the only limitation with `walkTokens` as it is now, is accessing siblings. You can already do everything you need to do with parent/grandparent/child by just starting your function at the parent or grandparent and looking down the children.

UziTech:
Except down isn't always in `tokens` which means they would have to re-create the logic of walkTokens anyway. Again I don't see this being easier than walking manually. Also I think this PR has done enough for now. We should get this merged and improve it from there.

calculuschild:
> Also I think this PR has done enough for now.

Fair enough. I'll work on that last change.

docs/USING_PRO.md
calculuschild: Ok this is changed.

UziTech (approved):
nice work. :+1: This is going to allow so many new things for marked.

calculuschild:
@styfle , @joshbruce , @davisjam 

Have you guys had a chance to look at this PR? I know its a hefty one so we could sure use some extra eyes on it!

joshbruce (approved):

joshbruce:
I'd be lying if I said I understood it all. I appreciate that performance isn't hit too hard given this is something the community has been wanting directly (or by proxy) for a while now. It seems like at least a couple of members from the community appreciate it as well. So, I'm good.

One thing I'd like to put a pin in is the use of `null` - just saw it in a new line for this PR - probably something for later and not sure how we could remove the reference...kind of a personal drive I'm starting to get with my coding style.

src/marked.js
```diff
@@ -141,46 +141,114 @@ marked.defaults = defaults;
...
+    if (pack.extensions) {
+      hasExtensions = true;
+      pack.extensions.forEach((ext) => {
+        if (!ext.name) {
+          throw new Error('extension name required');
         }
-        return ret;
-      };
+        if (ext.renderer) { // Renderer extensions
+          const prevRenderer = extensions.renderers?.[ext.name];
```
priyeshs11, line 158: @UziTech @calculuschild I think this line is causing the error below. Will create a seperate issue too.

```
node_modules/protobufjs/node_modules/marked/src/marked.js:158
const prevRenderer = extensions.renderers?.[ext.name];
                                                    ^
SyntaxError: Unexpected token '.'
```

priyeshs11: Issue here: https://github.com/markedjs/marked/issues/2107

PR #2099, opened 2021-06-13
Token Relational Properties

calculuschild:
Just a base attempt at adding some sibling properties to tokens, based on discussion #2097 . Builds in a `previousSibling` and `nextSibling` parameter to each token so you can more easily traverse the token tree (i.e., when building an Extension or using Walktokens). Next steps might include `parent` and `children` parameters as well.

Does break a lot of unit tests where a token is supposed to have no parameters.

A couple benchmark runs show that this might be a bit slower but it's hard to tell by how much. Just want rough feedback if this is even a good way to do it.

## Contributor

## Committer

### DIFF
--- a/lib/marked.esm.js
+++ b/lib/marked.esm.js
@@ -9,9 +9,13 @@
  * The code in this file is generated from files in ./src/
  */
 
-var defaults$5 = {exports: {}};
+function createCommonjsModule(fn) {
+  var module = { exports: {} };
+	return fn(module, module.exports), module.exports;
+}
 
-function getDefaults$1() {
+var defaults = createCommonjsModule(function (module) {
+function getDefaults() {
   return {
     baseUrl: null,
     breaks: false,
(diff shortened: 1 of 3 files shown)


UziTech:
This still seems to me like it can be solved better solved by providing all of the tokens in a hook. I feel like the amount of work needed to make this work would not be worth it since it is such a rare thing in markdown to make rules based on sibling or parent tokens.

calculuschild:
>  it can be solved better by providing all of the tokens in a hook. 

That would be solving a different problem though. It is related, but not directly the purpose of this PR.

1) **Providing tokens in a hook** *SOLVES:* Users do not have a way to access the token tree without splitting apart the render steps.

2) **Providing sibling tokens** *SOLVES:* Users do not have a straightforward way to access sibling tokens.

Hooks would allow a user to code up his own solution to 2), but it does not provide the solution to 2) directly.

> it is such a rare thing in markdown to make rules based on sibling or parent tokens.

But it does happen, and coding around those cases even in Marked.js is obnoxious. If you have a code block, you have to check the previous sibling to make sure it's not interrupting a paragraph. If you have a text token, you need to check if the previous sibling is also text so you can merge them. In a List or Block token, you need to know if the parent was at the top level or not to decide how to parse the internal tokens.

> the amount of work needed to make this work would not be worth it

But the work is already in the PR? At least for sibling tokens, and I would be happy to stop there without adding `children` and `parents` because we already have `tokens` properties that are user-accessible.

calculuschild:
But I agree that at a minimum we should expose tokens via hooks so users have at least *some* way to modify the token tree.

PR #2109, opened 2021-06-16
fix: fix node v12

### LINKED ISSUE #2106: Breaking change in 2.1.0
**Marked version: 2.1.0**

**Describe the bug**
The latest code uses the optional chaining operator which makes marked 2.1.0 incompatible with Node V12. It's breaking semantic release which uses Marked. Here:
https://github.com/markedjs/marked/commit/5be9d6d70ea35be6398ee35958fb8ac955a89fbe

Thanks!

UziTech:
**Marked version:** 2.1.0

## Description

remove optional chaining to fix node v12 support

update engines to support node v12

test on node v12

- Fixes #2106 

## Contributor

## Committer

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,6 @@
     "preversion": "npm run build && (git diff --quiet || git commit -am build)"
   },
   "engines": {
-    "node": ">= 8.16.2"
+    "node": ">= 12"
   }
 }
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     strategy:
       matrix:
-        node_version: ['lts/*', 'node']
+        node_version: [12, 'lts/*', 'node']
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -126,7 +126,8 @@ module.exports = class Lexer {
     let token, i, l, lastToken, cutSrc, lastParagraphClipped;
 
     while (src) {
-      if (this.options?.extensions?.block
+      if (this.options.extensions
+        && this.options.extensions.block
         && this.options.extensions.block.some((extTokenizer) => {
           if (token = extTokenizer.call(this, src, tokens)) {
             src = src.substring(token.raw.length);
@@ -244,7 +245,7 @@ module.exports = class Lexer {
       // top-level paragraph
       // prevent paragraph consuming extensions by clipping 'src' to extension start
       cutSrc = src;
-      if (this.options.extensions?.startBlock) {
+      if (this.options.extensions && this.options.extensions.startBlock) {
         let startIndex = Infinity;
         const tempSrc = src.slice(1);
         let tempStart;
@@ -400,7 +401,8 @@ module.exports = class Lexer {
       keepPrevChar = false;
 
       // extensions
-      if (this.options?.extensions?.inline
+      if (this.options.extensions
+        && this.options.extensions.inline
         && this.options.extensions.inline.some((extTokenizer) => {
           if (token = extTokenizer.call(this, src, tokens)) {
             src = src.substring(token.raw.length);
@@ -507,7 +509,7 @@ module.exports = class Lexer {
       // text
       // prevent inlineText consuming extensions by clipping 'src' to extension start
       cutSrc = src;
-      if (this.options.extensions?.startInline) {
+      if (this.options.extensions && this.options.extensions.startInline) {
         let startIndex = Infinity;
         const tempSrc = src.slice(1);
         let tempStart;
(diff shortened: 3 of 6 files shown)


calculuschild:
Aww no... my nice clean optional chaining...

This looks fine. But any chance we can solve the problem by making use of Babel and emitting the es5 version instead of muddying up our source code?

zzhjerry:
> Aww no... my nice clean optional chaining...
> 
> This looks fine. But any chance we can solve the problem by making use of Babel and emitting the es5 version instead of muddying up our source code? That's what Babel is for, right?

Refactoring with babel could come as a separate PR but this fix should be deployed asap as lots of project deployments are broken because of it.

calculuschild:
> Refactoring with babel could come as a separate PR 

We already generate a transpiled version of Marked with Babel that works with older versions of node. All we would need to do is change one line of code to emit that version by default.

zzhjerry:
> > Refactoring with babel could come as a separate PR
> 
> We already generate a transpiled version of Marked with Babel that works with older versions of node. All we would need to do is change one line of code to emit that version by default.

That's fine but v2.0.10 has become a broken change compare to v2.0.7. Could we at least make sure test passes before the package release?

calculuschild (approved):

ZarcoNontol (approved):
👍

package.json
```diff
@@ -83,6 +83,6 @@
     "preversion": "npm run build && (git diff --quiet || git commit -am build)"
   },
   "engines": {
-    "node": ">= 8.16.2"
+    "node": ">= 12"
```
alasdairhurst, line 86: It's one thing to ensure support in CI but it's another thing to change support in a minor version. I understand that only LTS/latest node versions are supported but real functional changes should only happen during a major version, otherwise it's going to break a lot of people.

alasdairhurst:
https://github.com/jsdoc/jsdoc/issues/1926

Note - Marked 2.x supported node 8 from the start. jsdoc 3.x depends on marked@^2.x so will pick up any (breaking) node.js support changes. even if things get working on Node.js 12, it's still broken since jsdoc expects support for node 8.15. Right now there's nothing it can do except roll back and pin marked to a version that did support node 8 or inherit breaking changes whenever marked wants to drop a node version.

package.json
alasdairhurst: Sorry, left this a while back but forgot to submit...

UziTech: If jsdoc wants to support node v8 they could import the es5 version of marked

```js
const marked = require('marked/lib/marked.js');
```

UziTech: For the record according to jsdoc [engines](https://github.com/jsdoc/jsdoc/blob/ac60926981d547688903eba505abd95422260602/package.json#L19-L21) field it only supports `>=v14.17.1`

Fleker: This change is breaking our CI build now. We use this through [Typedoc](https://github.com/TypeStrong/typedoc/blob/master/package.json), with that and our lib using Node `>= 10.18.0`. It would be great if you could revert this change. This is a breaking change as per semantic versioning. I'd recommend you do this in a v3.x release and have a deprecation period to allow dependents to update their packages or find some other mitigation.

```
error marked@2.1.1: The engine "node" is incompatible with this module. Expected version ">= 12". Got "10.18.0"
error Found incompatible module.
```

calculuschild:
@UziTech Looks like there are still breaking issues with this (might need to revert). I think we could solve a lot of problems (and open up the use of es6 in the source code) if we instead just default to the es5 version.

styfle:
@calculuschild The issues is not ES6 (aka ES2015). The issue is the use of optional chaining, which was introduced in ES2020.

See the features and versions here: https://github.com/tc39/proposals/blob/master/finished-proposals.md

This PR solves the root cause: we need to be testing in Node 12 if we intend to support it (thanks @UziTech  for also updated `engines` 🙂)

In the future, we need to drop support (or rather introduce new ES features) under a major semver.

package.json
styfle: @Fleker Node.js 10 reached end of life in April and is no longer supported https://nodejs.org/en/about/releases/

You can probably disable that error if you really need to with `yarn install --ignore-engines`

calculuschild:
> The issue is the use of optional chaining, which was introduced in ES2020.

Correct, I meant to say ES2020. But the solution remains: if we supply the es5 version as the default then we can still use optional chaining, whatever ES2020/ ES2015 feature we want in the source, since Babel nicely converts it all to a version that is compatible with older Node versions. That is the entire purpose of the existence of Babel.

UziTech:
@calculuschild we should also change the ci to testing the built version. I don't know how that will affect PRs or stack traces. Maybe we need to add sourcemaps as well? I feel like this small change isn't as small as first thought.

calculuschild:
Good points. I believe Babel does have an option for source maps which we probably should do.

As for getting into the ci I wouldn't know where to start for this project.

Issue #1923, opened 2021-02-01
Sublist behavior change: Sublists with single space are not rendered as sublists


bwittenberg:
**Marked version:**
v1.2.3

**Describe the bug**
Sublists single indent are not rendered as html sublists.

Why is this significant? To borrow from the github flavored markdown spec: (https://github.github.com/gfm/#why-is-a-spec-needed-)

> How much indentation is needed for a sublist? The spec says that continuation paragraphs need to be indented four spaces, but is not fully explicit about sublists. It is natural to think that they, too, must be indented four spaces, but Markdown.pl does not require that. This is hardly a “corner case,” and divergences between implementations on this issue often lead to surprises for users in real documents.

I upgraded from v0.8.2 to v1.2.5, and some customers relied on sublist indentation of a single space.

It looks like the behavior change shipped in v1.2.3.

Here is a comparison of changes between versions:
https://github.com/markedjs/marked/compare/v1.2.2...v1.2.3#

**To Reproduce**

[v1.2.3 Demo](https://marked.js.org/demo/?text=-%20one%0A%20-%20two&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=1.2.3)

[...]

PR #2112, opened 2021-06-18
fix: Full Commonmark compliance for Lists

calculuschild:
**Marked version:** 2.1.1

## Description

A reworking of the Block Lists tokenizer. Also adjusting some of the `New` unit tests because they are not accurate to the actual Commonmark spec and dingus results.

- Fixes Commonmark Examples 232, 234, 243, 244, 248, 250, 254, 276, 277, 287, 288, 289. This now passes all **Lists** and all **List Items** Commonmark tests!

- Fixes some gitihub issues probably. Need to dig through and see.

This seems to be about the same speed as Master or slightly slower, but it's always hard to tell.

Side note, should we move the `New` tests that feature the `pedantic` option into the `Original` folder to be with all the other pedantic tests?

## Contributor

## Committer


UziTech:
I believe `original` are actual specs for the original markdown.pl and `new` are basically any tests that are not actually part of any spec but should (for the most part) continue passing.

calculuschild:
Ok that makes sense. I could see that the `New` specs for Pedantic do seem to match the Daring Fireball dingus so that's fine. I could use some help figuring out the rule for `New: list_align_pedantic` though. There doesn't seem to be any logical pattern here:

```
- one
 - two
  - three
    - four
     - five
      - six
       - seven
```

becomes

```
<ul>
	<li>one
		<ul>
			<li>two</li>
			<li>three</li>
			<li>four
				<ul>
					<li>five</li>
					<li>six</li>
					<li>seven</li>
				</ul>
			</li>
		</ul>
	</li>
</ul>
```

Same with `New : list_item_text` :

```
  * item1

    * item2

  text
```

becomes

```
<ul><li><p>item1</p>  <ul><li>item2 </li></ul> <p>text</p> </li></ul>
```

calculuschild:
For `New : main`, the output HTML does not match the commonmark dingus. How do we want to change this test? Change the Markdown so it matches the current HTML, or change the HTML so it matches the current Markdown?

Or... after making either of those changes, this test might be redundant since I think it overlaps with existing tests everywhere...

UziTech:
`New: list_align_pedantic` fixes #1923

`New: list_item_text` fixes #1947

`New: main` should probably be marked as `pedantic`?

calculuschild:
Ok, thanks for tracking that down. I still am having trouble with the alignment rules for those two test cases though. Can you help me understand when a pedantic list considers something a sublist versus not? The daring fireball spec does not say anything about when to nest sublists that I can see.

UziTech:
> the dingus gives REALLY strange results with stuff 

The original spec was very lax on details because it assumed well formatted markdown, so bad markdown should give unspecified results (garbage in, garbage out). Unfortunately people who write bad markdown still want consistency.

The easiest way to get around it is to have `if (options.pedantic)` blocks that correct for the daringfireball dingus results.

UziTech:
> Every bullet with an indent between 1 and 4 spaces is nested at the same "level", and then 5-8 spaces is the next level

That seems to be the case.

> But maybe it's actually anywhere between 1-4 spaces and the indentation first bullet doesn't matter?

yes

calculuschild:
Right right. I'm all for that.

My question is more an ask for help in reverse engineering the dingus logic so I can get the same garbage in garbage out for consistency, as you say. Then I can work out where and what to put into the `options.pedsntic` sections.

calculuschild:
> New: main should probably be marked as pedantic?

@UziTech Unfortunately it doesn't follow Pedantic rules either. The problem area is this:

```
* List Item 2
  * New List Item 1
    Hi, this is a list item.
  * New List Item 2
    Another item
        Code goes here.
        Lots of it...      
  * New List Item 3
    The last item
```

The test wants it to become

```
<ul>
<li>New List Item 1 Hi, this is a list item.</li>
<li>New List Item 2 Another item 
<pre><code>Code goes here.
Lots of it...</code></pre>
</li>
<li>New List Item 3 The last item</li>
</ul>
```

However, an indented code block cannot interrupt a paragraph without a blank line before, both in CommonMark and in Pedantic.

So both Commonmark and Pedantic dingus give this:

```
<ul>
<li>New List Item 1 Hi, this is a list item.</li>
<li>New List Item 2 Another item Code goes here. Lots of it...</li>
<li>New List Item 3 The last item</li>
</ul>
```

UziTech:
Sounds good it looks like it was created well before my time so most likely it is out of date.

calculuschild:
Tadaa! Passing all spec tests now, plus 3 more Commonmark examples.

Unfortunately there are a bunch of Unit tests that are expecting tokens to look different now so that needs to be cleaned up.

calculuschild:
And now unit tests are all passing.

calculuschild:
Yeayah! Got all commonmark examples working!

calculuschild:
Bench speeds (really inconsistent):
**Master:*
```
es5 marked completed in 5114ms and passed 86.90%
es6 marked completed in 5281ms and passed 86.90%
es5 marked (gfm) completed in 5582ms and passed 86.29%
es6 marked (gfm) completed in 5705ms and passed 86.29%
es5 marked (pedantic) completed in 5456ms and passed 71.03%
es6 marked (pedantic) completed in 5342ms and passed 71.03%
commonmark completed in 4113ms and passed 100.00%
markdown-it completed in 4242ms and passed 89.21%
```

**This PR:**
```
es5 marked completed in 5322ms and passed 88.75%
es6 marked completed in 5418ms and passed 88.75%
es5 marked (gfm) completed in 5653ms and passed 88.14%
es6 marked (gfm) completed in 5771ms and passed 88.14%
es5 marked (pedantic) completed in 5839ms and passed 71.34%
es6 marked (pedantic) completed in 6011ms and passed 71.34%
commonmark completed in 4259ms and passed 100.00%
markdown-it completed in 4349ms and passed 89.21%
```

src/Tokenizer.js
```diff
@@ -194,134 +194,136 @@ module.exports = class Tokenizer {
...
-            l = i + 1;
-          }
+      bull = isordered ? `\\d{1,9}\\${bull.slice(-1)}` : `\\${bull}`;
+
+      if (this.options.pedantic) {
+        bull = isordered ? bull : '[*+-]';
+      }
+
+      // Get next list item
+      const itemRegex = new RegExp(`^( {0,3}${bull})((?: [^\\n]*|\\h*)(?:\\n[^\\n]*)*(?:\\n|$))`);
```
UziTech, line 221: is this supposed to be `\n`?
```suggestion
      const itemRegex = new RegExp(`^( {0,3}${bull})((?: [^\\n]*|\\n*)(?:\\n[^\\n]*)*(?:\\n|$))`);
```

calculuschild: It's all horizontal whitespaces. Could probably be ` ` but I wasn't sure if tabs or `nbsp` shows up in here.

src/Lexer.js
```diff
@@ -204,6 +204,10 @@ module.exports = class Lexer {
         l = token.items.length;
         for (i = 0; i < l; i++) {
           token.items[i].tokens = this.blockTokens(token.items[i].text, [], false);
+          if (token.items[i].tokens.some(t => t.type === 'space')) {
+            token.loose = true;
+            token.items[i].loose = true;
+          }
```
UziTech, line 210: This should be part of the list tokenizer. If someone wants to extend the list tokenizer to prevent loose lists they should only need to extend the list tokenizer.

src/rules.js
```diff
@@ -43,10 +43,10 @@ block.def = edit(block.def)
...
 
 block.bullet = /(?:[*+-]|\d{1,9}[.)])/;
-block.item = /^( *)(bull) ?[^\n]*(?:\n(?! *bull ?)[^\n]*)*/;
-block.item = edit(block.item, 'gm')
-  .replace(/bull/g, block.bullet)
-  .getRegex();
+// block.item = /^( {0,3})(bull)( [^\n]*(?:\n(?! {0,3}bull )[^\n]*)*\n?)/;
+// block.item = edit(block.item, 'gm')
+//   .replace(/bull/g, block.bullet)
+//   .getRegex();
```
UziTech, line 49: this should actually be deleted if it is not used

UziTech (changes requested):

src/Tokenizer.js
UziTech: In javascript `\h` just matches the `h` character. I don't think javascript has a horizontal whitespace character class.

UziTech: Tested in the devtools console.

image

calculuschild: Well I'll be. I guess we can change it to a normal space then.

calculuschild: Alright, I changed all instances of `\h` to ` `.

src/rules.js
calculuschild: Done.

calculuschild: Done.

src/Tokenizer.js
UziTech: Maybe `[ \t]` would be better?

calculuschild: I seem to be remembering... Doesn't marked already convert tabs to spaces before lexing or something?

src/Lexer.js
calculuschild: This relies on the previous line as well because we need to Lex the child list items first. Can I just put the whole section into the Tokenizer? (i.e., is it ok if a Tokenizer handles the lexing of its own children?)

https://github.com/markedjs/marked/blob/c934790c3fb2e45d4cbabf96c0536ce2e84a1efd/src/Lexer.js#L206-L210

src/Tokenizer.js
calculuschild: Ah yeah, here. Is it ok to just leave it as a space then?

https://github.com/markedjs/marked/blob/c934790c3fb2e45d4cbabf96c0536ce2e84a1efd/src/Lexer.js#L109-L110

UziTech: Ya I guess. I know there is an issue about replacing tabs in code blocks so sometime in the future we will need to change it but just spaces should work for now.

calculuschild:
Looks like speed is back to as good (or better) than Master with the last few changes:

Bench speeds (really inconsistent):
**Master:**
```
es5 marked completed in 4530ms and passed 86.90%
es6 marked completed in 4626ms and passed 86.90%
es5 marked (gfm) completed in 4997ms and passed 86.29%
es6 marked (gfm) completed in 5071ms and passed 86.29%
es5 marked (pedantic) completed in 4745ms and passed 71.03%
es6 marked (pedantic) completed in 4916ms and passed 71.03%
commonmark completed in 3668ms and passed 100.00%
markdown-it completed in 3526ms and passed 89.21%
```

**This PR:**
```
es5 marked completed in 4517ms and passed 88.75%
es6 marked completed in 4695ms and passed 88.75%
es5 marked (gfm) completed in 4786ms and passed 88.14%
es6 marked (gfm) completed in 4933ms and passed 88.14%
es5 marked (pedantic) completed in 4786ms and passed 71.34%
es6 marked (pedantic) completed in 4934ms and passed 71.34%
commonmark completed in 3590ms and passed 100.00%
markdown-it completed in 3559ms and passed 89.21%
```

**Previous speed** (3a409a166102678e218243295f106ea4ab0962aa)
**This PR:**
```
es5 marked completed in 4704ms and passed 88.75%
es6 marked completed in 4774ms and passed 88.75%
es5 marked (gfm) completed in 4913ms and passed 88.14%
es6 marked (gfm) completed in 5067ms and passed 88.14%
es5 marked (pedantic) completed in 4938ms and passed 71.34%
es6 marked (pedantic) completed in 5212ms and passed 71.34%
commonmark completed in 3530ms and passed 100.00%
markdown-it completed in 3552ms and passed 89.21%
```

src/Lexer.js
calculuschild: Unfortunately, determining whether the list is loose is not as simple as checking for blank lines in the list `raw` either, since code blocks for example can contain blank lines that don't count.

Kinda dumb, but what about assigning the lexer as a property of its tokenizer object so it can access parsing functions?
```
  static lex(src, options) {
    const lexer = new Lexer(options);
    lexer.tokenizer.lexer = lexer;
    return lexer.lex(src);
  }
```

UziTech: This is a tough one. I'm tempted to say that it is fine how it is since someone could create an extension that runs before the list tokenizer that consumes the list to create a non-loose list. But this is a slippery slope if we start moving stuff out of the tokenizers.

I'm thinking moving `this.blockTokens` into the tokenizers would be a better solution since it is available in the extensions.

Either way I think this will be a breaking change.

UziTech: To be honest it would be nice if all of the block tokens that could contain other block tokens would call `blockTokens` inside the tokenizer.

UziTech: > what about assigning the lexer as a property of its tokenizer object so it can access parsing functions?

That seems to be the easiest way to do it without passing all of the lexer functions into each tokenizer.

calculuschild: So you are saying it would be nice if Tokenizers handle their own children tokens instead of making the Lexer do it, right?

I'd be ok making a separate PR for that then, but what would be the preferred way of exposing lexer.blocktokens() to the tokenizers?

UziTech: or the Tokenizer class only stores `options` in the constructor. Maybe we should change it so the `this` in tokenizer functions is the `lexer` like in extension tokenizer functions?

calculuschild: > we should change it so the `this` in tokenizer functions is the `lexer` like in extension tokenizer functions?

In extensions we used `call` to do that, but my impression is `call` is somewhat slower than being able to just call the tokenizer functions directly.

Is there some other way to change `this` in the tokenizer functions?

UziTech: I am referring to `call` but if it is slower than we can do something like `this.tokenizer.lexer = this;`

I feel like it should be consistent with extensions so maybe we should change extensions as well?

calculuschild: Would we want that PR to refactor all the tokenizers with child tokens or should those all be separate PRs? Changing any of them is probably breaking so maybe all at once, no?

UziTech: Ya all at once. Maybe we should keep this PR as is then change it in that PR.

calculuschild: Sounds good.

UziTech (approved):

src/Lexer.js
```diff
@@ -217,13 +221,19 @@ module.exports = class Lexer {
...
-          this.tokens.links[token.tag] = {
-            href: token.href,
-            title: token.title
-          };
+        lastToken = tokens[tokens.length - 1];
+        if (lastToken && (lastToken.type === 'paragraph' || lastToken.type === 'text')) {
+          lastToken.raw += '\n' + token.raw;
+          lastToken.text += '\n' + token.raw;
+        } else {
+          if (!this.tokens.links[token.tag]) {
```
styfle, line 231: Nit pick: this could be changed to `else if` instead of nested `else` and `if`

styfle (approved):
Great work, thanks!

calculuschild:
Reminder to not merge this quite yet. Now that #2124 is approved I want to look over this again because it needs to be converted over to Lex its own child tokens in that new format.

UziTech:
I merged #2124 so you should be able to rebase this and make the changes for lexing the child tokens in the tokenizer.

calculuschild:
@UziTech .....sigh.... It is done.... :finnadie: 

I don't know why but that was a nightmare.

src/Tokenizer.js
```diff
@@ -164,145 +164,149 @@ module.exports = class Tokenizer {
...
+      // Do not consume newlines at end of final item. Alternatively, make itemRegex *start* with any newlines to simplify/speed up endsWithBlankLine logic
+      list.items[list.items.length - 1].raw = raw.trimRight();
+      list.items[list.items.length - 1].text = itemContents.trimRight();
+      list.raw = list.raw.trimRight();
+
+      const l = list.items.length;
+
+      // Item child tokens handled here at end because we needed to have the final item to trim it first
+      for (i = 0; i < l; i++) {
+        this.lexer.state.top = false;
```
UziTech, line 303: do we need to reset `top` after this for loop?

UziTech: I see it is reset at the end of the `blockTokens` function in the lexer.

UziTech (approved):

PR #2119, opened 2021-06-21
fix: add Node.js 10 to CI and loosen `engines` field

### LINKED ISSUE #2111: Breaking Change in 2.1.1
**Marked version:** 2.1.1

**Describe the bug**
Version 2.1.1 [drops support both for Node 8 and Node 10](https://github.com/markedjs/marked/pull/2109/files), mandating Node 12 and higher. This is a sudden change that [violates semantic versioning](https://semver.org/) by cutting off support for previously working versions. Doing this change to the `engines` field in a patch makes it very hard to catch ahead of time, but does result in breaking our CI pipeline via [Typedoc](https://github.com/TypeStrong/typedoc/blob/master/package.json). Our library and Typedoc both have a minimum of Node 10.

```
+ yarn
# ...
error marked@2.1.1: The engine "node" is incompatible with this module. Expected version ">= 12". Got "10.18.0"
error Found incompatible module.
```

**To Reproduce**

* Start with an environment using Node 10 (`nvm use 10.18.0`)
* Create a project with a dependency of `typedoc@^0.20.0` or `marked@^2.00`
* Run `yarn`

**Expected behavior**

I expect that a patch release will maintain existing support for Node platforms and that any deprecation is both clearly documented and breaking changes are placed in a major release.

Can you please revert this change to `engines` and release a 2.1.2 as soon as possible? I understand the value in dropping older versions, but this change was sudden and are having effects on dependents.

styfle:
Fixes #2111

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,6 +10,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     strategy:
       matrix:
+        # lowest verison here should also be in `engines` field
         node_version: [10, 12, 'lts/*', 'node']
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     strategy:
       matrix:
-        node_version: [12, 'lts/*', 'node']
+        node_version: [10, 12, 'lts/*', 'node']
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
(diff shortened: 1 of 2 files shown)


UziTech (approved):

UziTech:
This works but I don't think it solves the problem. V2.0.7 `engines` field was v8 so this is still dropping node versions without a major version change and it is supporting a node version after work which is kind of worst of both worlds.

styfle:
@UziTech I just [tried Node 8](https://github.com/markedjs/marked/pull/2119/checks?check_run_id=2886013344) and it failed so I think we can still say this is a patch version to fix the `engines` field

Fleker:
Thanks for doing this PR.

calculuschild (approved):

calculuschild:
Gosh dang. I hate when the close PR button pops up under my thumb on mobile.

PR #2124, opened 2021-06-30
Tokenizers lex their own child tokens

calculuschild:
Based on the conversation in https://github.com/markedjs/marked/pull/2112#discussion_r656741740, this is the start of an attempt to have the Tokenizers handle the lexing of their own children tokens, rather than making the Lexer.js do it.

For block tokens this was relatively simple. For inline tokens it's also not a huge issue, except for the ugliness that comes with passing in `inRawBlock` and `inLink` to a bunch of the Tokenizers since it kind of muddies up the legibility in what the Tokenizers are actually doing. Passing those values around seems like a [code smell](https://softwareengineering.stackexchange.com/questions/335005/is-there-a-name-for-the-anti-pattern-of-passing-parameters-that-will-only-be) we could avoid but I don't know how, or how much those variables actually need to be passed around. Any thoughts? I wanted to get some early feedback before going through the whole thing.

**Edit :** What about refactoring the `inLink` and `inRawBlock` flags to instead be properties of the Lexer? I.e. in the constructor: 
```
lexer.state = { //or lexer.flags, etc...
  inLink : false,
  inRawBlock : false
}
```

And a second question: do we want the logic of `inline()` from `Lexer.js` to also be handled by the Tokenizers themselves?

## Contributor

## Committer


UziTech:
> What about refactoring the inLink and inRawBlock flags to instead be properties of the Lexer?

I like that idea

> do we want the logic of inline() from Lexer.js to also be handled by the Tokenizers themselves?

Ya, The way I see it is the Lexer handles the overall state (including which order to call the tokenizers) and the tokenizers handle everything about creating the token (including children).

Currently the tokenizers handle everything about creating the token except for the children but as we see in https://github.com/markedjs/marked/pull/2112#discussion_r656741740 sometimes the tokens need to know the children to create the token.

calculuschild:
> What about refactoring the inLink and inRawBlock flags to instead be properties of the Lexer?

Alright. I've added this now. Just need to finish moving `inline()` over.

calculuschild:
@UziTech New issue I'm running into. In the case of adjacent tokens that we merge together (such as `paragraph` and `text`), generating child tokens before the merge leads to errors. If you try to do:

```
        if (lastToken && lastToken.type === 'text') {
          lastToken.raw += '\n' + token.raw;
          lastToken.text += '\n' + token.text;
          lastToken.tokens = lastToken.tokens.concat(token.tokens); //<== Trying to merge tokens
        }
```

You end up with trailing/starting spaces between the two `text` tokens being trimmed:

```
Expected: <ol><li><p>A paragraph with two lines.</p><
  Actual: <ol><li><p>A paragraphwith two lines.</p><p
```

This can also lead to some child tokens not being detected because the ending half is in the other token before merging.

Any ideas?

UziTech:
Would it be feasible to only do `paragraph` and `text` in the lexer after all block tokens? Do you feel that would be too inconsistent?

calculuschild:
I might be able to manage the other ones. I'm pretty sure I can get Tables at least.

It would be kind of inconsistent but it might be something we just need to leave until a deeper revision down the road.

calculuschild:
@UziTech Sigh... new problem.

Reflinks that are defined after they are referenced don't get handled since they aren't registered to the `this.tokens.links` yet when we try to parse child inlineTokens.

```
# [Foo]  //<-- parsing child tokens of this Heading before `[foo]` is defined.
[foo]: /url
```

I'm stumped on this one. How do we approach this?

UziTech:
Hmm. Maybe we need to have a property on the token that is a function that the lexer will call to parse the inline tokens after all block tokens are handled? Not sure what that would do with the benchmarks but It might slow it down too much.

Right now the extensions parse their inline tokens right away but we might want them to wait for all block tokens as well in case they need reflinks as well.

UziTech:
or instead of tokenizers calling `token.tokens = this.inlineTokens(token.text)` we could have a function queue so they could call `this.inline(token.text, token.tokens)`.

then after the block tokens are complete we would run
```js
while (const next = inlineQueue.shift()) {
  this.inlineTokens(next.src, next.tokens);
}
```

calculuschild:
We could do something like that. That might also better work with paragraphs and text that can't inline until they are merged.

The only problem I see with that, though, is the original problem of https://github.com/markedjs/marked/pull/2112#discussion_r656741740 is broken again.

UziTech:
I guess it would have to be a function to accomplish setting token properties based on children. Or should that be what walkTokens is for?

calculuschild:
Maybe we could use walktokens. I'm a little hesitant to break token function apart further but that might make sense.

Hm... What about handling it in the Renderer? The token can organize all the text info but the Renderer decides how to render it based on what the children look like?  Meh...

UziTech:
> The only problem I see with that, though, is the original problem of #2112 (comment) is broken again, since the parent token won't be able to see its child tokens until its too late.

The `space` token is a block token so we should be able to do `blockTokens` inside the tokenizer and use an `inlineQueue` for inline tokens.

calculuschild:
Ahhhhhhhh. Ok that just might work.

calculuschild:
Ok, so before I push anything up in case we need another way around this:

I set up the inlineQueue and it works fine. Reflinks seem to work fine.

But I still run into the final issue of `text` and `paragraph` tokens that haven't finished merging. If I use `this.inline(token.text, token.tokens)` inside the `text` tokenizer it only applies to the first `token.text` value before any merging happened.

calculuschild:
Alright... I kludged it with the same "check `lastToken` and update if needed" strategy. Let me know if this is acceptable.

src/Tokenizer.js
```diff
@@ -180,14 +149,15 @@ module.exports = class Tokenizer {
     }
   }
 
-  blockquote(src) {
+  blockquote(src, top) {
```
UziTech, line 152: should `top` be in the lexer state instead of passing it?

src/Tokenizer.js
```diff
@@ -521,7 +532,7 @@ module.exports = class Tokenizer {
           href = href.slice(1, -1);
         }
       }
-      return outputLink(cap, {
+      return outputLink.call(this, cap, {
```
UziTech, line 535: We could pass `this.lexer` as a parameter if it would be faster than `.call`

UziTech:
I like it! It makes the lexer a lot cleaner and the `inlineQueue` probably makes it faster since it doesn't have to traverse every token looking for ones that have inline tokens.

src/Tokenizer.js
calculuschild: Yeah probably.

calculuschild:
Do you have any ideas of on the unit tests? Particularly the 'inline' ones? Now that the original `inline()` function is totally removed, I'm not sure if we just need to delete those unit tests or if theres a good way to update them so they still cover the use cases they were originally made for.

calculuschild:
Also, just noticing that the new extension type has a different `this`, so they access the `inline` function via `this.inline()`, but the older style of extension has to use `this.lexer.inline()`

Should we make those consistent?

UziTech:
I'll have to look at the tests next week. I won't have good service until then.

We could do something like `extension.call({lexer: this})`

UziTech:
I just removed the inline tests. We do have tests that call `lexer.inline` in extensions so we should be good on those tests.

I also changed the extensions `this` to `{lexer: this}` for the extension tokenizers and `{parser: this}` for the extension renderers to be more consistent. We should also update the documentation for that.

src/Tokenizer.js
calculuschild: Just did this.

src/Tokenizer.js
calculuschild: I added this as well. Right now I just set `top` back to `true` at the end of `blockTokens` but this assumes you can only go one level deep, and this passes all the tests. Is there ever a case where you can get more than one level deep? Should `top` instead be `depth` and we add and subtract from it as we go down and up instead of setting true or false?

calculuschild:
I updated the documentation. Double-check it for typos if you like. I think everything else up to this point is resolved now?

UziTech (approved):
Nice work! :100:

UziTech:
Do we want to close #2126 since this PR also removes the nptable tokenizer?

calculuschild:
Yep, did it just now. 👍

calculuschild:
#2112 might need tweaking now, but it should go out in the same major version bump if this looks ready to go.

calculuschild:
@davisjam @joshbruce @styfle Is anyone able to take a look at this one? I'm eager to get this one merged!

styfle (approved):

calculuschild:
Hm, do we need to add to the documentation mention of the `lexer.state` properties? In case people want to use them in their extensions?

UziTech:
There are the following breaking changes that I can think of in this PR:

- Tokenizers will create their own tokens with `this.lexer.inline(text, tokens)`. The `inline` function will queue the token creation until after all block tokens are rendered.
- `nptable` tokenizer is removed and merged with `table` tokenizer.
- Extensions tokenizer `this` object will include the `lexer` as a property. `this.inlineTokens` becomes `this.lexer.inline`.
- Extensions parser `this` object will include the `parser` as a property. `this.parseInline` becomes `this.parser.parseInline`.
- `tag` and `inlineText` tokenizer function signatures have changed.

Am I missing any?

UziTech:
> do we need to add to the documentation mention of the lexer.state properties?

That wouldn't hurt but it could be done in a separate PR.

calculuschild:
> Am I missing any?

Just that some function signatures have changed. If people were overwriting any of those functions they might not work anymore.

calculuschild:
One minor thing is the naming of `inline()` and `inlineTokens()`. I have to keep reminding myself what the difference is. Is there another way we could name things? Maybe something like `inline` => `inlineTokens` and `inlineTokens` => `nestedInlineTokens` ?

Might make it easier for users to understand when to use each one.

UziTech:
> Might make it easier for users to understand when to use each one.

I don't think `inlineTokens` is something extension creators should call. Ideally `inlineTokens` should be a private method and `inline` should be a public method that should always be used to queue the creation of inline tokens.

PR #2126, opened 2021-07-02
Remove nptable tokenizer

calculuschild:
`nptable` tokenizer is 95% identical to `table`, except for handling a weird edge case.  This PR moves that special case to be handled in splitCells() to reduce code redundancy and improve speed (one less tokenizer to cycle through).

Also corrects one `new` test that did not follow the correct behavior compared to actual Github behavior: If a row of cells is missing the leading\trailing pipe, then the leading\trailing cell cannot be empty.

(this is also just some cleanup in prep for #2124, since moving things into the Tokenizer was resulting in large duplicated blocks in each both `table` and `nptable`)

```
 A | B
-- | --
   | 2
```

Results in:
   
 A | B
-- | --
   | 2
   
   
### Current Master Benchmark
```
es5 marked completed in 4795ms and passed 86.66%
es6 marked completed in 4803ms and passed 86.66%
es5 marked (gfm) completed in 5162ms and passed 86.04%
es6 marked (gfm) completed in 5227ms and passed 86.04%
es5 marked (pedantic) completed in 4980ms and passed 70.86%
es6 marked (pedantic) completed in 5040ms and passed 70.86%
```

### This PR Benchmark
```
es5 marked completed in 4673ms and passed 86.66%
es6 marked completed in 4777ms and passed 86.66%
es5 marked (gfm) completed in 4850ms and passed 86.04%
es6 marked (gfm) completed in 4985ms and passed 86.04%
es5 marked (pedantic) completed in 4774ms and passed 70.86%
es6 marked (pedantic) completed in 4922ms and passed 70.86%
```

## Contributor

- [ ]  no tests required for this PR.

## Committer

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -176,13 +176,6 @@ module.exports = class Lexer {
         continue;
       }
 
-      // table no leading pipe (gfm)
-      if (token = this.tokenizer.nptable(src)) {
-        src = src.substring(token.raw.length);
-        tokens.push(token);
-        continue;
-      }
-
       // hr
       if (token = this.tokenizer.hr(src)) {
         src = src.substring(token.raw.length);
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -134,42 +134,6 @@ module.exports = class Tokenizer {
     }
   }
 
-  nptable(src) {
-    const cap = this.rules.block.nptable.exec(src);
-    if (cap) {
-      const item = {
-        type: 'table',
-        header: splitCells(cap[1].replace(/^ *| *\| *$/g, '')),
-        align: cap[2].replace(/^ *|\| *$/g, '').split(/ *\| */),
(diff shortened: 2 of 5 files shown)


UziTech:
> If a row of cells is missing the leading\trailing pipe, then the leading\trailing cell cannot be empty.

Is this in the spec or just unspecified behavior that gfm seems to get wrong?

calculuschild:
It's not in the GFM spec one way or the other, but after playing around a lot with the Github parser, that appears to be the rule. So this is more "replicating what Github's parser does".

calculuschild:
Since this removes an entire Tokenizer, I presume this would be breaking? Is there a standard procedure for grouping multiple breaking PRs together so they can all go out in the same major version bump? (#2126, #2124 , #2112)

UziTech:
When I merge them I just cancel the release in CI until the last one.

UziTech (approved):

calculuschild:
Integrated into #2124

Issue #2021, opened 2021-04-20
Add module field to package.json


benmccann:
**What pain point are you perceiving?.**
`marked` does not play nicely with unbundled development tools like Snowpack, Vite, and WMR. E.g. https://github.com/snowpackjs/analyze-npm/issues/9. These tools highly prefer working with ESM code, but won't discover that an ESM version is available without the `module` field in `package.json`

**Describe the solution you'd like**
Align the ESM and CJS versions to have the same shape (https://github.com/markedjs/marked/pull/1571#issuecomment-562891963) and then add a `module` field to `package.json`

benmccann:
Actually, there is a named export called `parse`, so even without changing the shape, I can get this to work in SvelteKit simply by by adding the `module` field. However, without the `module` field, I need a workaround:

```
			optimizeDeps: {
				include: ['marked']
			},
```

UziTech:
If you want to create a PR we will be releasing v3 soon so any breaking changes can go into that release.

PR #2143, opened 2021-07-23
Add module field to package.json

benmccann:
**Marked version:** `master`

**Markdown flavor:** n/a

## Description

- Fixes https://github.com/markedjs/marked/issues/2021

Allows `marked` to be used with SvelteKit / Vite without the currently required workaround (https://github.com/markedjs/marked/issues/2021#issuecomment-885912209)

## Contributor

## Committer

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "author": "Christopher Jeffrey",
   "version": "2.1.3",
   "main": "./src/marked.js",
+  "module": "./lib/marked.esm.js",
   "browser": "./lib/marked.js",
   "bin": {
     "marked": "bin/marked"


UziTech:
It looks like the module field was remove previously because of an issue with webpack. see https://github.com/markedjs/marked/pull/1571#issuecomment-562891963

benmccann:
Yeah, default exports should be avoided because they're not compatible between CJS and ESM. Luckily there are named exports available, so people don't have to use the default export if they run into that. And since this is going into a major version it's a good time to do breaking changes

UziTech:
using a named export is something I would like to change to in the future along with having marked create instances instead of being one global object. Maybe that will be a v4 fix.

benmccann:
This PR shouldn't be an issue in the meantime. It may mean that different users have to import `marked` slightly differently if they want to use the default export depending on whether they're building for the browser or server, but at least they can import directly from `marked` instead of `./lib/marked.esm.js` if they want to use ESM

UziTech:
This PR is an issue for anyone using webpack according to https://github.com/markedjs/marked/pull/1571#issuecomment-562891963

benmccann:
There's a big difference in this PR vs the last time `module` was attempted to be added, which is that `package.json` now contains a `browser` field. Webpack, by default, will find the `browser` field first, which is the same as `main` and so shouldn't cause any issues for people using Webpack as a result.

Even if someone asked Webpack to use the `module` field for `marked`, it wouldn't stop people from using Webpack. It just wouldn't be backwards compatible with using the `main` field. But that shouldn't matter since the `browser` field will be chosen

The other solution to this would be to use the `exports` field in `package.json`

UziTech:
> Webpack, by default, will find the `browser` field first, which is the same as `main` and so shouldn't cause any issues for people using Webpack as a result.

OK i'm convinced.

> The other solution to this would be to use the `exports` field in `package.json`

There actually is a PR to add the `exports` field #1725 that looks abandoned. If you wanted to create a PR for that we could get that in as well.

UziTech (approved):

benmccann:
> There actually is a PR to add the exports field #1725 that looks abandoned. If you wanted to create a PR for that we could get that in as well.

That PR generally looks good to me. I just left some comments on it. It was adding build output in `lib/marked.esm.mjs` and I wasn't sure if it should be?

styfle (approved):
I'm not a huge fan of `module` since its not standardized, but it seems more tools are relying on it so we can try it out with 3.0 since its semver major

Issue #2165, opened 2021-08-06
Reorder Table Tokens properties



calculuschild:
Working on some custome extensions, I've noticed that Tables seem to be the only token with an odd use of it's `tokens` parameter. Instead of being an array of tokens, it's instead an object which then contains *two* token arrays (header and cells). This means we have an odd special case that we have to handle in walktokens for example, and becomes confusing when writing an extension.

I think it would make sense to reorder these properties to follow the format of other tokens to change 

```
tokens = {
  header : [ ],
  cells : [ ]
}
```

into

```
header.tokens = [ ];
cells.tokens = [ ];
```

This better follows the convention of other similar tokens with "sub-tokens" (e.g. Lists contain list items, which then contain tokens, rather than the list having a tokens object containing all the items). Lists are also a weird special case, but we might as well handle both special cases in the same way, and enforce that at least in our implementation, `tokens` always means an array of tokens.

This wouldn't change functionality, but just provide consistency between tokens.

Changes the token signature of Tables so maybe we can squeeze into version 3.0?

UziTech:
I think I would be ok with it. So `tokens` is always just an array of tokens.

calculuschild:
Right. We are on the same page then. 👍

calculuschild:
Fixed by #2166

PR #2166, opened 2021-08-08
fix: Refactor table tokens

calculuschild:
Pretty small change to address #2165. `header` and `cells` are now formatted as "sub-tokens" (similar to `list items` in a `list`) rather than having separate `headers` and `tokens.headers`. This also means child tokens are now in a `tokens` property to follow the convention of the other lexer tokens for consistency.

Changes the Table signature though so this is breaking and should be in V3.0

##### Previous token signature:
```
{
  type: 'table',
  align: [null, null],
  raw: '| a | b |\n|---|---|\n| 1 | 2 |\n',
  header: ['a', 'b'],
  cells: [['1', '2']],
  tokens: {
    header: [
      [{ type: 'text', raw: 'a', text: 'a' }],
      [{ type: 'text', raw: 'b', text: 'b' }]
    ],
    cells: [[
      [{ type: 'text', raw: '1', text: '1' }],
      [{ type: 'text', raw: '2', text: '2' }]
    ]]
  }
}
```

##### New token signature
```
{
  type: 'table',
  align: [null, null],
  raw: '| a | b |\n|---|---|\n| 1 | 2 |\n',
  header: {
    text: ['a', 'b'],
    tokens: [
      [{ type: 'text', raw: 'a', text: 'a' }],
      [{ type: 'text', raw: 'b', text: 'b' }]
    ]
  },
  cells: {
    text: [['1', '2']],
    tokens: [[
      [{ type: 'text', raw: '1', text: '1' }],
      [{ type: 'text', raw: '2', text: '2' }]
    ]]
  }
}
```

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -348,12 +348,16 @@ module.exports = class Tokenizer {
     if (cap) {
       const item = {
         type: 'table',
-        header: splitCells(cap[1].replace(/^ *| *\| *$/g, '')),
+        header: {
+          text: splitCells(cap[1].replace(/^ *| *\| *$/g, ''))
+        },
         align: cap[2].replace(/^ *|\| *$/g, '').split(/ *\| */),
-        cells: cap[3] ? cap[3].replace(/\n$/, '').split('\n') : []
+        cells: {
+          text: cap[3] ? cap[3].replace(/\n$/, '').split('\n') : []
+        }
       };
 
-      if (item.header.length === item.align.length) {
+      if (item.header.text.length === item.align.length) {
         item.raw = cap[0];
 
         let l = item.align.length;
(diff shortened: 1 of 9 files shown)


calculuschild:
Bonus! CodeQL found a Regex vulnerability for an edge case that we already handled in #2126

UziTech (approved):
LGTM

src/Tokenizer.js
```diff
@@ -370,32 +374,30 @@ module.exports = class Tokenizer {
...
+        item.cells.tokens = [];
 
         // header child tokens
-        l = item.header.length;
+        l = item.header.text.length;
         for (j = 0; j < l; j++) {
-          item.tokens.header[j] = [];
-          this.lexer.inlineTokens(item.header[j], item.tokens.header[j]);
+          item.header.tokens[j] = [];
+          this.lexer.inlineTokens(item.header.text[j], item.header.tokens[j]);
```
UziTech, line 390: Now that I think about it the `tokens` property here is an array of token arrays.

Maybe instead we should do:

```js
{
  header: {
    items: [
      {
        text: 'header cell 1 text',
        tokens: [/* array of tokens for header cell 1 */]
      },
      {
        text: 'header cell 2 text',
        tokens: [/* array of tokens for header cell 2 */]
      }
    ]
  }
	cells: [
    { // row 1
      items: [
        {
          text: 'row 1 cell 1 text',
          tokens: [/* array of tokens for row 1 cell 1 */]
        },
        {
          text: 'row 1 cell 2 text',
          tokens: [/* array of tokens for row 1 cell 2 */]
        }
      ]
    },
    { // row 2
      items: [
        {
          text: 'row 2 cell 1 text',
          tokens: [/* array of tokens for row 2 cell 1 */]
        },
        {
          text: 'row 2 cell 2 text',
          tokens: [/* array of tokens for row 2 cell 2 */]
        }
      ]
    }
  ]
}
```

That seems to be more consistent with the rest of the tokens (including lists)

calculuschild: Hm. Good point. I have a couple more thoughts:

First.... Is it worth it to have `header` and `row` as objects with a single property `items`? An object with a  single property seems redundant and maybe slightly slows things with extra object accesses? What if instead `header` and `rows` are just arrays like `items` is an array? So a List has `Items`, a Table has `Headers` and `Rows`, but they use the same format of being an array of "sub-tokens":

```
{
  type: 'table',
  align: [null, null],
  raw: '| a | b |\n|---|---|\n| 1 | 2 |\n',
  header: [
    {
      text: 'a',
      tokens: [{ type: 'text', raw: 'a', text: 'a' }],
    },
    {
      text: 'b',
      tokens: [{ type: 'text', raw: 'b', text: 'b' }],
    }
  ],
  rows: [ // replace `cells` with `rows`
    [ // row 1
      {
        text: '1',
        tokens: [{ type: 'text', raw: '1', text: '1' }],
      },
      {
        text: '2',
        tokens: [{ type: 'text', raw: '2', text: '2' }],
      }
    ]
  ]
}
```

UziTech: That sounds good

calculuschild: @UziTech This is now changed. Want another look at it?

UziTech (approved):

PR #2203, opened 2021-09-09
Fix detection of orphaned emStrong delimiters

### LINKED ISSUE #2202: Wrong Parsing when mixing asterisk and underline
**Marked version:** 3.0.2 (also tested in 2.1.3)

**Describe the bug**
When mixing asterisk (*) and underline (_) multiple times the parsing result is messed up. 

**To Reproduce**

For eaxmple: the string `_**123**_ **Test**`
See also here:
1. [Marked Demo](https://marked.js.org/demo/?text=_**123**_%20**Test**&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22extensions%22%3A%20null%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)
2. [CommonMark Demo](https://spec.commonmark.org/dingus/?text=_**123**_%20**Test**)

**Expected behavior**
Mixing Underline & asterisk Syntax for bold/italic shoud be substitutable. `123`should be bold and italic. Test should be bold.

calculuschild:
**Marked version:**

3.0.3

## Description

- Fixes #2202

The check for orphaned delimiters (i.e., to skip over the `_` in `**em_strong**`) looks for an isolated delimiter between two pairs of the opposite delimiter symbol (inside a `strong` of the other symbol). However, this was inadvertently including matches for at *any* point in the string, leading to the second `_` here being counted as an orphan between the middle two asterisk pairs:  <code>\_\*\*foo<b>\*\*\_ \*\*</b>bar\*\*</code>

This PR makes the orphan checker start from the beginning of the string to keep it from over-eagerly finding orphans.

## Contributor

## Committer


UziTech (changes requested):
Can you add a test for `*__foo__* __bar__` as well to make sure that is also fixed

UziTech (approved):

davisjam (commented):
No REDOS issues added here.

UziTech:
@calculuschild does this make this while loop irrelevant since `endReg` will only match once at the beginning of the line?

https://github.com/markedjs/marked/blob/784ba5fe9b10dab594c89560d54e314d4c3a229c/src/Tokenizer.js#L587

calculuschild:
> @calculuschild does this make this while loop irrelevant since `endReg` will only match once at the beginning of the line?

No. The `^` beginning of line only applies to the "orphan" checker. https://regexper.com/#%5ECheckOrphan%7CotherwiseCase1%7CotherwiseCase2

The other matches need to keep looping through until we find a valid end delimiter run.

calculuschild:
Hmm.  Hold up. I just realized this case will still be broken:

`_**foo**_ **bar** _**foo**_ `

Needs a bit more work.

calculuschild:
Ah, nevermind. Seems to work just fine. Updated the tests to include those cases though for future reference.

UziTech:
This seems to break `_**foo_bar**_`

calculuschild:
> This seems to break `_**foo_bar**_`

I've got a fix for that too.

calculuschild:
Using a lookahead so the orphan checker doesn't actually consume that last set of asterisks/underscores which was preventing detection of the ending delimiter.

UziTech:
It looks like `*__foo*bar__*` is not supposed to work the same way.

[demo](https://spec.commonmark.org/dingus/?text=*__foo*bar__*)

calculuschild:
Huh. That's really odd.

Ok, I have removed that test.

UziTech (approved):
Nice work!

styfle (approved):

PR #2209, opened 2021-09-27
export helpers

GNHua:
Export helper function in order to re-use them to create React components


UziTech:
I'm not sure we want to export these since they are just implementation details and we would be restricted to updating them with breaking changes in major versions.

If someone wants to use them they can copy the function to their own project so we don't need to worry about maintaining them.

PR #2223, opened 2021-10-05
Expose named exports for ESM build

benmccann:
**Marked version:**

Tested against 3.0.4

**Markdown flavor:** all

## Description

Currently, you can only used named exports from the CJS build. This will write them in such a way that Rollup can figure out they're exposed and should be exported in the ESM build as well.

## Expectation

I should be able to do `import { parse } from 'marked';`. I expect `lib/marked.esm.js` to `export` the `parse` method

## Result

Today I get:
```
'parse' is not exported by 'node_modules/.pnpm/marked@3.0.4/node_modules/marked/lib/marked.esm.js'
```

That's because `lib/marked.esm.js` only has a `default` export

## What was attempted

I tried to do `import { parse } from 'marked';`

## Contributor

## Committer


Conduitry:
Would it be a good idea to add a `pkg.exports` map at the same time, so that native Node ESM consumers (not using a bundler) will also get the ESM version rather than the CJS version?

benmccann:
There's a pending PR for it: https://github.com/markedjs/marked/pull/1725. Maybe not a bad idea, but I'd prefer to do it in a separate PR to keep this one simpler

Conduitry:
Yup, makes sense.

UziTech (approved):

benmccann:
Thanks for the approval @UziTech! Just curious, is there anything else needed before this can be merged?

UziTech:
@splch you should set a version on your dependencies so they don't break automatically.

benmccann:
Here's a fix for the issue: https://github.com/markedjs/marked/pull/2226

PR #2224, opened 2021-10-06
fix: revert #2223

UziTech:
It seems like #2223 turned lib/marked.js into a ES Module which would be a breaking change.

This reverts #2223


UziTech:
looks like #2225 didn't fix it it only removed the `__esModule` property

benmccann:
Here's a fix for the issue: https://github.com/markedjs/marked/pull/2226

PR #2231, opened 2021-10-07
Add an `exports` map to `package.json`.

benmccann:
**Marked version:** `master`

**Markdown flavor:** all

## Description

Adds an `exports` map to `package.json`. There is an existing PR for this, but it changes a bunch of unnecessary things. The [comment on that PR](https://github.com/markedjs/marked/pull/1725#issuecomment-838959788) was that it would be fine to merge if it was rebased, so I'm sending this one.

It may be better to release this as a major version bump because it will stop people from reaching into the internals. They will only be able to use officially exported APIs. Alternatively, I think we could update this to expose the `./src` directory. If it's going to be a major version bump, we might be able to make other changes as well such as setting `"type": "module"`

## Contributor

## Committer

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,13 @@
     "man/",
     "marked.min.js"
   ],
+  "exports": {
+    ".": {
+      "import": "./lib/marked.esm.js",
+      "default": "./lib/marked.js"
+    },
+    "./package.json": "./package.json"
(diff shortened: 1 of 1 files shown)


benmccann:
Sounds good. Moved to that PR

PR #2240, opened 2021-10-15
Docs: Nowrap on list elements

Yash-Singh1:
Simply makes a styling change to the docs to prevent this:

Screenshot from 2021-10-14 18-38-05

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -80,11 +80,14 @@ td, th {
 a {
     color: #0366d6;
     text-decoration: none;
-    white-space: nowrap;
 }
 
 a:hover {
     text-decoration: underline;
+}
+
+nav a {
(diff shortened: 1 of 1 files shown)


Yash-Singh1:
But are there any anchors that are not in the sidebar that have to be wrapped?

Yash-Singh1:
> There definitely could be.

Sure, I will create a commit to fix that

> Plus it doesn't look like they wrap for me even at the smallest width.
> 
> image

The problem occurs when you press on it on Desktop. On mobile, the problem just occurs by default:

Screenshot from 2021-10-14 20-50-38

Yash-Singh1:
Actually looking deeper into this, I don't think this can be solved and should be considered as expected behavior.

Issue #2265, opened 2021-11-03
module import/export issue


paddotk:
**Marked version:**
4.0.0

**Describe the bug**
I updated marked in our project from `^3.0.2` to `4.0.0` (before it worked fine), and after the update I get an error `marked__WEBPACK_IMPORTED_MODULE_5___default(...) is not a function` when using it. It seems that the module is no longer exported properly, 
I tried to change the import `import { marked } from 'marked'` as well as using `marked.parse()` as shown in the readme, but that doesn't seem to help.

**To Reproduce**
Use a webpack/npm project, import by means of `import marked from 'marked'`. Use the `marked()` function.

UziTech:
It looks like somewhere in your code you are trying to import the default export like `import marked from 'marked'`.

Make sure you replace all occurrences with `import { marked } from 'marked'`

AlanMorel:
I'm using TypeScript and can also confirm that I am also experiencing the same issue on `4.0.0`. I am importing it using `import { marked } from "marked";` but getting `'marked' can only be imported using 'import marked = require("marked")' or a default import.`. It is possible that I just need to wait until `@types/marked` is bumped up a major version though.

paddotk:
@UziTech I tried that but didn't work...
@AlanMorel From the error it looks like a Webpack issue. It also shows in my unit tests which are .js files.

AlanMorel:
It seems unlikely to be a webpack-specific issue because the project I'm using Marked on is not using webpack at all.

maurelio1234:
>  It is possible that I just need to wait until `@types/marked` is bumped up a major version though.

I have the same problem and `@types/marked` is still in 3.X

UziTech:
[@types/marked v4](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56967) should be released soon

AlanMorel:
Great, I believe that update to `@types/marked` will resolve my issue, so I will wait for that, thanks! 

Side question though: any reason `marked` itself is not providing the typings for the library? It would both eliminate needing a separate `@types/marked` dependency while also ensuring that you can bump both the library and typings at the same time. Not a huge deal though, just curious.

rtjtc:
Also waiting for `@types/marked` ... Do you guys have a workaround or maybe I should go back to previous stable version?

AlanMorel:
Pretty sure your only options are wait for the updated types (which should be coming soon) or use a previous version.

paddotk:
[...]

PR #2281, opened 2021-11-12
Create separate CJS and UMD builds

benmccann:
**Marked version:** 4.0.1

**Markdown flavor:** n/a

## Description

Fixes https://github.com/markedjs/marked/issues/2265

This comment pointed out that it would be very weird to use a CJS file in the browser: https://github.com/markedjs/marked/issues/2265#issuecomment-966732495

That's true, but it's not actually CJS. It's UMD. And I think UMD was specified for the CDNs, which makes a lot more sense. It's just really confusing the way we have it setup now because the file extension doesn't match the actual file type.

Putting a UMD file as `main` is breaking people in 4.0.1. I didn't realize we were doing that in the last PR because I was thrown off by the file extensions. Bundlers generally know how to deal with CJS and ESM, but not UMD. It's reported that this change fixes things: https://github.com/markedjs/marked/issues/2265#issuecomment-967311475

## Contributor

## Committer


UziTech (approved):

UziTech:
@benmccann You have been helping out a lot recently. Would you like to join @markedjs and help maintain marked?

benmccann:
Thanks for the offer, but I'll probably disappear once this conversion to ESM has settled down :smile:  I've got a fair amount of open source projects on my plate already

styfle (approved):
Its an unfortunate state of the JS ecosystem but this change looks correct, thanks!

splch:
Just to understand this correctly, if we're using marked on the client side, we now will be using `lib/marked.umd.js`?

UziTech:
@splch if you are using marked in a script tag you should use `/marked.min.js`. if you use marked in node or webpack you should use `import { marked } from 'marked';` and the loader should choose the correct file.

PR #2302, opened 2021-11-24
fix: speed up parsing long lists

### LINKED ISSUE #2297: Performance of parse() is 15 times worse in releases after 2.1.3
**Marked version: >2.1.3**

**Describe the bug**
The speed of parse() call with large markdown source was about 15 times better in v2.1.3 than releases after it. Maximum size of markdown source that can be processed in the browser is respectively way smaller.

**To Reproduce**

const startTime = Date.now();
marked.parse(mdData800kB);
console.log('time in milliseconds', Date.now() - startTime);

Output:
time in milliseconds 30000

**Expected behavior**

time in milliseconds 2000

calculuschild:
**Marked version:**

v3.0.0 +

## Description

- Fixes #2297. The rework of Lists in v3.0.0 accidentally introduced a O(n^2) slowdown issue for long documents.

The List tokenizer was using a RegEx to capture the potential next list Item, and then split that captured text to work line-by-line to determine if it had proper indentation, etc. and whether each line should be part of the current list Item.

Problem is, the captured text was literally *the entire document*, so for every potential list item, we were capturing the entire document and then splitting the document into lines. For longer documents, this meant spending the majority of time just splitting the document into lines over and over.

Here's the offending Regex, and notice that `(?:\\n[^\\n]*)*` matches *everything to the end of the document*:

https://github.com/markedjs/marked/blob/d098d554894ac4eeb4f7226dd80a5daebd59911b/src/Tokenizer.js#L193

This PR changes that so we only capture the first line (with it's bullet point), and once we verify that it is a candidate for starting a new list Item, we just traverse the SRC one line at a time. No more mass line-splitting when we really only need to look one line at a time anyway.

Offending line-splitting line where 90% of processing time was spent:

https://github.com/markedjs/marked/blob/d098d554894ac4eeb4f7226dd80a5daebd59911b/src/Tokenizer.js#L205

Needs a bit of cleanup, but it's passing tests.

## Contributor

## Committer


test/specs/bug/adjacent_lists.md
```diff
@@ -1 +0,0 @@
-*foo __bar *baz bim__ bam*
```
UziTech, line 1: Why was this test removed? Is it failing with this PR?

UziTech: nvm I see this was an accidental addition.

calculuschild: That was a test I added playing around trying to debug my program and accidentally committed it to the PR.

test/specs/redos/quadratic_lists.cjs
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  markdown: '- a\n'.repeat(10000),
+  html: `<ul>${'<li>a</li>'.repeat(10000)}</ul>`
+};
```
UziTech, line 4: I can confirm this speeds up parsing long lists.

This test takes about 2 seconds on my machine on master and about 100 ms with this PR.

UziTech (approved):

test/specs/redos/quadratic_lists.cjs
calculuschild: Hooray!

calculuschild:
I want to go over this code once more before it gets merged. I have a nagging sense that there's still some debris left behind I should clean up but exactly where is eluding me at the moment.

lib/marked.cjs
```diff
@@ -561,35 +561,41 @@ var Tokenizer = /*#__PURE__*/function () {
 
         if (!(cap = itemRegex.exec(src))) {
           break;
-        }
+        } //lines = cap[2].split('\n');
```
styfle, line 564: Should this comment be removed?

lib/marked.cjs
```diff
@@ -657,8 +665,7 @@ var Tokenizer = /*#__PURE__*/function () {
           loose: false,
           text: itemContents
         });
-        list.raw += raw;
-        src = src.slice(raw.length);
+        list.raw += raw; //src = src.slice(raw.length);
```
styfle, line 668: Also here, should this comment be removed?

lib/marked.cjs
calculuschild: This was leftover in the compiled file looks like. Recompiled so this should now be up to date.

calculuschild:
@Uzitech Cleaned up the logic how I wanted. Passes the specs but is failing the SNYK security test here and I'm not sure why.

Otherwise, this is now ready to merged.

src/Tokenizer.js
```diff
@@ -304,9 +300,9 @@ export class Tokenizer {
       for (i = 0; i < l; i++) {
         this.lexer.state.top = false;
         list.items[i].tokens = this.lexer.blockTokens(list.items[i].text, []);
-        if (list.items[i].tokens.some(t => t.type === 'space')) {
+        if (!list.loose && list.items[i].tokens.some(t => t.type === 'space')) {
           list.loose = true;
-          list.items[i].loose = true;
+          // list.items[i].loose = true;
```
calculuschild, line 305: Do list items need to be marked as loose or just the list itself? This is a spot we could speed up slightly if we keep it commented out; it doesn't fail any tests.

UziTech: It looks like the parser only uses the `loose` property on the list not the list items. I think it would be fine to not have this on the list items but it might be considered a breaking change.

calculuschild: I will leave that up to you then. If needed we can lump it in with other breaking changes at a later date?

UziTech: Actually it looks like that property is not necessarily correct in the list items anyway. Only the list items with a `space` token have the loose property set to true but it should be all items if any of them have a `space` token.

UziTech: Ya I would say leave it right now and we can remove it later.

calculuschild: Ok, it's uncommented now.

> Actually it looks like that property is not necessarily correct in the list items anyway

Ah yeah, you're right. I guess that mistake didn't get caught because it's never used anywhere.

UziTech (approved):

lib/marked.esm.js
```diff
@@ -587,9 +583,9 @@ class Tokenizer {
       for (i = 0; i < l; i++) {
         this.lexer.state.top = false;
         list.items[i].tokens = this.lexer.blockTokens(list.items[i].text, []);
-        if (list.items[i].tokens.some(t => t.type === 'space')) {
+        if (!list.loose && list.items[i].tokens.some(t => t.type === 'space')) {
           list.loose = true;
-          list.items[i].loose = true;
+          // list.items[i].loose = true;
```
styfle, line 588: Remove this comment too?

lib/marked.umd.js
```diff
@@ -673,11 +666,10 @@
           this.lexer.state.top = false;
           list.items[i].tokens = this.lexer.blockTokens(list.items[i].text, []);
 
-          if (list.items[i].tokens.some(function (t) {
+          if (!list.loose && list.items[i].tokens.some(function (t) {
             return t.type === 'space';
           })) {
-            list.loose = true;
-            list.items[i].loose = true;
+            list.loose = true; // list.items[i].loose = true;
```
styfle, line 672: Another comment that seems out of place here

lib/marked.cjs
```diff
@@ -671,11 +664,10 @@ var Tokenizer = /*#__PURE__*/function () {
         this.lexer.state.top = false;
         list.items[i].tokens = this.lexer.blockTokens(list.items[i].text, []);
 
-        if (list.items[i].tokens.some(function (t) {
+        if (!list.loose && list.items[i].tokens.some(function (t) {
           return t.type === 'space';
         })) {
-          list.loose = true;
-          list.items[i].loose = true;
+          list.loose = true; // list.items[i].loose = true;
```
styfle, line 670: This one too

lib/marked.esm.js
calculuschild: These are just the compiled files which are out of date. They are not present in the actual source code so I don't think they matter here (they get regenerated when we merge, right?), but I will recompile now to see what happens.

calculuschild: There. Lib files were updated.

styfle (approved):

PR #2303, opened 2021-11-24
chore(docs): Improved CLI Examples

alexandercerutti:
## Description

I tried to improve examples for what concerns CLI usages, which are not very clear.
I wanted also to compare Options from Advanced Usage with the one specified in the man file, but I wasn't able to understand quickly if all the options that can be set both through Node.JS API and CLI or if some are valid for CLI and others only for node.js (e.g. baseUrl). Let me know if I can improve also this aspect.

Let me know if I can change or if you don't like something. Thank you!

## Contributor

## Committer

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,12 +47,19 @@ Also read about:
 
 **CLI**
 
+
 ``` bash
+# Example with stdin input
 $ marked -o hello.html
 hello world
 ^D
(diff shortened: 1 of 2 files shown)


UziTech (approved):
Thanks for helping improve marked! :100:

PR #2391, opened 2022-02-27
feat: add raw parameter to paragraph

sxyazi:
**Marked version:** v4.0.12

## Proposal

I'm using the `paragraph` renderer to add LaTex support to my work, but unfortunately, the parameter passed into the paragraph was escaped, as in

```
$$
12\,34 \\
$$
```

will be escaped to

```
$$
12,34 \
$$
```

The backslashes were removed, so I propose to add a `raw` parameter to the paragraph, which represents the original unescaped text, and it looks like:

```typescript
const renderer = {
  paragraph(text: string, raw: string): string {
  },
}
```

## Contributor

## Committer


calculuschild:
@sxyazi This functionality is already possible using [Custom Extensions](https://marked.js.org/using_pro#extensions). Using that approach, the entire token is available and you can access the `raw` parameter that way. Take a look and see if you can use that instead.

sxyazi:
@calculuschild Hi, I just tested that method and it works for me. Thanks, I will close this PR.

calculuschild:
@sxyazi If you've created an extension you think other people would be interested in (LaTeX is a fairly common request) we may put together a list of known extensions so other users can benefit. Feel free to let us know if you want to be included in that. https://github.com/markedjs/marked/discussions/2321#discussioncomment-1800936

sxyazi:
@calculuschild Yeah, I like this idea, I will try to pull the LaTex part of the code out of my project and create a new package.

PR #2405, opened 2022-03-11
add async parser

### LINKED ISSUE #458: async renderer support

UziTech:
**Marked version:** 4.0.12

## Description

Add async parser that can be accessed with the `async: true` option.

This allows using async renderer methods:

```js
const renderer = {
  async image(href, title, text) {
    // do something asynchronous
  }
}
marked.use({renderer, async: true});
const html = await marked(markdown);
```

To prevent slowing down the synchronous parsing I cloned the `Parser` class to create an `AsyncParser` class that mimics `Parser` in every way but returns a promise for the `.parse` and `.parseInline` methods as well as awaiting the results of every renderer method.

---

### Things to consider:

- AsyncParser is about half the speed of the regular parser because it awaits every renderer call.
- Can we prevent duplicate code in parsers?
  - This approach has the major downside of having to maintain two parsing classes. If we change one we will have to remember to change the other. I did mitigate the risks of forgetting a fix to one by running all spec tests through both.
- Should we also make the tokenizer async?
- Should we also make walkTokens async?

---

- Fixes #458 

## Contributor

## Committer

### DIFF
--- a/docs/USING_ADVANCED.md
+++ b/docs/USING_ADVANCED.md
@@ -44,6 +44,7 @@ console.log(marked.parse(markdownString));
 
 |Member      |Type      |Default  |Since    |Notes         |
 |:-----------|:---------|:--------|:--------|:-------------|
+|async     |`boolean`  |`false`   |4.1.0    |Use the asynchronous parser to call renderer functions.|
 |baseUrl     |`string`  |`null`   |0.3.9    |A prefix url for any relative link. |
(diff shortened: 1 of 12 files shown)


UziTech:
I'm going to close this.

The main issues:

- The way we fallback to previous renderer functions when returning `false` needs to be async also.
- Duplicate code makes it harder to maintain.
- Slows down marked considerably even for non async workflow.

PR #2415, opened 2022-03-21
chore: clarify custom extension RegEx anchoring to string start

### LINKED ISSUE #2414: Having Trouble to Understand the Extension Example
I'm new to marked and I would like to create an own extension. I have trouble understanding and migrating the example (https://marked.js.org/using_pro#extensions) to my needs. IMHO the documentation is a bit short on explanations, especially in essential parts of the code.

It seems that the most central part of the extensions are the regular expression matches in start and tokenize.
I have trouble understanding what exactly they are doing and why they need to be written the way they are in order to work for marked.js

When I tried to create my own extension with a much easier regex, my extension ended up consuming and replacing the whole text body instead only the desired syntax pattern. So there must be something crucial that I am missing.

Could you please help me out with this by explaining the essential bits of extensions more thorough? Maybe it's also a good idea to improve the current docs a bit as well.

Thanks for your help and great work :)

calculuschild:
**Marked version:**

4.0.12

**Markdown flavor:** n/a

## Description

- Fixes #2414 core issue (confusion about how Custom Extensions work)

This has been a repeat issue with users missing that the Tokenizer in a Custom Extension needs to be anchored to the beginning of the `src` string. Just a small edit to the docs to clarify. Open to any suggestions on changes here.

## Contributor

## Committer


UziTech:
Looks like the [`tokenizer`](https://github.com/markedjs/marked/blob/master/docs/USING_PRO.md#the-tokenizer--tokenizer) example also isn't anchored.

```js
...
// Override function
const tokenizer = {
  codespan(src) {
    const match = src.match(/\$+([^\$\n]+?)\$+/);
    if (match) {
      return {
        type: 'codespan',
...
```

calculuschild:
> Looks like the [`tokenizer`](https://github.com/markedjs/marked/blob/master/docs/USING_PRO.md#the-tokenizer--tokenizer) example also isn't anchored.

Alright, I added an anchor to that example as well.

UziTech (approved):

PR #2428, opened 2022-03-29
low level stream / iterator support

jimmywarting:
My approche at potentially tackling a streamer / async render

## Description
Add sync parser that can iterated over all token and yield the response back to the user to allow for some sort of streaming / flushing out the result.

the idea is not to concatenate the hole response into one single string, and instead yield all data as it spits it out

This could later allow for a higher level component to await for something if it where to return a promise

To prevent slowing down the synchronous parsing and slamming in async everywhere (like PR: #458) that would otherwise cause the event loop to trigger unnecessary delays everywhere. i opted to take a different approach in the way that the developer will have to wait for a chunk to resolve if it returns a promise. (similar to how i suggested it at https://github.com/markedjs/marked/issues/458#issuecomment-1081954012)

### Things to consider:
  * This approach dose not mean we have to maintain two parsing classes.
  * if some renderer returns a promise then that will be returned to the streamer as a unresolved promise. 

Fixes [async renderer support #458](https://github.com/markedjs/marked/issues/458)

## Contributor
some few stuff is breaking...

## Committer
In most cases, this should be a different person than the contributor. and it also need to be bench tested against prev version

* [ ]  CI is green (no forced merge required).
* [ ]  Squash and Merge PR following [conventional commit guidelines](https://www.conventionalcommits.org/).

### DIFF
--- a/test/unit/marked-spec.js
+++ b/test/unit/marked-spec.js
@@ -615,6 +615,31 @@ used extension2 walked</p>
     });
   });
 
+  it('should allow for iterator', () => {
+    const tokens = lexer(`
+# heading
+\`\`\`JSON
+{ "foo": "bar" }
+\`\`\`
+# footer
+    `);
+    use({
+      renderer: {
+        async code(code, lang) {
+          return `<pre><code>foo</code></pre>`;
+        }
+      }
+    });
+    const res = marked.Parser.getIterator(tokens);
+    return Promise.all([...res]).then(results => {
+      const actual = results.join('\n');
+      expect(actual).toBe(
+        '<h1 id="heading">heading</h1>\n\n'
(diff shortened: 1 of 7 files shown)


UziTech:
This could work. We would have to get the tests working to run an accurate benchmark against master.

test/unit/marked-spec.js
```diff
@@ -615,6 +615,31 @@ used extension2 walked</p>
...
+# footer
+    `);
+    use({
+      renderer: {
+        async code(code, lang) {
+          return `<pre><code>foo</code></pre>`;
+        }
+      }
+    });
+    const res = marked.Parser.getIterator(tokens);
```
jimmywarting, line 633: fyi, the `res` will yield `string, promise<string>, string`

PR #2429, opened 2022-03-30
fix: alphanumeric check on older browsers doesn't throw

### LINKED ISSUE #2430: Fails with a regular expression error in older browsers
**Marked version:** 4.0.12

**Describe the bug**
- Fails with an error `invalid identity escape in regular expression` on older browsers (Firefox <= 72)

Снимок экрана 2022-03-30 в 14 15 14

**To Reproduce**

**Expected behavior**

sheremet-va:
Fixes #2430

**Marked version:** 4.0.12

**Markdown flavor:** n/a

## Description

- Fixes an error on older browsers when doing alphanumeric check. Right now it throws an error `invalid identity escape in regular expression`

Снимок экрана 2022-03-30 в 14 15 14

I think `\w` should be enough for a fallback. I wanted to to `\uXXX-\uXXX` for a regexp, but it was too long: https://util.unicode.org/UnicodeJsps/regex.jsp?a=%5Cp%7BL%7D%5Cp%7BN%7D&b=

## Contributor

## Committer


src/helpers.js
```diff
@@ -247,3 +247,14 @@ export function repeatString(pattern, count) {
   }
   return result + pattern;
 }
+
+export function isAlphanumeric(text) {
+  // \p cases older browsers (Firefox <= 72) to throw an error
```
UziTech, line 252:
```suggestion
  // \p causes older browsers (Firefox <= 72) to throw an error
```

UziTech (changes requested):

UziTech:
Looks like this affects about 7% of people. https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes. The only thing I'm concerned about is without a test this could easily be broken again in the future without us knowing about it.

One alternative could be to use a [babel plugin](https://babeljs.io/docs/en/babel-plugin-proposal-unicode-property-regex) to convert this in the `umd` and/or `min` versions.

calculuschild:
I was gonna say... I think we should leverage Babel for compatibility issues. Building our own polyfills/workarounds for old browsers seems outside the scope of Marked, plus a pain to test and maintain.

UziTech:
I'm going to close this as it is already fixed.

PR #2489, opened 2022-05-30
chore(docs): Update installation instructions

rukshn:
Documentation update

This PR will make the installation commands for CLI and browser clickable and copiable to clipboard.


UziTech (approved):
Thanks! 💯

PR #2570, opened 2022-09-01
GitHub Workflows security hardening

sashashura:
The top level default permissions was read/write when triggered `on: push`. Granted read/write only to the specific job for pushing the release.


UziTech (approved):
Thanks! 💯

joshbruce (approved):

PR #2588, opened 2022-09-19
docs: add known extensions

UziTech:
## Description

Add a list of known extensions to the docs

## Contributor

## Committer


calculuschild:
Should we add a small comment for users who want to add their extension to the list?

UziTech:
> Should we add a small instruction for users who want to add their extension to the list?

I don't think we need any instructions other than to keep the list in alphabetical order, which I put in a comment above the list.

> Also, for the future, are we planning to just list any extension we come across or should there be some kind of criteria for "good" extensions?

I think any extension that is maintained is good. If anyone wants to add or remove an extension they can create a PR.

calculuschild (approved):

Issue #2516, opened 2022-06-27
Breaking change of leading tab in unordered list


ErazerBrecht:
**What pain point are you perceiving?.**
When using 2 leading tabs in a list it starts seeing it as a code block. Which could make sense
But starting from 4.0.15 it also places them on another level which I really can't explain.

Marked has introduced this breaking change between version 4.0.13 & 4.0.14, and 4.0.15 makes it even worse.
Caused by #2434 

[Working example 4.0.13](https://marked.js.org/demo/?text=%23%23%23%20%E2%9A%A0%EF%B8%8F%20Hello%20World%20%E2%9A%A0%EF%B8%8F%0A%0A-%20A%0A-%20%20%20B%0A-%09C%0A-%09D%0A-%09E%0A-%09F&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22extensions%22%3A%20null%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=4.0.13)
[...]

PR #2604, opened 2022-10-10
2516: updated the readme on pedantic usage

lasmil:
- when using tabs in unordered lists, we should use pedantic set to true to obtain a flat list
- closes: https://github.com/markedjs/marked/issues/2516


UziTech:
I don't think this actually fixes #2516. We shouldn't just document things that are broken otherwise the documentation might get missed when it is fixed.

lasmil:
> I don't think this actually fixes #2516. We shouldn't just document things that are broken otherwise the documentation might get missed when it is fixed.

Yeah, I agree with you, this is not a fix, it's just a note that the pedantic option is useful in some scenarios. The reporter of the issue is looking into the solution with pedantic.

UziTech:
It would be nice to have a list of all the things pedantic changes. Pedantic usually isn't a good fix for anything because it could break many other things.

UziTech:
I am going to close this because I don't think we want people to use `pedantic` to fix issues in `commonmark` or `gfm`.

PR #2605, opened 2022-10-10
docs: Add note about zero width unicode characters (#2139)

### LINKED ISSUE #2139: First line is not parsed as markdown
**Marked version:** 2.1.3

**Describe the bug**
I'm using `marked` in my `next.js` app. I have this code:

```
const fs = require('fs');
const marked = require("marked");

var content = fs.readFileSync('/sample/markdown/file.md', 'utf8');
content = marked(content);
```

And this is my `file.md`:

```
# Formal sciences

***

## Tools

- [GeoGebra](https://www.geogebra.org/)
```

And this is my output:

imageedit_1_3630574482

But if I only add one line to the beginning of my `file.md`, then I get this result:

imageedit_3_9689168460

**To Reproduce**

It works on your demo playground. But it doesn't work in my `next.js` app. And I only call your `marked` method passing my file's content. Nothing more.

- Create a `next.js` app
- Add `marked`
- Create an `html` file and serve it.
- Now change the extension of that `HTML` file to `.md`.
- First line is not respected as markdown.

If we create `.md` file in the first place, things work as expected. But when we change the extension of an existing HTML file to .md, first line breaks.

**Expected behavior**
The first line should be respected as markdown too. I can't add an empty line to all of my files.

matijagaspar:
## Description

- Fixes #2139

## What was attempted

Added clarification about ZERO WIDTH unicode characters interfering with parsing. 

## Contributor

## Committer

### DIFF
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -34,6 +34,15 @@ npm install @types/marked # For TypeScript projects
 DOMPurify.sanitize(marked.parse(`<img src="x" onerror="alert('not happening')">`));
 ```
 
+**⚠️ Input: special ZERO WIDTH unicode characters (for example `\uFEFF`) might interfere with parsing. Some text editors add them at the start of the file (see: [#2139](https://github.com/markedjs/marked/issues/2139)).**
+
+```js
+// remove the most common zerowidth characters from the start of the file 
+marked.parse(
+  contents.replace(/^[\u200B\u200C\u200D\u200E\u200F\uFEFF]/,"")
+)
+```
+
 **CLI**
 
 ``` bash


UziTech (approved):
Thanks! :100:

docs/INDEX.md
```diff
@@ -34,6 +34,8 @@ npm install @types/marked # For TypeScript projects
 DOMPurify.sanitize(marked.parse(`<img src="x" onerror="alert('not happening')">`));
 ```
 
+**⚠️ Input: special ZERO WIDTH unicode characters (for example `\uFEFF`) might interfere with parsing. Some text editors add them at the start of the file (see: [#2139](https://github.com/markedjs/marked/issues/2139)).**
+
```
styfle, line 38: Let’s add an example below (like we do for sanitizing)

matijagaspar: Amended the commit

styfle (approved):
Thanks!

Issue #23, opened 2012-01-24
Chrome Renderer crashes


rcarmo:
I've been doing a few tests with marked, and I've found that if I feed it large documents with embedded HTML (or plain HTML alone), it will crash Chrome Renderer.

I've been able to reproduce this reliably by force-feeding it raw HTML - notably the contents of the container div at http://oss.sapo.pt/ (which I'm re-doing using marked, BTW - you'll notice the site currently fetches Markdown and parses it on the fly using showdown...).

Further testing reveals that Safari and Firefox are immune to this, BTW - I can toss in the same documents and they come out fine.

PR #2628, opened 2022-10-27
fix: fix escape characters in links

UziTech:
**Marked version:** 4.1.1

## Description

Fixes [backslash escapes](https://spec.commonmark.org/0.30/#backslash-escapes) for common mark specs [`#23`](https://spec.commonmark.org/0.30/#example-23), [`#24`](https://spec.commonmark.org/0.30/#example-24), and [`#202`](https://spec.commonmark.org/0.30/#example-202)

## Contributor

## Committer

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -99,11 +99,12 @@ export class Tokenizer {
     if (cap) {
       const raw = cap[0];
       const text = indentCodeCompensation(raw, cap[3] || '');
+      const lang = (cap[2] ? cap[2].trim() : cap[2]).replace(this.rules.inline._escapes, '$1');
 
       return {
         type: 'code',
         raw,
-        lang: cap[2] ? cap[2].trim() : cap[2],
+        lang,
         text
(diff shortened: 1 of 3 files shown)


src/Tokenizer.js
```diff
@@ -103,7 +103,7 @@ export class Tokenizer {
       return {
         type: 'code',
         raw,
-        lang: cap[2] ? cap[2].trim() : cap[2],
+        lang: cap[2] ? cap[2].trim().replace(this.rules.inline._escapes, '$1') : cap[2],
```
calculuschild, line 106: Would #2627 be achievable using something this approach, a' la DRY? #2627 works, but that regex is becoming more and more unwieldy.

UziTech: No because that has to do with finding the correct tokens. This only works for removing escapes after the token is already found.

UziTech:
any objections to merging this?

styfle (approved):

PR #2629, opened 2022-10-28
feat: export bin/marked

kapouer:
This allows tools like marked-man to extend marked and still use its cli.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
       "import": "./lib/marked.esm.js",
       "default": "./lib/marked.cjs"
     },
+    "./bin/marked": "./bin/marked.js",
     "./package.json": "./package.json"
   },
   "repository": "git://github.com/markedjs/marked.git",


kapouer:
For example:

```js

#!/usr/bin/node

import { markedMan } from 'marked-man';
import { marked } from 'marked';

marked.use(markedMan);

import 'marked/bin/marked';
```
That makes it easy for an extension to be callable from command-line.

UziTech (approved):
LGTM

styfle (approved):

kapouer:
That's super cool, and marked 4 is so much easier to use.
marked-man becomes almost trivial to write.

UziTech:
Awesome! Maybe we should use marked-man in marked to keep our [man file](https://github.com/markedjs/marked/blob/master/man/marked.1) up to date.

kapouer:
I've started updating marked-man to run with latest marked 4.2, so wait a little bit. I'll submit a PR when ready.

PR #2641, opened 2022-11-06
doc: use marked-man to generate man page

kapouer:
Hi,

please find a proof of concept in this PR.

I think it'd be better to just remove man/marked.1* from the git repository and add them to .gitignore, since now they are generated from marked-man.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,10 @@ clean:
 bench:
 	@node test --bench
 
-man/marked.1.txt:
+man/marked.1.txt: man/marked.1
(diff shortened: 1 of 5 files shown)


kapouer:
Updated, with several improvements.

UziTech:
I agree we should gitignore the man files and build in GitHub actions near https://github.com/markedjs/marked/blob/master/.github/workflows/tests.yml#L63

UziTech:
Done in #2963

Issue #34, opened 2012-02-18
links parsed twice + nested if href == content?


leeoniya:
```
[http://www.facebook.com/developers/](http://www.facebook.com/developers/)
```

gives

```
<p><a href="http://www.facebook.com/developers/"><a href="http://www.facebook.com/developers/">http://www.facebook.com/developers/</a></a></p>
```

should be 

```
<p><a href="http://www.facebook.com/developers/">http://www.facebook.com/developers/</a></p>

```

PR #2652, opened 2022-11-12
fix: fix entity specs

UziTech:
**Marked version:** 4.2.2

## Description

Fix `Entity and numeric character references` specs [#28](https://spec.commonmark.org/0.30/#example-28) and [#34](https://spec.commonmark.org/0.30/#example-34).

In order to fix the last 2 `Entity and numeric character references` specs we would need to convert html entities to percent encoded with this json map: https://html.spec.whatwg.org/entities.json. This would increase the size of marked significantly (almost double 48k -> 84k) and provide very little benefit to the actual output.

## Contributor

## Committer

### DIFF
--- a/src/Renderer.js
+++ b/src/Renderer.js
@@ -32,7 +32,7 @@ export class Renderer {
 
     return '<pre><code class="'
       + this.options.langPrefix
-      + escape(lang, true)
+      + escape(lang)
       + '">'
       + (escaped ? code : escape(code, true))
       + '</code></pre>\n';
@@ -170,7 +170,7 @@ export class Renderer {
     if (href === null) {
       return text;
     }
-    let out = '<a href="' + escape(href) + '"';
+    let out = '<a href="' + href + '"';
     if (title) {
       out += ' title="' + title + '"';
     }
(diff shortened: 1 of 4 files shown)


calculuschild (approved):

calculuschild:
Is there really no built-in JS function to convert html entities?

UziTech:
Unfortunately no. The reason given is because they are HTML entities not JavaScript entities. 🫤

UziTech:
Any objections to merging this?

calculuschild (approved):

styfle:
> This would increase the size of marked significantly (almost double 48k -> 84k)

That would be unfortunate. Perhaps we could add the most common ones to `escapeReplacements`?

src/Renderer.js
```diff
@@ -170,7 +170,7 @@ export class Renderer {
     if (href === null) {
       return text;
     }
-    let out = '<a href="' + escape(href) + '"';
+    let out = '<a href="' + href + '"';
```
styfle, line 173: What about when `href` contains a `'` or `"`? 

Do we have a test for that?

UziTech: the href runs through `encodeURI` in the `cleanURL`

https://github.com/UziTech/marked/blob/57a203182cddb4dc1ae4ac8f7c72650eddf3442d/src/helpers.js#L98

styfle (approved):
Thanks!

PR #2655, opened 2022-11-13
fix: fix multiline setext headings

UziTech:
**Marked version:** 4.2.2

## Description

Allow multiline setext headings. This fixes 3 `Setext headings` spec tests but breaks one test that was only passing because we didn't allow multiline setext headings.

The last setext heading test will actually have to be fixed in the blockquote tokenizer.

There was also one unit test that was wrong so I just deleted it. There are other tests that prevent the same regression.

## Contributor

## Committer


UziTech:
Any objections to merging this?

calculuschild (approved):

styfle (approved):

PR #2701, opened 2023-01-07
chore(ci): add version, unify output

### LINKED ISSUE #2190: marked.js and marked.min.js file's content header does not has version
Like https://cdn.jsdelivr.net/gh/markedjs/marked/marked.min.js content header:
```
/**
 * marked - a markdown parser
 * Copyright (c) 2011-2021, Christopher Jeffrey. (MIT Licensed)
 * https://github.com/markedjs/marked
 */
```
I don't know current version and I don't know how to verify. I think should add version like that:
```
/**
 * marked 3.0.0 - a markdown parser
 * Copyright (c) 2011-2021, Christopher Jeffrey. (MIT Licensed)
 * https://github.com/markedjs/marked
 */
```

LeoDog896:
This adds two closely related changes:
- Unifies two related outputs to only parse the AST once (decreasing build time)
- Adds version to header (fixes #2190)

Concern:
- To grab the version, [import assertions](https://github.com/tc39/proposal-import-assertions) are used. Marked.js supports node v12 and up, but based on the fact that [`package-lock.json` is version 2](https://github.com/markedjs/marked/blob/a1f33a7dfbc491e845530ba9059cd150eb1a38db/package-lock.json#L4) instead of 1 (which was added in Node 16), I'm also assuming the build chain is on v16 and can use import assertions.


UziTech:
I think we used to have the version but the problem is our release pipeline is automated with semantic-release so the steps will be

1. Build files
2. Update version in package.json
3. Release

If we can figure out how to move step 1 after step 2 it could work. I have done it in [signature_pad](https://github.com/szimek/signature_pad/pull/673/commits/8962adc17ce8327d6a3a7798f9856b37eb44e156) by getting the version before the build step.

LeoDog896:
Sort of just applied the changes from your commit here -- not sure if it'll work or not.

UziTech:
I can't find any documentation for `import { defineConfig } from 'rollup';` other than it is used for typescript intellisense.

LeoDog896:
It can be removed -- I put it there *for* the typescript intellisense, but it can be replaced with JSDocs or removed entirely.

UziTech (changes requested):
I think this should be good to go once the lint error is fixed

.github/workflows/tests.yml
```diff
@@ -86,4 +88,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: npx semantic-release
+        run: echo npx semantic-release
```
UziTech, line 91: This will prevent semantic-release from running, but that is ok initially. I will do some testing to make sure this actually works before we release it.

UziTech (approved):

calculuschild (approved):

.github/workflows/tests.yml
LeoDog896: yeah -- that's what I was thinking the purpose of that change was.

styfle (approved):

PR #2728, opened 2023-02-04
fix: always return promise if async

### LINKED ISSUE #2721: async: true inconsistent return type (Promise | string) + docs update
marked provides an `async` option to enable asynchronous `walkTokens`.
`@types/marked` for version 4.x provides the following overload which looks perfectly logical/ reasonable:
```ts
export function marked(src: string, options: marked.MarkedOptions & {async: true}): Promise<string>;
```

However, the `async` true option **only returns a promise if there is an async walkTokens**.
Compare: 
```js
const immediatelyRendered = marked('# h1', { async: true })
// returns  '<h1>h1</h1>'

const notRerendered = marked('# h1', {
  async: true,
  walkTokens(token) {
    setTimeout(() => { token += 'h' }, 1000)
  }
})
// returns  '<h1>h1</h1>'

const soonRendered = marked('# h1', {
  async: true,
  walkTokens() {
    return Promise.resolve()
  }
})
// returns Promise<'<h1>h1</h1>'>
```

Additionally, in the docs it is said under the [Walk Tokens section](https://marked.js.org/using_pro#walk-tokens):

> The return value of the function is ignored.

Evidently, this is not (or no longer) true

When a user instructs marked to render `async: true`  the API should consistently return a Promise, regardless of whether there is an async walkTokens extension included. Additionally, the docs should be updated to "The return value should be a promise for asynchronous walkTokens functions.

Tested using the CJS build of marked 4.2.12

UziTech:
**Marked version:** 4.2.12

## Description

Always return a promise if `async` option is `true`.

- Fixes #2721 

## Contributor

## Committer

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -255,7 +255,7 @@ smartypants('"this ... string"')
 
 <h2 id="walk-tokens">Walk Tokens : <code>walkTokens</code></h2>
 
-The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. The return value of the function is ignored but awaited if running marked in [`async`](#async) mode.
+The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. When [`async`](#async) mode is enabled, the return value is awaited. Otherwise the return value is ignored. 
 
 `marked.use()` can be called multiple times with different `walkTokens` functions. Each function will be called in order, starting with the function that was assigned *last*.
 
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -255,7 +255,7 @@ smartypants('"this ... string"')
 
 <h2 id="walk-tokens">Walk Tokens : <code>walkTokens</code></h2>
 
-The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. The return value of the function is ignored.
+The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. The return value of the function is ignored but awaited if running marked in [`async`](#async) mode.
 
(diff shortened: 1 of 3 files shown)


src/marked.js
```diff
@@ -108,28 +108,35 @@ export function marked(src, opt, callback) {
   function onError(e) {
     e.message += '\nPlease report this to https://github.com/markedjs/marked.';
     if (opt.silent) {
-      return '<p>An error occurred:</p><pre>'
+      const msg = '<p>An error occurred:</p><pre>'
         + escape(e.message + '', true)
         + '</pre>';
+      if (opt.async) {
+        return Promise.reject(msg);
+      }
```
styfle, line 116: It’s unusual to reject with a string. Usually it’s an Error

docs/USING_PRO.md
```diff
@@ -255,7 +255,7 @@ smartypants('"this ... string"')
 
 <h2 id="walk-tokens">Walk Tokens : <code>walkTokens</code></h2>
 
-The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. The return value of the function is ignored.
+The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. The return value of the function is ignored but awaited if running marked in [`async`](#async) mode.
```
styfle, line 258:
```suggestion
The walkTokens function gets called with every token. Child tokens are called before moving on to sibling tokens. Each token is passed by reference so updates are persisted when passed to the parser. When [`async`](#async) mode is enabled, the return value is awaited. Otherwise the return value is ignored. 
```

src/marked.js
UziTech: This is in silent mode so maybe we should actually resolve with the string?

UziTech: Currently the string is returned in silent mode instead of thrown

UziTech: I updated it to resolve on silent and reject with error when not in silent mode. This seems to be the closest async equivalent to what happens when synchronous.

styfle (approved):
Great, thanks!

PR #2783, opened 2023-04-22
chore(docs): update renderer docs

### LINKED ISSUE #2082: Renderer not merged when used in marked() options
**Marked version:** 2.0.6

**Markdown flavor:** n/a

## Expectation

Based on the documentation for [marked.use()](https://marked.js.org/using_pro#use), which states that:

> The renderer and tokenizer options can be an object with functions that will be merged into the renderer and tokenizer respectively.

and also based on the fact that description of `renderer` option in the [Advanced section](https://marked.js.org/using_advanced#options) I would expect that the option passed to `marked()` would behave the same way as the option passed to `marked.use()`.

I therefore passed in a renderer with just `image()` implemented.

## Result

I got an exception that says `renderer.paragraph()` is not a function.

After looking at the source code, it does not seem like the renderer is "merged into" anything when passed directly to `marked()`. Documentation should clarify this, preferably before linking to the section on extending.

## What was attempted

```
let renderer = attachments => ({
  image (href, title) {
    if (!attachments) return false
    return `<img src="${attachments.get(href)?.url || href}" alt="${title}>`
  },
})

// ....

let render = (text, attachments) => marked(text, { renderer: renderer(attachments) })
```

## Call stack & console log

```TypeError: this.renderer.paragraph is not a function```

### PR DESCRIPTION
**Pull request by stephenhutchings**
**Marked version:** latest

**Markdown flavor:** n/a

## Description

- Fixes #2082 

## Contributor

## Committer

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -60,9 +60,11 @@ Marked provides methods for directly overriding the `renderer` and `tokenizer` f
 
 <h2 id="renderer">The Renderer : <code>renderer</code></h2>
 
-The renderer defines the HTML output of a given token. If you supply a `renderer` object to the Marked options, it will be merged with the built-in renderer and any functions inside will override the default handling of that token type.
+The renderer defines the HTML output of a given token. If you supply a renderer in the options object passed to `marked.use()`, any functions in the object will override the default handling of that token type.
 
-Calling `marked.use()` to override the same function multiple times will give priority to the version that was assigned *last*. Overriding functions can return `false` to fall back to the previous override in the sequence, or resume default behavior if all overrides return `false`. Returning any other value (including nothing) will prevent fallback behavior.
+Calling `marked.use()` to override the same function multiple times will give priority to the version that was assigned last. Overriding functions can return false to fall back to the previous override in the sequence, or resume default behavior if all overrides return false. Returning any other value (including nothing) will prevent fallback behavior.
+
+Note that a renderer created with `new Renderer()` will not be created with the default token handlers.
 
 **Example:** Overriding output of the default `heading` token by adding an embedded anchor tag like on GitHub.
 
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -60,7 +60,7 @@ Marked provides methods for directly overriding the `renderer` and `tokenizer` f
 
 <h2 id="renderer">The Renderer : <code>renderer</code></h2>
 
-The renderer defines the HTML output of a given token. If you supply a renderer in the options object passed to `marked.use()`, any functions in the object will override the default handling of that token type.
+The renderer defines the HTML output of a given token. If you supply a `renderer` in the options object passed to `marked.use()`, any functions in the object will override the default handling of that token type.
 
 Calling `marked.use()` to override the same function multiple times will give priority to the version that was assigned *last*. Overriding functions can return `false` to fall back to the previous override in the sequence, or resume default behavior if all overrides return `false`. Returning any other value (including nothing) will prevent fallback behavior.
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/USING_PRO.md [hunk 1]
```diff
@@ -60,10 +60,12 @@ Marked provides methods for directly overriding the `renderer` and `tokenizer` f
 
 <h2 id="renderer">The Renderer : <code>renderer</code></h2>
 
-The renderer defines the HTML output of a given token. If you supply a `renderer` object to the Marked options, it will be merged with the built-in renderer and any functions inside will override the default handling of that token type.
+The renderer defines the HTML output of a given token. If you supply a `renderer` in the options object passed to `marked.use()`, any functions in the object will override the default handling of that token type.
 
 Calling `marked.use()` to override the same function multiple times will give priority to the version that was assigned *last*. Overriding functions can return `false` to fall back to the previous override in the sequence, or resume default behavior if all overrides return `false`. Returning any other value (including nothing) will prevent fallback behavior.
 
+Note that a renderer created with `new Renderer()` will not be created with the default token handlers.
```
UziTech, line 67: What do you mean by this? Are you saying that using `marked.use({ renderer: new Renderer() });` will not use the default renderers, but it will.

UziTech: I think you are confused between the difference between `marked.use()` and `marked.setOptions()` or passing options directly to marked like `marked.parse('...', options)`.

`marked.use()` is for creating extensions including extending the renderer and tokenizer by providing an object.  
If you pass an object like:

```js
marked.use({ renderer: {
  em(text) {
    return `<em class="my-em-class">${text}</em>`
  }
}})
```

the `em` renderer will be overwritten but all other renderers will use the default.

`marked.setOptions()` and `marked.parse('...', options)` are for setting options.  
If you pass an object like:

```js
marked.parse('...', { renderer: {
  em(text) {
    return `<em class="my-em-class">${text}</em>`
  }
}})
```

the whole renderer will be overwritten and it will fail to render tokens other than `em`.

To override the renderer in the options you will have to extend `Renderer` like:

```js
import { marked, Renderer } from 'marked';

class MyRenderer extends Renderer {
  em(text) {
    return `<em class="my-em-class">${text}</em>`
  }
};

marked.parse('...', { renderer: new MyRenderer() });
```

Both are used for different things and have different benefits.

stephenhutchings: Yes sorry, you're correct @UziTech. This line should go.

stephenhutchings:
```js
class MyRenderer extends Renderer {
  em(text) {
    return `<em class="my-em-class">${text}</em>`
  }
};
```

Maybe this part is obvious to others, but I was scratching my head trying to extend the renderer.

Perhaps it's also worth including in the docs, if it isn't there already.

UziTech: Another way that is commonly used to extend the renderer is to just reassign the renderer functions like:

```js
const renderer = new Renderer();
renderer.em = (text) => {
  return `<em class="my-em-class">${text}</em>` 
};

marked.parse('...', { renderer });
```

I think we want people to use `marked.use()` and move people away from using `marked.setOptions()` and `marked.parse('...', options)`.

Once we have a way to create an instance of marked (#907) I think those methods will be deprecated.

stephenhutchings: Makes sense. I was trying to avoid `marked.use` for the same reason listed in that issue.

UziTech (approved):

UziTech:
Thanks for updating this! :100:

stephenhutchings:
@UziTech No problem, and thanks a lot for your helpful comments!

PR #2795, opened 2023-05-03
Suppress deprecation warnings for default values

### LINKED ISSUE #2793: Erroneous warnings about deprecated options when using parse()
**Marked version:**
5.0.0

**Describe the bug**
Calling `parse` emits erroneous warnings about deprecated options that are not actually being used.

**To Reproduce**
1. Install `marked@5.0.0`
2. Run `require('marked').parse('foo')` in REPL
3. Observe warnings

**Expected behavior**
The warnings should only be emitted for deprecated options that I am using.

mateuscruz:
**Marked version:**

5.0.0

**Markdown flavor:** n/a

## Description

- Fixes #2793

## Expectation

When no optionss are passed to `marked()`, no deprecation warnings should be shown

## Result

`marked` shows several deprecation warnings for unused options.

## What was attempted

Taking a quick look, I noticed that a modified `opt` object that is populated with marked defaults was being passed to the `checkDeprecations` function, instead of the original `opt` object.  Using the original `opt` object suppresses unnecessary deprecation warnings.

## Contributor

## Committer

### DIFF
--- a/src/marked.js
+++ b/src/marked.js
@@ -64,7 +64,7 @@ function parseMarkdown(lexer, parser) {
         + Object.prototype.toString.call(src) + ', string expected'));
     }
 
-    checkDeprecations(opt, callback);
+    checkDeprecations(origOpt, callback);
 
     if (opt.hooks) {
       opt.hooks.options = opt;


UziTech:
This will also only alert if the user passes the options to `marked.parse`. If the options are set using `marked.setOptions` or `marked.use` they will not be in `origOpt`

mateuscruz:
> We still want to alert people that the options are deprecated and will change in the future if they are left as the default.

I disagree. I don't see the point of alerting people of things that will change in the future that they're _not_ using . I see the point in alerting them if they _are_ using it. I haven't seen this behavior in any other dependency I have used before. I would argue that the common practice is to alert only if you're using it. I use marked only to parse markdown to html. No advanced features. I just call `marked(text)` and that's it. I think it's a fair assumption that many `marked` users do the same and all this is doing for them is pollute their logs with warnings they can do nothing about.

> This will also only alert if the user passes the options to marked.parse. If the options are set using marked.setOptions or marked.use they will not be in origOpt

That's a fair point. I'll look into it.

UziTech:
> I don't see the point of alerting people of things that will change in the future that they're not using.

The problem is that they are using them by not turning them off. I think the correct solution should be to turn them off by default but we thought that was a bigger change than telling people to turn them off if they don't want to fill up their logs.

quinton-ashley:
@UziTech You should actually listen to the people using your code instead of doubling down and defending your mistakes. You are wrong and you need to remove the warnings if you want people to enjoy using marked.

mateuscruz:
Well it was worth a shot. Thanks for the feedback @UziTech. Closing this one out.

PR #2796, opened 2023-05-04
fix: only warn if langPrefix is changed

UziTech:
**Marked version:** 5.0.0

## Description

This removes the warning for langPrefix unless it is changed.

The default warnings can be silenced by setting the following options:

```js
marked.use({
  mangle: false,
  headerIds: false
});
```

## Contributor

## Committer

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -245,7 +245,7 @@ export function checkDeprecations(opt, callback) {
     console.warn('marked(): sanitize and sanitizer parameters are deprecated since version 0.7.0, should not be used and will be removed in the future. Read more here: https://marked.js.org/#/USING_ADVANCED.md#options');
   }
 
-  if (opt.highlight || opt.langPrefix) {
+  if (opt.highlight || opt.langPrefix !== 'language-') {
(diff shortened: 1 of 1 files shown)


calculuschild (approved):
Ok. This seems reasonable.

src/defaults.js
```diff
@@ -9,7 +9,7 @@ export function getDefaults() {
     headerPrefix: '',
     highlight: null,
     hooks: null,
-    langPrefix: 'language-',
+    langPrefix: null,
```
styfle, line 12: Technically this is semver major because its changing the public API

```js
import marked from 'marked';
const result = marked.getDefaults();
```

joshbruce: Have we considered using (and is it possible to use) empty types instead of `null`?

Coming at it from someone who doesn't know the architecture and intent, I'd need to search and find what's expected from the argument list, yeah?

UziTech: That is true. And any extensions that rely on `options.langPrefix` could break.

UziTech: I changed this PR to only check if langPrefix has changed from it's default value before warning about it.

styfle (approved):

PR #2802, opened 2023-05-08
Clarify deprecation messages for default parameters

### LINKED ISSUE #2797: CLI deprecation warning
## Marked version
5.0.1

## Describe the bug
Since the last upgrade (from v4.x to 5.0.1) I'm seeing the following warnings:
```
marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-mangle.
marked(): headerIds and headerPrefix parameters are deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-gfm-heading-id.
```

I don't use the js API, just run the CLI via npm scripts.

## To Reproduce
This is the command I use: `marked --gfm -i CHANGELOG.md -o changelog.html`

## Expected behavior
No warnings :-)

calculuschild:
**Marked version: 5.0.1**

**Markdown flavor:** n/a

## Description

- Fixes #2797 

Adds a few clarifying words to the deprecation messages for those parameters that are enabled by default. Users are confused about why they are receiving warnings for parameters they didn't manually set; this help indicate that they are enabled by default and they can also just disable the setting if they don't want to get the extension.

I could see changing "historically" to "currently" (or just removing), if that is more clear wording.

## Contributor

## Committer

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -250,7 +250,7 @@ export function checkDeprecations(opt, callback) {
   }
 
   if (opt.mangle) {
-    console.warn('marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Note the "mangle" parameter is historically enabled by default, but can be manually disabled, or instead use https://www.npmjs.com/package/marked-mangle.');
+    console.warn('marked(): mangle parameter is enabled by default, but is deprecated since version 5.0.0, and will be removed in the future. To clear this warning, install https://www.npmjs.com/package/marked-mangle, or disable by setting `{mangle: false}`.');
   }
 
   if (opt.baseUrl) {
@@ -266,7 +266,7 @@ export function checkDeprecations(opt, callback) {
   }
 
   if (opt.headerIds || opt.headerPrefix) {
-    console.warn('marked(): headerIds and headerPrefix parameters are deprecated since version 5.0.0, should not be used and will be removed in the future. Note the "headerIds" parameter is historically enabled by default, but can be manually disabled, or instead use https://www.npmjs.com/package/marked-gfm-heading-id.');
+    console.warn('marked(): headerIds and headerPrefix parameters enabled by default, but are deprecated since version 5.0.0, and will be removed in the future. To clear this warning, install  https://www.npmjs.com/package/marked-gfm-heading-id, or disable by setting `{headerIds: false}`.');
   }
 }
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -250,7 +250,7 @@ export function checkDeprecations(opt, callback) {
   }
 
   if (opt.mangle) {
-    console.warn('marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-mangle.');
+    console.warn('marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Note the "mangle" parameter is historically enabled by default, but can be manually disabled, or instead use https://www.npmjs.com/package/marked-mangle.');
   }
 
   if (opt.baseUrl) {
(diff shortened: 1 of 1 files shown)


UziTech (approved):

UziTech:
We want to alert people who don't set it but still want it that it is deprecated and let them know they should use the extension.

I'm not sure how much time should pass before we change the default to false for those settings but it should probably be soon. This will be a breaking change of course so we will need to release marked v6

styfle:
The message is still confusing to me.

The beginning "should not be used and will be removed in the future" seems contradictory to he second part "can be manually disabled".

I think anyone getting this warning today is going to immediately set `{mangle: false, headerIds: false}` to silence it, so we are basically increasing usage 😅

Maybe we should create an upgrade guide and have the warning print a link to it.

calculuschild:
> The beginning "should not be used and will be removed in the future" seems contradictory to he second part "can be manually disabled".

I'm not sure I understand your point here. Mangle by default is "true". However, is is being deprecated, so users need to disable it if they want the deprecation message to go away.

> I think anyone getting this warning today is going to immediately set `{mangle: false, headerIds: false}` to silence it, so we are basically increasing usage 😅

No, that would decrease usage?

styfle:
> (or use the separate extension) if they want the deprecation message to go away.

Does using the extension automatically disable it?

calculuschild:
Ah, probably not. I guess they need to disable the Mangle option either way to get rid of the warning. Then *if they still want the mangle functionality*, they should get the extension.

src/helpers.js
```diff
@@ -250,7 +250,7 @@ export function checkDeprecations(opt, callback) {
   }
 
   if (opt.mangle) {
-    console.warn('marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-mangle.');
+    console.warn('marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Note the "mangle" parameter is historically enabled by default, but can be manually disabled, or instead use https://www.npmjs.com/package/marked-mangle.');
```
styfle, line 253:
```suggestion
    console.warn('marked(): mangle parameter is deprecated since version 5.0.0, and will be removed in the future. Use `{mangle: false}` to hide this warning and optional enable using https://www.npmjs.com/package/marked-mangle.');
```

src/helpers.js
```diff
@@ -266,7 +266,7 @@ export function checkDeprecations(opt, callback) {
   }
 
   if (opt.headerIds || opt.headerPrefix) {
-    console.warn('marked(): headerIds and headerPrefix parameters are deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-gfm-heading-id.');
+    console.warn('marked(): headerIds and headerPrefix parameters are deprecated since version 5.0.0, should not be used and will be removed in the future. Note the "headerIds" parameter is historically enabled by default, but can be manually disabled, or instead use https://www.npmjs.com/package/marked-gfm-heading-id.');
```
styfle, line 269:
```suggestion
    console.warn('marked(): headerIds and headerPrefix parameters are deprecated since version 5.0.0, and will be removed in the future. Use `{headerIds: false}` to hide this warning and optionally enable using https://www.npmjs.com/package/marked-gfm-heading-id');
```

styfle (changes requested):
I see. That's what I thought. So the message is not very clear. Let's update the message to make this clear 👍

UziTech:
The extension does disable it

UziTech:
The header id extension doesn't but I will fix that tonight.

UziTech:
I fixed the header-ids extension. :+1:

calculuschild:
Alright. So if the extension disables the setting, is the wording clear as-is, @styfle?

styfle:
@calculuschild I updated my suggestions

calculuschild:
My main goal with this PR was to alert users that these options are enabled by default, to address the confusion of "why am I getting warnings for this feature". So something to that effect needs to remain in the message.

I will take your suggestions, @styfle, and tweak to reintroduce the mention of "enabled by default" so this can still solve the issue it was intended to.

styfle:
We need to make sure the message is actionable. 

Alerting for the sake of alerting is going to cause more pain for users.

calculuschild:
Right. Actionable, but also explain why the action is needed. We can do both.

styfle (approved):
Perfect, thanks! 🥇

PR #2805, opened 2023-05-14
fix: Migrate to Typescript

### LINKED ISSUE #1732: Add Declaration Files
Now I need install @types/marked before using marked in TS.
 
Package declaration files with marked.

DreierF:
## Description

This PR migrated the codebase to Typescript and automatically generates and publishes a declaration file which fixes #1732.
I tried to keep the actual code changes that change semantics of the code as small as possible. To simplify the publishing I replaced the rollup setup with tsup, which makes bundling the declaration files a lot easier. I started off with the declarations from Definitely Typed pulled them into the actual code as far as it made sense to me. I also copied over the tests from Definitely Typed.

In contrast to the Definitely Typed types the types are no longer part of the marked Namespace i.e. `marked.MarkedOptions` -> `MarkedOptions` as I did not find a way to preserve both the call signature on the namespace (`marked(...)`) in addition to exporting the types within the namespace. As this has no effect on runtime behavior and only requires trivial changes in Typescript code on user side this was the best tradeoff I could think of. Also namespaces are generally no longer recommended in Typescript so this seemed like a good direction anyway.

I also had to rename most internal symbols i.e. `Tokenizer` to `_Tokenizer` to fix an issue in the generated `.d.ts` file, which would otherwise look like 
```
namespace marked {
   var Tokenizer: typeof Tokenizer; // Cyclic reference while the second Tokenizer actually did not mean the variable, but the import within the file. Also aliasing the classes on import did not fix the issue.
}
```

## Future work
Obviously there is a bunch of places that could be more strictly/cleanly typed, but for now I focused on the exposed API as fixing the type errors within the functions will most of the time require code refactorings which would have made the PR un-reviewable.

@UziTech Please have a look and let me know what you think :)
I have pulled out the renaming from *.js to *.ts to a separate commit to make it easier to review.

## Contributor

## Committer


UziTech:
Thank you for this! :100:

This looks like a great start. I have skimmed it, but I will do a more thorough review within the next week. 

A couple things I see that could change:

- try to remove `any` wherever possible. If it requires `any` add a comment about why and maybe what needs to change to update it in the future.
- instead of `@ts-ignore` can we do `@ts-expect-error` with a comment why it errors. That will allow us to fix it easier in the future.

DreierF:
Most of the `any`/`ts-ignore` were there because you'd have to change the actual code (not just the type annotations) to make it work with Typescript. I got rid of most `any` and `ts-ignore` now and replaced the remaining with `ts-expect-error`.

src/Lexer.ts
```diff
@@ -144,7 +161,7 @@ export class Lexer {
     while (src) {
       if (this.options.extensions
         && this.options.extensions.block
-        && this.options.extensions.block.some((extTokenizer) => {
+        && this.options.extensions.block.some((extTokenizer: any) => {
```
UziTech, line 164:
```suggestion
        && this.options.extensions.block.some((extTokenizer: TokenizerExtension['tokenizer']) => {
```

src/Parser.ts
```diff
@@ -1,44 +1,50 @@
...
+  static parseInline(tokens: Token[], options?: MarkedOptions) {
+    const parser = new _Parser(options);
     return parser.parseInline(tokens);
   }
 
   /**
    * Parse Loop
    */
-  parse(tokens, top = true) {
+  parse(tokens: Token[], top = true): string | undefined {
```
UziTech, line 47: This should always return string. If it is undefined it should be changed to an empty string.

src/Parser.ts
```diff
@@ -208,7 +214,7 @@ export class Parser {
   /**
    * Parse Inline Tokens
    */
-  parseInline(tokens, renderer) {
+  parseInline(tokens: Token[], renderer?: _Renderer | _TextRenderer): string | undefined {
```
UziTech, line 217:
```suggestion
  parseInline(tokens: Token[], renderer?: _Renderer | _TextRenderer): string {
```

any undefined should be an empty string.

src/Renderer.ts
```diff
@@ -1,19 +1,22 @@
-import { defaults } from './defaults.js';
+import { _defaults } from './defaults.ts';
 import {
   cleanUrl,
   escape
-} from './helpers.js';
+} from './helpers.ts';
+import type { MarkedOptions } from './MarkedOptions.ts';
+import { Slugger } from './marked.js';
```
UziTech, line 7:
```suggestion
import { Slugger } from './marked.ts';
```

src/helpers.ts
```diff
@@ -1,19 +1,23 @@
...
+
 /**
  * Helpers
  */
 const escapeTest = /[&<>"']/;
 const escapeReplace = new RegExp(escapeTest.source, 'g');
 const escapeTestNoEncode = /[<>"']|&(?!(#\d{1,7}|#[Xx][a-fA-F0-9]{1,6}|\w+);)/;
 const escapeReplaceNoEncode = new RegExp(escapeTestNoEncode.source, 'g');
-const escapeReplacements = {
+const escapeReplacements: Record<string, string> = {
```
UziTech, line 11:
```suggestion
const escapeReplacements = {
```

Not sure why this is needed

UziTech (changes requested):
This is going to be a breaking change so I think it will take some time to actually get merged in.

src/helpers.ts
DreierF: You are right. This also works without explicit type. Removed it.

src/Lexer.ts
```diff
@@ -1,13 +1,15 @@
...
+import { _defaults } from './defaults.ts';
+import { block, inline } from './rules.ts';
+import { repeatString } from './helpers.ts';
+import type { Token, TokensList } from './Tokens.ts';
+import type { MarkedOptions, Rules } from './MarkedOptions.ts';
 
 /**
  * smartypants text replacement
- * @param {string} text
+ * @param text
```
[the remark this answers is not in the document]
DreierF: No. I also removed all other superfluous @param comments now.

DreierF:
Thanks for the review. I hope addressed all your comments now and rebased the changes on master.

UziTech:
@styfle @calculuschild should we focus on this for v6? and setting deprecated options to false?

calculuschild:
Yeah, sounds good.

UziTech (changes requested):
@DreierF if you can rebase this I think we can merge this next so we don't have to keep rebasing.

DreierF:
I'm done with the rebase. I squashed all the rework commits into a single commit to make the rebase less painful.

src/Instance.ts
```diff
@@ -317,10 +324,10 @@ export class Marked {
...
       if (opt.async) {
         return Promise.resolve(opt.hooks ? opt.hooks.preprocess(src) : src)
-          .then(src => lexer(src, opt))
-          .then(tokens => opt.walkTokens ? Promise.all(this.walkTokens(tokens, opt.walkTokens)).then(() => tokens) : tokens)
-          .then(tokens => parser(tokens, opt))
-          .then(html => opt.hooks ? opt.hooks.postprocess(html) : html)
+          .then(src => lexer(src, opt as MarkedOptions))
+          .then(tokens => (opt as MarkedOptions).walkTokens ? Promise.all(this.walkTokens(tokens, (opt as MarkedOptions).walkTokens!)).then(() => tokens) : tokens)
+          .then(tokens => parser(tokens, opt as MarkedOptions))
+          .then(html => (opt as MarkedOptions).hooks ? (opt as MarkedOptions).hooks!.postprocess(html) : html)
```
styfle, line 330: Something seems wrong here - we shouldn't need to type cast on every line.

Can we do it once above (preferably in a runtime safe way)?

src/Slugger.ts
```diff
@@ -1,31 +1,30 @@
...
-  serialize(value) {
+  serialize(value: string) {
     return value
       .toLowerCase()
       .trim()
-      // remove html tags
+    // remove html tags
       .replace(/<[!\/a-z].*?>/ig, '')
-      // remove unwanted chars
+    // remove unwanted chars
```
styfle, line 19: Looks like indentation changed unexpectedly

src/Slugger.ts
```diff
@@ -1,31 +1,30 @@
...
   }
 
   /**
-   * Finds the next safe (unique) slug to use
-   * @param {string} originalSlug
-   * @param {boolean} isDryRun
-   */
-  getNextSafeSlug(originalSlug, isDryRun) {
+     * Finds the next safe (unique) slug to use
+     */
```
styfle, line 26: indentation looks wrong here

src/helpers.ts
```diff
@@ -48,18 +49,14 @@ export function unescape(html) {
...
- * @param {string | RegExp} regex
- * @param {string} opt
- */
-export function edit(regex, opt) {
+export function edit(regex: Rule, opt?: string) {
   regex = typeof regex === 'string' ? regex : regex.source;
   opt = opt || '';
   const obj = {
-    replace: (name, val) => {
+    replace: (name: string | RegExp, val: any) => {
```
styfle, line 56: Can we have a more specific type than `any`?

src/rules.ts
```diff
@@ -157,7 +213,9 @@ block.pedantic = {
 /**
  * Inline-Level Grammar
  */
-export const inline = {
+// Not all rules are defined in the object literal
+// @ts-expect-error
+export const inline: Record<InlineRuleNames, Rule> & Record<InlineSubRuleNames, Rules> & Rules = {
```
styfle, line 218: If we use `as const` at the end of `inline`, instead of using types here, would that solve the problem below so we could remove `as Rule`?

test/types/marked.ts
```diff
@@ -0,0 +1,348 @@
+/* eslint-disable */
+import { marked } from 'marked';
+import { expectType } from "ts-expect";
```
styfle, line 3: Mixing single quotes and double quotes

tsconfig-type-test.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "target": "es2018",
```
styfle, line 3:
```suggestion
    "target": "es2022",
```

We support Node.js 18 or newer so we can target ES 2022

https://github.com/markedjs/marked/blob/a26f002d1819462c6cae75bb0c26dc064e17b4db/package.json#L91-L93

tsconfig-type-test.json
```diff
@@ -0,0 +1,22 @@
...
+    "module": "esnext",
```
styfle, line 4: I think this should be `nodenext` to handle the case of file extensions and all the other strict rules for Node.js ESM

```suggestion
    "module": "nodenext",
```

tsconfig-type-test.json
```diff
@@ -0,0 +1,22 @@
...
+    "isolatedModules": true,
+    "strict": true,
+    "verbatimModuleSyntax": true,
+    "noEmit": true,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "bundler",
+    "noImplicitAny": false,
```
styfle, line 11: We should use `noImplicitAny: true` which should be default when `strict: true`

tsconfig.json
styfle, line 1: Same comments on this config file too

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "target": "es2018",
+    "module": "esnext",
```
styfle, line 4:
```suggestion
    "module": "nodenext",
```

src/Instance.ts
```diff
@@ -1,49 +1,56 @@
...
-import { TextRenderer } from './TextRenderer.js';
-import { Slugger } from './Slugger.js';
+import { _getDefaults } from './defaults.js';
+import { _Lexer } from './Lexer.js';
+import { _Parser } from './Parser.js';
+import { _Hooks } from './Hooks.js';
+import { _Renderer } from './Renderer.js';
+import { _Tokenizer } from './Tokenizer.js';
+import { _TextRenderer } from './TextRenderer.js';
+import { _Slugger } from './Slugger.js';
```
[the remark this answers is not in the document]
DreierF: As I mentioned already in the description, I had to rename most internal symbols i.e. `Tokenizer` to `_Tokenizer` to fix an issue in the generated `.d.ts` file, which would otherwise look like

```
namespace marked {
   var Tokenizer: typeof Tokenizer; // Cyclic reference while the second Tokenizer actually did not mean the variable, but the import within the file. Also aliasing the classes on import did not fix the issue.
} 
```

Do you have any better idea on how to overcome this issue?
You should be able to revert that once you get rid of the `marked` export and it's extension fields.

src/Instance.ts
DreierF: The problem is that `opt` has type `MarkedOptions | ResultCallback` due to the function overload. I tried to convert the code to Typescript (at first) without changing the resulting code. I added a new variable now to get rid of the casts.

src/Slugger.ts
DreierF: 🤷‍♂️ Must have been the IntelliJ internal formatter. Tried to revert those

src/helpers.ts
DreierF: Sure. Adjusted to `string | RegExp`, but had to adapt the code below to handle it in a type safe way

src/rules.ts
DreierF: No because it would then complain for each `inline.additionalRule = ...` line with `TS2551: Property  _punctuation  does not exist on type` or `TS2540: Cannot assign to  punctuation  because it is a read-only property.`

test/helpers/helpers.js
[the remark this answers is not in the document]
DreierF: I would suggest to migrate the tests to Typescript in a separate step

tsconfig-type-test.json
DreierF: This would require a lot of additional type casts `@ts-expect-error` in places where the current API is nearly impossible to statically type.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "target": "es2022",
+    "module": "NodeNext",
+    "isolatedModules": true,
+    "strict": true,
+    "verbatimModuleSyntax": true,
+    "noEmit": true,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "nodenext",
```
styfle, line 10: Nit pick: `module` and `moduleResolution` are using a different case

package.json
```diff
@@ -65,27 +67,31 @@
...
     "test:unit": "npm test -- test/unit/**/*-spec.js",
     "test:specs": "npm test -- test/specs/**/*-spec.js",
     "test:lint": "eslint .",
     "test:redos": "node test/recheck.js > vuln.js",
     "test:update": "node test/update-specs.js",
     "rules": "node test/rules.js",
-    "bench": "npm run rollup && node test/bench.js",
+    "bench": "npm run build && node test/bench.js",
     "lint": "eslint --fix .",
+    "type-check": "tsc",
```
styfle, line 90: This isn't really `type-check`, its running `tsc` which will compile.

If you only want type checking without emitting js output, it should be `tsc --noEmit`

src/Lexer.ts
```diff
@@ -1,12 +1,13 @@
-import { Tokenizer } from './Tokenizer.js';
-import { defaults } from './defaults.js';
-import { block, inline } from './rules.js';
+import { _Tokenizer } from './Tokenizer.ts';
```
styfle, line 1: I thought the `.ts` extension wasn't supported and it had to use `.js` (like you are doing in the Instance.ts file already)

Is this a tsup feature?

DreierF: It's a typescript feature [`allowImportingTsExtensions`](https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions) and esbuild, which is used for the actual transpilation in tsup, also supports it during bundling https://esbuild.github.io/content-types/#default-interop

styfle: I see. Any reason why the `Instance.ts` file is using `.js` imports?

DreierF: I just missed to adapt them during the rebase. Fixed now

src/Tokens.ts
```diff
@@ -0,0 +1,199 @@
...
+    export interface Space {
+        type: 'space';
+        raw: string;
+    }
+
+    export interface Code {
+        type: 'code';
+        raw: string;
+        codeBlockStyle?: 'indented' | undefined;
+        lang?: string | undefined;
```
calculuschild, line 34: My understanding is the question mark already identifies this property as optional.

```suggestion
        codeBlockStyle?: 'indented';
        lang?: string;
```

Is this equivalent? If so, I would prefer this simplified notation rather than mixing both the `?` and ` | undefined`

src/Instance.ts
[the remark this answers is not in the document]
calculuschild: I don't have a solution, but I'll second that I don't like the underscores.

src/Tokens.ts
DreierF: The mark markers the property as options, so you can omit it when declaring an object literal, which results in undefined if you try to access it, while the explicit undefined type allows to explicitly assign the value undefined to the property which happens for example in the defaults.

src/MarkedOptions.ts
```diff
@@ -0,0 +1,218 @@
...
+export interface SluggerOptions {
+  /** Generates the next unique slug without updating the internal accumulator. */
+  dryrun?: boolean;
+}
+
+export type Rule = RegExp | string;
+
+export interface Rules {
+  [ruleName: string]: Pick<RegExp, 'exec'> | Rule | Rules;
+}
```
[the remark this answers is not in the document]
DreierF: Because e.g. `block.normal`, `block.gfm` etc. is a Rules object nested within the block rules object. Therefore the recursion.

I moved the `Rule` and `Rules` types to `rules.ts`

src/Tokenizer.ts
```diff
@@ -586,10 +600,10 @@ export class Tokenizer {
     }
   }
 
-  reflink(src, links) {
+  reflink(src: string, links: Record<string, Pick<Tokens.Link | Tokens.Image, 'href' | 'title'>>): Tokens.Link | Tokens.Image | Tokens.Text | undefined {
```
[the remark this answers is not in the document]
DreierF: I extracted the type into a separate type (still using `Pick`). I thought it would be better because it ensures the types stay in sync and will yell at you when you make changes to `Tokens.Image` that are incompatible with what reflink expects. But if you'd prefer I can also make it its own type.

src/rules.ts
```diff
@@ -1,12 +1,68 @@
...
+
+type InlineSubRuleNames = 'gfm' | 'emStrong' | 'normal' | 'pedantic'| 'breaks';
 
 /**
  * Block-Level Grammar
  */
-export const block = {
+// Not all rules are defined in the object literal
+// @ts-expect-error
+export const block: Record<BlockRuleNames, Rule> & Record<BlockSubRuleNames, Rules> & Rules = {
```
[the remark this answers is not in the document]
DreierF: The part in the block object is no problem and would work nicely with this type, but the code below the initial object definition would need a typecast on every usage where the rules are passed into `edit`, `replace` or are spread into subrules.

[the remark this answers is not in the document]
calculuschild: What is the error that happens if we don't typecast?

src/Tokenizer.ts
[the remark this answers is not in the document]
calculuschild: I would personally prefer typing all the different Tokens in the same way, in the same place, for the sake of future maintainability.

Yes, the two types wouldn't "sync", but at least we don't have this one type definition hiding in here where all the others are together in a different file.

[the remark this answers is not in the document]
UziTech: I think the token types should stay consistent across different tokenizers so when someone uses walktokens and sees `token.type === "image"` they don't have to care if that image token was created by a reflink or image tokenizer.

If we want the reflink token to be different from image we should have a "reflink" type.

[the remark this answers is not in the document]
calculuschild: @Uzitech Here I'm talking about the typescript type definition, not the Token.type object property.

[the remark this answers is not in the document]
UziTech: Are you talking about the return type of reflink (`Tokens.Link | Tokens.Image | Tokens.Text | undefined`) or the type of the links param (`Record<string, Pick<Tokens.Link | Tokens.Image, 'href' | 'title'>>`)?

src/rules.ts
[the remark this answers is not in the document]
DreierF: There are two categories of failures:
The first one for
```
block.normal = { ...block };
```
it would complain with 
```
TS2322: Type  { [x: string]: Rule; }  is not assignable to type  Rule 
Type  { [x: string]: Rule; }  is missing the following properties from type  RegExp :  exec, test, source, global , and  14  more.
```

The second category concerns 
```
block.gfm.table = edit(block.gfm.table as Rule)
```
and similar. It complains with 
```
TS2339: Property  table  does not exist on type  Rule 
Property  table  does not exist on type  string 
```

src/Tokenizer.ts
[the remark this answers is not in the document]
calculuschild: I'm talking about `Record<string, Pick<Tokens.Link | Tokens.Image, 'href' | 'title'>>` on the links param. Perhaps I misunderstood what was going on. But it just seemed weird to create a type on the fly here where all our other Token-related types are in `src/Tokens.ts`.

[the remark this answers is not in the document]
DreierF: Like so? https://github.com/markedjs/marked/pull/2805/commits/48db511a35da04ea5d593d5767b385df7d2b91c7

UziTech (approved):
This looks like a good start. We can fix the anys and ignores in future PRs.

tsconfig-type-test.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "target": "es2022",
```
styfle, line 3:
```suggestion
    "target": "es2021",
```

https://github.com/markedjs/marked/pull/2886

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "target": "es2022",
```
styfle, line 3:
```suggestion
    "target": "es2021",
```

https://github.com/markedjs/marked/pull/2886

UziTech:
@calculuschild do you have any other concerns? I would like to get this merged soon.

calculuschild:
I guess we can merge. I feel like there are probably some more tweaks we can make to clear up the typing syntax in a few places but I think that can be pushed to a later PR.

calculuschild (approved):

UziTech:
Ya I think we should merge this and get a couple other breaking changes in before releasing v6.

@DreierF if you can rebase this I can merge it.

DreierF:
Great to hear! Rebase is done ✅

PR #2820, opened 2023-05-29
fix: Replace custom `repeatString` function with `repeat()`

DreierF:
**Marked version:**

5.0.3

## Description

While working on #2805 I noticed that there are a few `repeatString` functions that could trivially be replaced with [`String.proptotype.repeat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat), which is already used in some other places in the code anyway.

## Contributor

  - New em_list_links
  - New em_and_reflinks
  - CommonMark Emphasis and strong emphasis
  - CommonMark Links
  - CommonMark HTML blocks
  - Original markdown_documentation_basics
  - Original markdown_documentation_syntax
  - GFM HTML blocks
  - GFM Emphasis and strong emphasis
  - GFM Links

## Committer


UziTech (approved):
Thanks for doing this!

joshbruce (approved):

styfle (approved):
Great, thanks!

davisjam:
Should be same-or-faster unless we're repeating really small strings, in which case the implementation shouldn't matter anyway. But it's nice to use the built-in language feature.

Do all marked-supported language runtimes now support `repeat`?

styfle:
We don't have a list of supported runtimes outside of Node.js as far I know.

We support Node.js 18 and newer since Marked 5.0.0 (see https://github.com/markedjs/marked/pull/2767)

PR #2831, opened 2023-06-05
feat: add Marked instance

### LINKED ISSUE #907: Setting options interferes with other "require"-calls to marked.

UziTech:
## Description

Add `import { Marked } from 'marked'` to get an instance of marked that will not change global options when options are changed or extensions are added.

```js
import { Marked } from 'marked';
const marked = new Marked(options); // or new marked.Instance(options)
marked.use(extension);

marked.parse(markdown);
```

This initial parameters are passed to `use` so you can add options or extensions.

```js
const marked = new Marked(
  options,
  extension1,
  extension2
)

// same as

const marked = new Marked();
marked.use(
  options,
  extension1,
  extension2
);
```

`import { marked } from 'marked'` still works the way it always has by updating global options.

- Fixes #907 

## Contributor

## Committer

### DIFF
--- a/docs/USING_ADVANCED.md
+++ b/docs/USING_ADVANCED.md
@@ -1,8 +1,8 @@
 ## Marked instance
 
-By default marked stores options and extensions in the global scope. In node js when you import marked into two scripts and change the options in one script the options will be changed in the other script.
+By default, Marked stores options and extensions in the global scope. That means changing the options in one script will also change the options in another script since they share the same instance.
 
-You can create an instance of marked to keep options and extensions locally scoped.
+If you don't want to mutate global scope, you can create a new instance of Marked to ensure options and extensions are locally scoped.
 
 ```js
 import { Marked } from 'marked';
--- a/docs/USING_ADVANCED.md
+++ b/docs/USING_ADVANCED.md
@@ -184,6 +184,7 @@ markedWorker.onmessage = (e) => {
 
 markedWorker.postMessage(markdownString);
 ```
+
 <h2 id="cli-extensions">CLI Extensions</h2>
 
 You can use extensions in the CLI by creating a new CLI that imports marked and the marked binary.
--- a/docs/USING_ADVANCED.md
+++ b/docs/USING_ADVANCED.md
@@ -9,9 +9,11 @@ import { Marked } from 'marked';
 const marked = new Marked([...useArgs]);
 ```
 
-```js
-import { marked } from 'marked';
-const myMarked = new marked.Instance([...useArgs]);
+```html
+<script src='path/to/marked.min.js'></script>
+<script>
+  const myMarked = new marked.Instance([...useArgs]);
+</script>
 ```
 
 |Argument                 |Type   |Notes                                                                  |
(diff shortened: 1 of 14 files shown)


docs/USING_ADVANCED.md
```diff
@@ -1,3 +1,23 @@
+## Marked instance
+
+By default marked stores options and extensions in the global scope. In node js when you import marked into two scripts and change the options in one script the options will be changed in the other script.
+
+You can create an instance of marked to keep options and extensions locally scoped.
```
styfle, line 5:
```suggestion
By default, Marked stores options and extensions in the global scope. That means changing the options in one script will also change the options in another script since they share the same instance.

If you don't want to mutate global scope, you can create a new instance of Marked to ensure options and extensions are locally scoped.
```

docs/USING_ADVANCED.md
```diff
@@ -1,3 +1,23 @@
...
+
+```js
+import { Marked } from 'marked';
+const marked = new Marked([...useArgs]);
```
styfle, line 9: Is this supposed to be an array? The tests make it seems like its an object, not array.

docs/USING_ADVANCED.md
```diff
@@ -1,3 +1,23 @@
...
+```
+
+```js
+import { marked } from 'marked';
+const myMarked = new marked.Instance([...useArgs]);
```
styfle, line 14: I'm not sure this usage is necessary. Is there a reason to use this instead of the one above?

calculuschild: I agree. If we provide multiple ways to do the same thing, that's just an additional point of confusion for users asking what the difference is.

calculuschild (approved):
I approve after @styfle 's points get addressed.

docs/USING_ADVANCED.md
UziTech: The constructor can take any number of options objects just like the use function. The square brackets mean it is optional. I know it is a weird way of writing that but that is how we indicate optional arguments in other parts of the docs.

docs/USING_ADVANCED.md
UziTech: I changed it to show a script tag so people know this is for client side javascript

docs/USING_ADVANCED.md
UziTech: I updated it to be more informative

docs/USING_ADVANCED.md
UziTech: Ya I agree we can remove this for now and add it later if someone has a valid use case.

styfle (approved):
Great, thanks! 🎉 

Perhaps we can move toward a future were we deprecate and remove the old `marked` that mutates global state in favor of this new class instance 👍

calculuschild (approved):

vorant94:
when is it expected to get 5.1.0 version of @types/marked to use instance-based marked in typescript?

PR #2838, opened 2023-06-05
fix: Fix more Escape conflicts

calculuschild:
**Marked version:** 5.0.4

**Markdown flavor:** GFM

## Description

- Simplifies the previous solution from #2627, and also fixes some missing edge cases mentioned therein: [comment](https://github.com/markedjs/marked/pull/2627/files#r1010018940)
## What was attempted

Literally the only cause of #2280 failing was because `\` was not included in the punctuation list. I added that, and undid the complex solution from #2627.

Also changed the "mask out escaped EmStrong" code to mask out *all* escaped tokens, which handled the new test case `_\\_` where the backslash is itself escaped. Could have been handled by a more complex escapedEmStrong regex, but this solution I think can apply to a few other locations where escapes come up, in a future cleanup PR, and doesn't appear to add any slowdown compared to the current solution.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -347,10 +347,9 @@ export class Lexer {
       maskedSrc = maskedSrc.slice(0, match.index) + '[' + 'a'.repeat(match[0].length - 2) + ']' + maskedSrc.slice(this.tokenizer.rules.inline.blockSkip.lastIndex);
     }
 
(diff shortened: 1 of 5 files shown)


UziTech (approved):
According to recheck those regexps are safe. 👍

styfle (approved):
Thanks!

PR #2841, opened 2023-06-07
fix: Simplify unicode punctuation

calculuschild:
**Marked version:** 5.0.5

## Description

Cleans up the unicode punctuation from #2811 by using `\p{P}` instead of a long list of unicode characters. There are a handful of punctuation characters ``$+<=>`^|~`` not included in that set for whatever reason, so they are still specified here. Includes the accompanying tweaks to a couple other regexes to apply it correctly. This also lets us cover a slightly larger punctuation set since my understanding is JS unicode symbols end at `\uFFFF` but there are a few more after that.

And a tiny unrelated logic simplification in the emStrong Tokenizer.

My only question is if there is a better way to exclude single characters from `\p{P}`, for instance in the emStrong, we don't include the current delimiter `*` or `_` in the punctuation checks. I get around this now with an additional lookahead regex:

Something like `(?!_)\p{P}`

For instance, this example lets us exclude `_` from the `\p{P}` group. I'm ok with this, but if there is some secret "subtraction from a unicode set" syntax, I would like to know.

I didn't add tests, but could potentially look up some of the characters that were missing previously and add them to the existing unicode test.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -348,8 +348,8 @@ export class Lexer {
     }
 
     // Mask out escaped characters
-    while ((match = this.tokenizer.rules.inline.escapedPunct.exec(maskedSrc)) != null) {
-      maskedSrc = maskedSrc.slice(0, match.index) + '++' + maskedSrc.slice(this.tokenizer.rules.inline.escapedPunct.lastIndex);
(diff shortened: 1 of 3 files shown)


UziTech (approved):
LGTM. I think the current tests are enough to tell if this is working as expected.

styfle (approved):

PR #2886, opened 2023-07-22
fix: Add support for Node 16

benmccann:
**Marked version:** `master`

**Markdown flavor:** n/a

## Description

- Fixes https://github.com/markedjs/marked/pull/2767#issuecomment-1534262376

Node 16 is still supported until September and there's no code in `marked` that doesn't work on Node 16, so it's unnecessarily aggressive to disallow it.

## Expectation

Should be able to run marked on Node 16

## Result

`engines` field gets in the way prior to this PR

## Contributor

## Committer

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         # lowest verison here should also be in `engines` field
-        node_version: [18, 'lts/*', '*']
+        node_version: [16, 18, 'lts/*', '*']
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
--- a/package.json
+++ b/package.json
@@ -89,6 +89,6 @@
     "minify": "uglifyjs lib/marked.umd.js -cm  --comments /Copyright/ -o marked.min.js"
   },
   "engines": {
-    "node": ">= 18"
+    "node": ">= 16"
   }
 }


styfle (approved):
I’m surprised tests are passing. We probably need to bump the target output to truly drop node 16 come September https://nodejs.org/gl/blog/announcements/nodejs16-eol

UziTech:
Our [https://github.com/markedjs/marked#compatibility](docs) state that we support lts and latest node. The engines field should reflect what versions we support not what versions happen to work.

benmccann:
Well Node 16 is an LTS release: https://nodejs.dev/en/about/releases/. And in any case, we could easily change the language on https://github.com/markedjs/marked#compatibility to cover Node 16. It's pretty frustrating that this package works on Node 16 without issue and I'm being banned from running it there anyway. I don't quite understand the reasoning behind that or what the benefit is. And it seems like a lot of other people are frustrated too: https://github.com/markedjs/marked/pull/2767#issuecomment-1534262376

UziTech:
The reasoning is that none of us on the markedjs team want to debug issues on older versions of node. We have limited free time and would like to focus that time on versions of node we use. If you would like to join the team and help maintain older versions you are welcome to. We do appreciate the work you have done on marked 😁👍.

You can still use it. There are many well documented ways to get around the engines version. You can use `--ignore-engines` or create your own npm package that just depends on marked and uses babel to ensure a certain es output.

benmccann:
Sure, if there are any issues that arise on the oldest version of Node, but not more recent versions, I don't mind being responsible for those

styfle:
Sounds easy enough - once 16.x added to CI then the issues will present themselves if a PR causes 16.x to regress 👍

yoyo837:
Can this PR solve the problem mentioned in https://github.com/markedjs/marked/pull/2767#issuecomment-1532417540? `Semantic release requires node v18`.

yoyo837:
@UziTech Can you help check this?

UziTech (approved):

benmccann:
thank you!!

PR #2891, opened 2023-07-29
fix: noImplicitAny: true

UziTech:
## Description

Remove `noImplicitAny: false` from tsconfig

## Contributor

## Committer


lib/marked.cjs
```diff
@@ -2024,23 +2027,27 @@ var Marked = class {
       if (pack.renderer) {
         const renderer = this.defaults.renderer || new _Renderer(this.defaults);
         for (const prop in pack.renderer) {
-          const prevRenderer = renderer[prop];
-          renderer[prop] = (...args2) => {
-            let ret = pack.renderer[prop].apply(renderer, args2);
+          const rendererFunc = pack.renderer[prop];
+          const rendererKey = prop;
```
styfle, line 2031: Is the variable for rendererKey needed? Can you use prop instead?

UziTech: This is so Typescript knows what type prop is. Without this it is a string and renderer[prop] is `any` (implicitly) This is a generated file. in the .ts file it is `const rendererKey = prop as  keyof _Renderer;`

styfle: Oops, I keep looking at the wrong files.

I think there is a way to tell github which files are generated so they start out collapsed in code reviews 🕵️‍♂️

styfle: Added in PR #2908

src/Hooks.ts
```diff
@@ -16,14 +16,14 @@ export class _Hooks {
   /**
    * Process markdown before marked
    */
-  preprocess(markdown: string) {
+  preprocess(markdown: string): string | Promise<string> {
```
styfle, line 19: Why `Promise<string>` here if its always returning a string?

UziTech: good catch! This was an early change

src/Instance.ts
```diff
@@ -15,6 +15,9 @@ import type { Token, TokensList } from './Tokens.ts';
 
 export type ResultCallback = (error: Error | null, parseResult?: string) => undefined | void;
 
+type unknownFunction = (...args: unknown[]) => unknown;
```
styfle, line 18:
```suggestion
type UnknownFunction = (...args: unknown[]) => unknown;
```

Lets use the same convention for all types (PascalCase)

src/Instance.ts
```diff
@@ -140,25 +144,29 @@ export class Marked {
...
+            return ret || '';
           };
         }
         opts.renderer = renderer;
       }
       if (pack.tokenizer) {
         const tokenizer = this.defaults.tokenizer || new _Tokenizer(this.defaults);
         for (const prop in pack.tokenizer) {
-          const prevTokenizer = tokenizer[prop];
+          const tokenizerFunc = pack.tokenizer[prop as keyof MarkedExtension['tokenizer']] as unknownFunction;
```
styfle, line 164:
```suggestion
          const tokenizerFunc = pack.tokenizer[prop as keyof MarkedExtension['tokenizer']] as UnknownFunction;
```

src/MarkedOptions.ts
```diff
@@ -98,8 +100,8 @@ export interface MarkedExtension {
    * postprocess is called to process html after marked has finished parsing.
    */
   hooks?: {
-    preprocess: (markdown: string) => string,
-    postprocess: (html: string | undefined) => string | undefined,
+    preprocess: (markdown: string) => string | Promise<string>,
+    postprocess: (html: string) => string | Promise<string>,
```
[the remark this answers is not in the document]
UziTech: The hooks can return promises if async option is true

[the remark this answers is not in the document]
styfle: Hmm, then we might need to overload the types in order to avoid all the `as string` in `src/Instance.ts`

styfle (approved):

UziTech:
@calculuschild any issue with merging this PR?

calculuschild (approved):
This is fine.

PR #2893, opened 2023-07-30
fix: fix more types

UziTech:
## Description

Make types more strict

## Contributor

## Committer


src/Tokenizer.ts
```diff
@@ -69,6 +69,7 @@ function indentCodeCompensation(raw: string, text: string) {
  */
 export class _Tokenizer {
   options: MarkedOptions;
+  // TOOD: Fix this rules type
```
styfle, line 72:
```suggestion
  // TODO: Fix this rules type
```

src/MarkedOptions.ts
```diff
@@ -169,46 +173,45 @@ export interface MarkedExtension {
...
+  walkTokens?: ((token: Token) => void | unknown | Promise<void>) | undefined | null;
   /**
    * Generate closing slash for self-closing tags (<br/> instead of <br>)
    * @deprecated Deprecated in v5.0.0 use marked-xhtml to emit self-closing HTML tags for void elements (<br/>, <img/>, etc.) with a "/" as required by XHTML.
    */
   xhtml?: boolean | undefined;
 }
 
-export interface MarkedOptions extends Omit<MarkedExtension, 'extensions' | 'renderer' | 'tokenizer' | 'walkTokens'> {
+export interface MarkedOptions extends Omit<MarkedExtension, 'renderer' | 'tokenizer' | 'extensions' | 'walkTokens'> {
```
styfle, line 184: Any reason the order changed here? Seems like it will be the same type regardless of order?

src/Instance.ts
```diff
@@ -63,13 +63,13 @@ export class Marked {
           break;
         }
         default: {
-          if (this.defaults.extensions && this.defaults.extensions.childTokens && this.defaults.extensions.childTokens[token.type]) { // Walk any extensions
-            this.defaults.extensions.childTokens[token.type].forEach((childTokens) => {
-              // @ts-expect-error we assume token[childToken] is an array of tokens but we can't be sure
-              values = values.concat(this.walkTokens(token[childTokens], callback));
+          const genericToken = token as Tokens.Generic;
+          if (this.defaults.extensions && this.defaults.extensions.childTokens && this.defaults.extensions.childTokens[genericToken.type]) { // Walk any extensions
+            this.defaults.extensions.childTokens[genericToken.type].forEach((childTokens) => {
```
styfle, line 68:
```suggestion
          const array = this.defaults.extensions?.childTokens?.[genericToken.type] ?? []
          // Walk any extensions
          array.forEach((childTokens) => {
```

I think you can simplify this a bit

calculuschild: I think we originally had something similar but there was some issue with the optional chaining / nullish coalescing syntax at the time. Is that still the case?

styfle: We use typescript now with `target` configured, so it should downlevel automatically (if necessary)

styfle: And I realize now we support node 16 or newer so it already natively supports those features, no downleveling needed

https://node.green

src/MarkedOptions.ts
UziTech: Not sure, I copied a few types from @types/marked and that might be how they were arranged. I don't think it matters either way.

src/Instance.ts
```diff
@@ -113,7 +113,7 @@ export class Marked {
               throw new Error("extension level must be 'block' or 'inline'");
             }
             if (extensions[ext.level]) {
```
styfle, line 115:
```suggestion
            const arr = extensions[ext.level];
            if (arr) {
```

src/Instance.ts
```diff
@@ -113,7 +113,7 @@ export class Marked {
...
-              extensions[ext.level].unshift(ext.tokenizer);
+              extensions[ext.level]!.unshift(ext.tokenizer);
```
styfle, line 116:
```suggestion
              arr.unshift(ext.tokenizer);
```

src/Parser.ts
```diff
@@ -45,35 +45,16 @@ export class _Parser {
...
-      checked,
-      task,
-      checkbox,
-      ret;
+    let out = '';
 
-    const l = tokens.length;
-    for (i = 0; i < l; i++) {
-      token = tokens[i];
+    for (let i = 0; i < tokens.length; i++) {
```
styfle, line 50: Is there any perf difference between storing `tokens.length` in a variable vs checking in here each time? I suspect we won't see much of a different if the length is small but it might add up for large arrays.

UziTech: There is no performance benefit from storing the variable first. https://stackoverflow.com/a/17995000/806777

styfle (approved):
Thanks! 🥇

src/MarkedOptions.ts
```diff
@@ -169,46 +173,45 @@ export interface MarkedExtension {
...
-    | (TokenizerAndRendererExtension[] & {
-    renderers: Record<string, RendererExtensionFunction>,
-    childTokens: Record<string, string[]>,
-    block: any[],
-    inline: any[],
-    startBlock: Array<(this: TokenizerThis, src: string) => number | void>,
-    startInline: Array<(this: TokenizerThis, src: string) => number | void>
-  })
-    | undefined | null;
+  walkTokens?: null | ((token: Token) => void | (unknown | Promise<void>)[]);
```
krassowski, line 216: This breaks downstream compilation because `(unknown | Promise<void>)[]` gets collapsed to `unknown[]` https://github.com/jupyterlab/jupyterlab/issues/15020 see https://stackoverflow.com/questions/61685819/typescript-does-not-recognize-a-union-of-type-unknown-and-type-promiseunknown

Should we adapt the type donwstream, or do you consider it a bug?

styfle: @UziTech Perhaps this should be `(void | Promise<void>)[]` instead?

UziTech: Ya I think `(void | Promise<void>)[]` should work

UziTech: The only time we care about the return value is if `async` is true. Than we await the return value. I think the problem was that we can't await `void` in typescript but it is perfectly fine in JavaScript.

styfle: > The only time we care about the return value is if async is true.

Maybe we need different types when `async: true` if we want it to be more precise

UziTech: Should be fixed by #2955

PR #2916, opened 2023-08-03
fix: use rollup to output js files

### LINKED ISSUE #2913: 6.0 broke the build
**Marked version:**

6.0.0

**Describe the bug**
We've been using Marked as a CDN resource for multiple demo pages we create. With the 6.0 release, these pages stopped working. 

I am not sure if there is a bug or if there is a new config consideration that I am not seeing in the docs. 

```
Uncaught (in promise) ReferenceError: marked is not defined
```

**To Reproduce**
You can see how we use Marked in this example. 
https://github.com/AlaskaAirlines/auro-alert/blob/9188f2b4cbf0f319e3ce59bfa326611ecc3c0b5e/demo/index.html#L19-L29

Again, setting the version to 5.1.0, this code renders the markdown as expected. 

```js
    <script type="module">
      import 'https://unpkg.com/marked@5.1.0/marked.min.js';
      import 'https://unpkg.com/prismjs@latest/prism.js';
      fetch('/demo/demo.md')
        .then((response) => response.text())
        .then((text) => {
          const rawHtml = marked.parse(text);
          document.querySelector('main').innerHTML = rawHtml;
          Prism.highlightAll();
        });
    </script>
```

**Expected behavior**

The markdown renders as HTML as expected.

UziTech:
**Marked version:** v6.0.0

## Description

Fix umd output by removing tsup and adding rollup

Add umd-test to check umd adds global marked

- Fixes #2913 

## Contributor

## Committer


calculuschild (approved):
Lgtm

UziTech:
@styfle any issue with merging this?

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2021",
+    "target": "es2022",
```
styfle, line 3: I believe we chose ES2021 to support Node.js 16

package.json
```diff
@@ -67,31 +68,32 @@
...
     "test:specs": "npm test -- test/specs/**/*-spec.js",
     "test:lint": "eslint .",
     "test:redos": "node test/recheck.js > vuln.js",
+    "test:types": "tsc --project tsconfig-type-test.json",
+    "test:umd": "node test/umd-test.js",
     "test:update": "node test/update-specs.js",
     "rules": "node test/rules.js",
     "bench": "npm run build && node test/bench.js",
     "lint": "eslint --fix .",
-    "type-check": "tsc",
```
styfle, line 90: Maybe I'm missing it, but where do we run `tsc` using the root `tsconfig.json`? Did you verify that a typo will fail the build?

tsconfig.json
UziTech: Es 2021 doesn't support private methods that we already use. This still passes with node 16 with 2022.

package.json
UziTech: Yes rollup handles running tsc and will fail to build on typescript error

styfle (approved):

PR #2928, opened 2023-08-09
docs: update docs

UziTech:
## Description

Update docs

- Update highlight js css
- Use markedHighlight extension
- Add Instance link
- Fix headers with headerIds: false
- Import esm version in worker
- Remove unfetch and ESPromise dependencies
- Remove non-worker demo parser
- Animate response time when markdown updates

## Contributor

## Committer

### DIFF
--- a/build-docs.js
+++ b/build-docs.js
@@ -1,6 +1,7 @@
 import { promises } from 'fs';
 import { join, dirname, parse, format } from 'path';
-import { parse as marked } from './lib/marked.esm.js';
+import { Marked } from './lib/marked.esm.js';
+import { markedHighlight } from 'marked-highlight';
 import { HighlightJS } from 'highlight.js';
 import titleize from 'titleize';
 
@@ -12,6 +13,12 @@ const outputDir = join(cwd, 'public');
 const templateFile = join(inputDir, '_document.html');
 const isUppercase = str => /[A-Z_]+/.test(str);
 const getTitle = str => str === 'INDEX' ? '' : titleize(str.replace(/_/g, ' ')) + ' - ';
+const marked = new Marked(markedHighlight((code, language) => {
+  if (!language) {
+    return highlightAuto(code).value;
+  }
+  return highlight(code, { language }).value;
+}));
 
 async function init() {
   console.log('Cleaning up output directory ' + outputDir);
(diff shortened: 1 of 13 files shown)


calculuschild:
Looks fine to me. Only comment is the response time seems to be way off. I.e., if I  paste in a huge block, the output panel will time out saying (took longer than 1 second... 2 seconds.... 3 seconds) and eventually render, then the Response Time happily animates to say "177ms!".

calculuschild (approved):

UziTech:
The response time is the time it took marked to parse the markdown. The time it takes to send the payload to and from the worker is not included.

build-docs.js
```diff
@@ -38,14 +45,7 @@ async function build(currentDir, tmpl) {
       let buffer = await readFile(filename);
```
styfle, line 45: I think we should add `utf8` here to work with the string instead of the buffer..

That can be a future PR 😄

docs/CONTRIBUTING.md
```diff
@@ -10,14 +10,14 @@
 - [ ] Run `npm run build:reset` to remove changes to compiled files.
 - [ ] Submit a Pull Request.
 
-## Design principles
+<h2 id="design-principles">Design principles</h2>
```
styfle, line 13: Should we enable header ids instead of doing this manually?

docs/demo/worker.js
```diff
@@ -121,22 +112,40 @@ function jsonString(input, level) {
...
+        if (!marked) {
+          throw Error('No marked');
+        } else if (marked.marked) {
+          versionCache[ver] = marked.marked;
+        } else if (marked.default) {
+          versionCache[ver] = marked.default;
+        } else if (marked.parse) {
+          versionCache[ver] = marked;
+        } else {
+          throw new Error('Can not find marked');
```
styfle, line 141:
```suggestion
          throw new Error('Cannot find marked');
```

https://www.merriam-webster.com/grammar/cannot-vs-can-not-is-there-a-difference

docs/CONTRIBUTING.md
UziTech: We are already doing it manually in most of our pages so the ids don't get too long. I think we should keep it consistent and we don't want to change it for existing ids otherwise urls will break.

PR #2937, opened 2023-08-13
add config option to bin/marked

### LINKED ISSUE #2617: Can't use extensions from CLI
**Describe the feature**
I would like the ability to use extensions from the CLI version of the marked. It might already be possible, but there is no documentation to this effect in the CLI help.

Extensions can be used in the other ways of calling marked. They should also be usable from the CLI.

N/A

### LINKED ISSUE #2810: How to set "mangle" option if installed globally?
using marked as a globally installed module gives you a warning that asks you to set mangle option to false.

How do you do that? Is there a config file to change such options?

Sorry, if it's a noob question...

UziTech:
**Marked version:** 7.0.5

**Markdown flavor:** Marked CLI

## Description

Load config with marked cli. This will allow extensions to be used with the marked cli.

- Fixes #2810
- Fixes #2617

## TODO

- [x] docs
- [x] tests

## Contributor

## Committer


bin/main.js
```diff
@@ -0,0 +1,270 @@
...
+        throw Error(`Cannot load config file '${config}'`);
+      }
+
+      await runConfig(config);
+    } else {
+      const defaultConfig = [
+        '~/.marked.json',
+        '~/.marked.js',
+        '~/.marked/index.js'
+      ];
```
styfle, line 204: Do we need to consider `.mjs` and `.cjs` here too?

It might be better to ship this with only `.json` first and see if JS support is actually needed.

UziTech: I think the `.js` support will be more useful to be able to include extensions. `.mjs` and `.cjs` works for the config option. I don't think we need to include them by default.

styfle (approved):

PR #2879, opened 2023-07-17
revert 2841

UziTech:
**Marked version:** 5.1.1

## Description

Revert #2841

- Fixes #2865

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR); or,
- [ ] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] CI is green (no forced merge required).
- [ ] Squash and Merge PR following [conventional commit guidelines](https://www.conventionalcommits.org/).

PR #2942, opened 2023-08-14
Fix unicode Regex miscounting emoji length

calculuschild:
## Description

- Fixes #2841
- Supersedes #2879

Many emojis are 2+ unicode chars long. The \u tag which allows searching for punctuation also counts emojis as single chars, which throws off character count when slicing strings. Spreading the strings into an array restores the correct character count. There is probably some overhead slowdown but not that I can detect.

Eventually the `\v` regex tag (in Node 20) can replace`\u` to get an accurate char count natively.

## Contributor

## Committer

### DIFF
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -625,7 +625,8 @@ export class _Tokenizer {
     const nextChar = match[1] || match[2] || '';
 
     if (!nextChar || !prevChar || this.rules.inline.punctuation.exec(prevChar)) {
-      const lLength = match[0].length - 1;
+      // unicode Regex counts emoji as 1 char; spread into array for proper count (used multiple times below)
+      const lLength = [...match[0]].length - 1;
       let rDelim, rLength, delimTotal = lLength, midDelimTotal = 0;
 
       const endReg = match[0][0] === '*' ? this.rules.inline.emStrong.rDelimAst : this.rules.inline.emStrong.rDelimUnd;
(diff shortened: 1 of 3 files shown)


src/Tokenizer.ts
```diff
@@ -625,7 +625,8 @@ export class _Tokenizer {
     const nextChar = match[1] || match[2] || '';
 
     if (!nextChar || !prevChar || this.rules.inline.punctuation.exec(prevChar)) {
-      const lLength = match[0].length - 1;
+      // unicode Regex counts emoji as 1 char; spread into array for proper count (used multiple times below)
```
styfle, line 628: Instead of the "used multiple times below" comment, can we refactor this into a function such as `countUnicodeChars(str: string)`?

calculuschild: For something that is single-line and only happens 3 times, I figured the speed overhead of wrapping it in a function wasn't worth it.

UziTech (approved):

src/Tokenizer.ts
styfle: Yeah and I guess the 3rd usage is different. Ok lets ship it <picture data-single-emoji=":shipit:" title=":shipit:">:shipit:</picture>

styfle (approved):

src/Tokenizer.ts
MathiasWP: Performance wise:

- **Creating a function**: so small you won't even notice it
- **Using the spread operator**: bad

UziTech: @MathiasWP do you know a better way to count Unicode characters in JavaScript?

PR #3017, opened 2023-10-02
fix: run spec tests with testutils

UziTech:
## Description

use @markedjs/testutils to run tests with node test runner

BREAKING CHANGE: test runner doesn't work in node v16

## Contributor

## Committer


test/bench.js
```diff
@@ -135,9 +143,13 @@ function parseArg(argv) {
...
         // e.g. -abc
-        argv = arg.substring(1).split('').map(ch => {
-          return `-${ch}`;
-        }).concat(argv);
+        argv = arg
+          .substring(1)
+          .split('')
+          .map((ch) => {
+            return `-${ch}`;
+          })
```
styfle, line 151:
```suggestion
          .map((ch) => `-${ch}`)
```

Nit pick: no need for `return` here

test/run-spec-tests.js
```diff
@@ -0,0 +1,49 @@
...
+  defaultMarkedOptions: { gfm: false, pedantic: false }
+});
+
+outputCompletionTable('GFM', tests.GFM);
+runTests({
+  tests: tests.GFM,
+  parse,
+  defaultMarkedOptions: { gfm: true, pedantic: false }
+});
+
```
styfle, line 34: Are you missing `outputCompletionTable` here?

test/unit/Hooks.test.js
```diff
@@ -1,7 +1,14 @@
-import { marked } from '../../src/marked.js';
+import { Marked } from '../../lib/marked.esm.js';
 import { timeout } from './utils.js';
+import { describe, it, beforeEach } from 'node:test';
+import assert from 'node:assert';
```
styfle, line 4: Its a matter of preference, but you could change this to `node:assert/strict` here and then `assert.strictEqual` becomes `assert.equal`. Or to make it more terse:

```
import { equal } from 'node:assert/strict';
```

package.json
```diff
@@ -62,9 +62,7 @@
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-n": "^16.1.0",
     "eslint-plugin-promise": "^6.1.1",
-    "front-matter": "^4.0.2",
     "highlight.js": "^11.8.0",
-    "jasmine": "^5.1.0",
```
calculuschild, line 67: What benefits does node:test providee over Jasmine? Not really a concern, but just curious why the switch.

test/run-spec-tests.js
```diff
@@ -0,0 +1,49 @@
+import { Marked } from '../lib/marked.esm.js';
```
calculuschild, line 1: So an extension package will just need to make their own version of this file to run spec tests, right?

I'm most interested in spec tests, but is there a similar capability for running unit tests?

test/unit/Hooks.test.js
UziTech: There are tests that can't use strict. I would rather keep it consistent.

package.json
UziTech: Mostly just so @markedjs/testutils doesn't require extensions to use jasmine. Also I wanted to try the new node test runner.

styfle: I would flip the question around - what benefits does jasmine provide over the built in test runner? 😄

Because jasmine is 3MB of code https://packagephobia.com/result?p=jasmine

test/run-spec-tests.js
UziTech: No outputCompletionTable is only for CommonMark and GFM. None of the other spec groups have shouldFail tests

test/run-spec-tests.js
UziTech: Ya, there is a [`getAllMarkedSpecTests()`](https://github.com/markedjs/testutils#getallmarkedspectests) method to get the marked specs.

UziTech: I'm not using that method in marked because we want to be able to run tests that are created in PRs.

UziTech: Here is an example PR for an extension to run all marked tests

 https://github.com/UziTech/marked-katex-extension/pull/255/files

styfle (approved):

PR #3025, opened 2023-10-03
feat: add `--no-clobber` option to CLI

### LINKED ISSUE #2012: Is it possible to add a noclobber option?
**Describe the feature**
so that any existing file is not overwritten

prevent accidentally overwriting the file

skoriop:
**Marked version:** c9ffa649a4671afec2f18fe11777693b9007ee6a

**Markdown flavor:** n/a

## Description

- Fixes #2012
 
If the `--no-clobber` / `-n` option is set, it throws an error if the output file already exists.

## Contributor

## Committer

### DIFF
--- a/bin/main.js
+++ b/bin/main.js
@@ -116,7 +116,7 @@ export async function main(nodeProcess) {
           config = argv.shift();
           break;
         case '-n':
-        case '--noclobber':
+        case '--no-clobber':
           noclobber = true;
           break;
         case '-h':
--- a/bin/main.js
+++ b/bin/main.js
@@ -62,6 +62,7 @@ export async function main(nodeProcess) {
     let tokens;
     let config;
     let opt;
+    let noclobber;
 
     function getArg() {
       let arg = argv.shift();
@@ -113,6 +114,10 @@ export async function main(nodeProcess) {
         case '-c':
         case '--config':
           config = argv.shift();
+          break;
+        case '-n':
+        case '--noclobber':
+          noclobber = true;
           break;
         case '-h':
         case '--help':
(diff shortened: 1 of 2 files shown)


UziTech (approved):
LGTM

styfle (approved):
Looks good. It would be nice to add a test for this one too 👍

bin/main.js
```diff
@@ -114,6 +115,10 @@ export async function main(nodeProcess) {
         case '--config':
           config = argv.shift();
           break;
+        case '-n':
+        case '--noclobber':
```
styfle, line 119: Should we call this `--no-clobber` instead?

UziTech: Good catch. `cp` has a [`--no-clobber`](https://man7.org/linux/man-pages/man1/cp.1.html) option. I think we should keep it consistent

UziTech (changes requested):
Please change the option to `--no-clobber` to be consistent with other commands. https://man7.org/linux/man-pages/man1/cp.1.html

bin/main.js
skoriop: Yup, fixed in b55144d0d403b64a93df4b6e06e7c0ae98c746c6

bin/main.js
```diff
@@ -216,6 +221,10 @@ export async function main(nodeProcess) {
       : await marked.parse(data, options);
 
     if (output) {
+      if (noclobber && await fileExists(output)) {
+        nodeProcess.stderr.write('marked: output file \'' + output + '\' already exists, disable the \'-n\' / \'--noclobber\' flag to overwrite\n');
```
styfle, line 225:
```suggestion
        nodeProcess.stderr.write('marked: output file \'' + output + '\' already exists, disable the \'-n\' / \'--no-clobber\' flag to overwrite\n');
```

styfle (approved):

UziTech (approved):

PR #3056, opened 2023-10-24
chore: Update marked-man@2.0.0

kapouer:
Which is compatible with marked@9

Very few changes besides that.


UziTech:
You will need to run`npm i` to update package.json

UziTech (approved):
Thanks! :100:

PR #3103, opened 2023-11-22
fix: fix marked types

### LINKED ISSUE #3101: marked.parseInline() returns string | Promise<string> even when { async: false }
**Marked version:** 10.0.0

**Describe the bug**
`marked.parseInline("markdown", { async: false })` returns a `string | Promise<string>`.

**Expected behavior**
`marked.parseInline("markdown", { async: false })` should return a `string`. This is already the case with `marked.parse("markdown", { async: false })`

UziTech:
**Marked version:** 10.0.0

## Description

Fix `marked` and `marked.parse` types. `async: false` could still return a `Promise<string>` if an extension sets `async: true`.

- Fixes #3101 

## Contributor

## Committer

### DIFF
--- a/src/marked.ts
+++ b/src/marked.ts
@@ -26,13 +26,13 @@ const markedInstance = new Marked();
 export function marked(src: string, options: MarkedOptions & { async: true }): Promise<string>;
 
 /**
- * Compiles markdown to HTML synchronously.
+ * Compiles markdown to HTML.
  *
  * @param src String of markdown source to be compiled
  * @param options Optional hash of options
- * @return String of compiled HTML
+ * @return String of compiled HTML. Wil be a Promise of string if async is set to true by any extensions.
  */
-export function marked(src: string, options?: MarkedOptions): string;
+export function marked(src: string, options?: MarkedOptions): string | Promise<string>;
 export function marked(src: string, opt?: MarkedOptions): string | Promise<string> {
   return markedInstance.parse(src, opt);
 }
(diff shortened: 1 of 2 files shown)


crystalfp:
Well, with version 11 I have to cast the result to string to have it works:
```
marked.parse(markdown, {gfm: true, async: false}) as string).replaceAll(...)
```
Never needed before.

UziTech:
Yes but it is actually correctly typed now. Before it would be typed as a string even though it could return a Promise<string>.

It would be great if typescript provided a way to return the type based on `marked.defaults.async`. But as far as I know typescript have a feature to do that.

crystalfp:
Ah, this explain everything and I'm happy with the cast solution.
Other possibility, instead of relying on "async" option, is to provide two different entry points like `fast-glob` does: `fglob(): Promise<string[]>` and `fglob.sync(): string[]`.
Anyway, thanks for clarifying the issue.
mario

SupremeVoid:
> Yes but it is actually correctly typed now. Before it would be typed as a string even though it could return a Promise.
> 
> It would be great if typescript provided a way to return the type based on `marked.defaults.async`. But as far as I know typescript doesn't have a feature to do that.

How about using a conditional type?
```
type ParseType<T> = T extends { async: true } ? Promise<string> : string;

function parse<T extends MarkedOptions>(src: string, options?: T): ParseType<T> {}

const string = parse('SomeSrc'); // this returns a string
const string2 = parse('SomeSrc', {async: false});  // this still returns a string
const string3 = parse('SomeSrc', {async: true});  // this still returns a Promise<string>
```

Issue #2639, opened 2022-11-03
Parser.parse extension to be able to change tokens as a whole


kapouer:
**Describe the feature**
Ability to get the "root" tokens list before rendering - either through a Parser.parse extension, or another mecanism.

**Why is this feature necessary?**
I need my extension to be able to prepend a default (depth 1) heading token when one is missing.

**Describe alternatives you've considered**
```js
const defaultParse = Parser.parse;
Parser.parse = (tokens, options) => {
	const [{ type, depth }] = tokens;
	if (type != "heading" || depth != 1) {
		tokens.unshift({
			type: 'heading',
			depth: 1,
			tokens: []
		});
	}
	return defaultParse.call(Parser, tokens, options);
};
```
but i hate monkey-patching.

calculuschild:
The [walkTokens](https://marked.js.org/using_pro#walk-tokens) function might be what you are looking for. It lets you modify the token tree before being sent to the final HTML renderer.

kapouer:
Yes, but (unless I'm mistaken) it's not possible to know if the token is the first token of the document.
I tried setting a flag, but that fails when using marked as a library.

UziTech:
If all you need are the root tokens you can get the array from `marked.lexer`

```js
const tokens = marked.lexer(markdown, options);
const [{ type, depth }] = tokens;
if (type != "heading" || depth != 1) {
  tokens.unshift({
    type: 'heading',
    depth: 1,
    tokens: []
  });
}
marked.parser(tokens, options);
```

There isn't currently an easy way to make this into an extension. For that we may need to implement [hooks](https://github.com/markedjs/marked/pull/1878).

PR #3114, opened 2023-11-30
feat: add processAllTokens hook

UziTech:
**Marked version:** 11.0.0

## Description

Add `tokens` hook that receives all tokens and can change the tokens.

This hook can allow extensions that change all tokens before walkTokens functions are called. For example an extension that adds the parent and sibling properties to tokens.

## Contributor

## Committer

### DIFF
--- a/src/Hooks.ts
+++ b/src/Hooks.ts
@@ -12,7 +12,7 @@ export class _Hooks {
   static passThroughHooks = new Set([
     'preprocess',
     'postprocess',
-    'tokens'
+    'processAllTokens'
   ]);
 
   /**
@@ -30,9 +30,9 @@ export class _Hooks {
   }
 
   /**
-   * Process tokens before walk tokens
+   * Process all tokens before walk tokens
    */
-  tokens(tokens: Token[] | TokensList) {
+  processAllTokens(tokens: Token[] | TokensList) {
     return tokens;
   }
 }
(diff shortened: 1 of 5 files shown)


calculuschild:
Ok, so this is a "middle Process" hook along the lines of Preprocess and Postprocess? I like that. Does this also give access to the `tokens.links` property so we can manipulate the reflink definitions? For example to inject a common link that can be referenced globally without being defined within the markdown input?

UziTech:
Yes it would. It is passed the `tokens` object that is returned by the lexer after all tokens have been created and before walkTokens functions are called.

calculuschild:
Adding to this then: could we have two hooks? one immediately after block tokens, and another immediately after inline tokens? I'm thinking of a case where you might want to alter the block-level tokens before their internal tokens get lexed.

UziTech:
A hook that captures block tokens before creating the inline tokens will need to take a bit more thought because of the way we do async marked.

The hook will have to be in the `Lexer.lex` method and have to return a promise if `options.async` is `true` but it currently doesn't return a promise. That might have to be a breaking change.

I think it could be done in another PR.

UziTech:
@calculuschild should I rename this to allTokens so we can add blockTokens later?

calculuschild:
Yeah, probably a good idea. I might suggest we use a verb  like we have with the other steps, i.e. "preprocess", "walktokens", "parse", etc.

- Preprocess
- Lex
- BlockTokens????
- AllTokens????
- WalkTokens
- Parse
- Postprocess

Maybe "ProcessTokens" or something?

Or like, ProcessBlockTokens, ProcessAllTokens.

Something like that?

UziTech:
Ya I think I like processAllTokens then we can do processBlockTokens later

calculuschild (approved):

kapouer:
For reference, this also solves #2639 and proves mind communication.

PR #3115, opened 2023-11-30
fix: fix hooks type

dansalias:
It shouldn't be required to define both a preprocess and postprocess hook.

## Description

At the moment both `preprocess` and `postprocess` must be defined when creating custom hooks, forcing the following to keep the TypeScript compiler happy:

```ts
marked.use({
  hooks: {
    preprocess: (md) => { /* my preprocess hook */ },
    postprocess: md => md, // dummy postprocess hook I have to define
  },
})
```

This PR makes both hooks optional.

### DIFF
--- a/src/MarkedOptions.ts
+++ b/src/MarkedOptions.ts
@@ -72,8 +72,8 @@ export interface MarkedExtension {
    * postprocess is called to process html after marked has finished parsing.
    */
   hooks?: {
-    preprocess: (markdown: string) => string | Promise<string>,
-    postprocess: (html: string) => string | Promise<string>,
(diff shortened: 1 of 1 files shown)


UziTech:
Could you actually change this to work similar to the renderer and tokenizer objects so we don't need to list each hook.

Something like https://github.com/markedjs/marked/pull/3114/files#diff-7a98348a48285a8aef2260392017b4306c7fcf0714da56f53ae780de81aac4e3

Also make sure to add type tests in https://github.com/markedjs/marked/blob/master/test/types/marked.ts

UziTech:
this should be fixed in v11.1.0

PR #3116, opened 2023-12-01
fix: marked and parse type overload with discrinated union options

Yimiprod:
**Marked version:** 11.0.0

## Description

Fixes typescript expected error which force to use `as string` or `as Promise` or `ts-expect-error` comments by using disciminated union for options.

Added two type guards to detect which options object is used.

## Expectation

- `as string` or `as Promise<string>` can be safely removed when using `marked()` and `marked.parse()` with respective config object
- No breaking change for library mutating default options

## Result

No regression, no behavior change

## Contributor

## Committer

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -8,7 +8,12 @@ import { _TextRenderer } from './TextRenderer.ts';
 import {
   escape
 } from './helpers.ts';
-import type { MarkedExtension, MarkedOptions } from './MarkedOptions.ts';
+import {
+  isAsyncOptions,
+  isSyncOptions,
+  type MarkedExtension,
+  type MarkedOptions
+} from './MarkedOptions.ts';
 import type { Token, Tokens, TokensList } from './Tokens.ts';
 
 export type MaybePromise = void | Promise<void>;
@@ -265,11 +270,11 @@ export class Marked {
 
   #parseMarkdown(lexer: (src: string, options?: MarkedOptions) => TokensList | Token[], parser: (tokens: Token[], options?: MarkedOptions) => string) {
     return (src: string, options?: MarkedOptions | undefined | null): string | Promise<string> => {
-      const origOpt = { ...options };
-      const opt = { ...this.defaults, ...origOpt };
+      const origOpt: MarkedOptions = { ...options };
+      const opt: MarkedOptions = { ...this.defaults, ...origOpt };
 
       // Show warning if an extension set async to true but the parse was called with async: false
-      if (this.defaults.async === true && origOpt.async === false) {
+      if (isAsyncOptions(this.defaults) && isSyncOptions(origOpt)) {
         if (!opt.silent) {
           console.warn('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.');
         }
@@ -292,7 +297,7 @@ export class Marked {
         opt.hooks.options = opt;
       }
 
-      if (opt.async) {
+      if (isAsyncOptions(opt)) {
         return Promise.resolve(opt.hooks ? opt.hooks.preprocess(src) : src)
           .then(src => lexer(src, opt))
           .then(tokens => opt.hooks ? opt.hooks.processAllTokens(tokens) : tokens)
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -273,16 +273,12 @@ export class Marked {
       const origOpt: MarkedOptions = { ...options };
       const opt: MarkedOptions = { ...this.defaults, ...origOpt };
 
-      // Show warning if an extension set async to true but the parse was called with async: false
-      if (isAsyncOptions(this.defaults) && isSyncOptions(origOpt)) {
-        if (!opt.silent) {
-          console.warn('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.');
-        }
-
-        opt.async = true;
-      }
-
       const throwError = this.#onError(!!opt.silent, !!opt.async);
+
+      if (isAsyncOptions(this.defaults) && isSyncOptions(origOpt)) {
+        // Throw an error if an extension set async to true but the parse was called with async: false
+        return throwError(new Error('marked(): The async option was set to true by an extension. Remove the async: false option to continue.'));
+      }
 
       // throw error in case of non string input
       if (typeof src === 'undefined' || src === null) {
--- a/src/MarkedOptions.ts
+++ b/src/MarkedOptions.ts
@@ -64,8 +64,8 @@ export interface MarkedExtension {
    * Add tokenizers and renderers to marked
    */
   extensions?:
-    | TokenizerAndRendererExtension[]
-    | undefined | null;
+  | TokenizerAndRendererExtension[]
+  | undefined | null;
 
   /**
    * Enable GitHub flavored markdown.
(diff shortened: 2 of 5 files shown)


src/Instance.ts
```diff
@@ -250,11 +255,11 @@ export class Marked {
...
-      const origOpt = { ...options };
-      const opt = { ...this.defaults, ...origOpt };
+      const origOpt: MarkedOptions = { ...options };
+      const opt: MarkedOptions = { ...this.defaults, ...origOpt };
 
       // Show warning if an extension set async to true but the parse was called with async: false
-      if (this.defaults.async === true && origOpt.async === false) {
+      if (isAsyncOptions(this.defaults) && isSyncOptions(origOpt)) {
         if (!opt.silent) {
           console.warn('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.');
```
Yimiprod, line 264: Question for maintainer: should this warning not be silent ?
Because as far as i know https://github.com/markedjs/marked/pull/3103 tried to "fix" the type to avoid unexepected Promises.

With this PR, the type returned by `marked()` and `marked.parse()` will have miss if a library change the default.async to true and user should be warned about it.

UziTech: The silent option is for when they want things to fail instead of get warnings. If they want to see the error they can just set the silent option to false.

UziTech:
Before #3103 and if we merge this PR the type retuned be marked.parse could be wrong. If you are ok with not knowing the correct return type, just don't use typescript.

What is the difference from a user adding `marked.parse(markdown, { async: false})` and ` marked.parse(markdown) as string`? I think the second one is better sync it is explicit what you are doing.

Yimiprod:
The second one will break at runtime if a library change the default config and dev not knowing it will don't understand why since they force casted a type to the output.

UziTech:
The first one will also break at runtime with this PR and before #3103

UziTech:
So I think the only solution to not break at runtime if the developer doesn't know if they added an async extension is to leave it the way it is and await marked.parse

UziTech:
If the developer does know that they didn't add an async extension than `as string` seems to be the cleanest solution to me.

Yimiprod:
True that, that's why suggested adding a breaking change in my previous comment, making the behavior more explicit:
If the developper explicitly pass `async: true` or `async: false` marked will behave like expected, returning Promise or string.
Else if nothing is set, return Promise or string, based on default.

I don't know what impact it will have for library using marked, but since it's a breaking change, this will change the major version of marked, this way preventing update for user prefixing their package.json version with at least `^` ?

src/Instance.ts
UziTech: So you are suggesting throwing an error here instead of logging a massage? I think we could do that. This would need to be a breaking change.

vzakharov:
Can you please merge this? Having .parse return `| Promise` when it definitely isn’t one is such a bugger.

UziTech:
If this was complete we could merge it in the next major version.

A couple of things that need to be fixed:

- throw an error instead of warn when `async: false` is used with an async extension.
- test `new Marked().parse(md, { async: true/false })` returns the correct type.
- fix merge conflicts

Yimiprod:
I'll try to find some time to fix it as soon as possible 👍

Yimiprod:
@UziTech added a new commit that replace the warning with a throw, and adding tests reflecting this changes.
The commit is flagged as breaking change (but i can remove it if you want).

src/Instance.ts
```diff
@@ -265,13 +270,13 @@ export class Marked {
...
+      const origOpt: MarkedOptions = { ...options };
+      const opt: MarkedOptions = { ...this.defaults, ...origOpt };
 
-      // Show warning if an extension set async to true but the parse was called with async: false
-      if (this.defaults.async === true && origOpt.async === false) {
+      if (isAsyncOptions(this.defaults) && isSyncOptions(origOpt)) {
+        // Throw an error if an extension set async to true but the parse was called with async: false
         if (!opt.silent) {
-          console.warn('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.');
+          throw new Error('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.');
```
UziTech, line 279: Can you use the `throwError` function instead of throwing the error. Similar to other errors in this function. You will have the move the creation of the function above this if statement.

test/types/marked.ts
```diff
@@ -246,21 +261,25 @@ marked.use(asyncExtension);
 const md = '# foobar';
 const asyncMarked: string = await marked(md, { async: true });
 const promiseMarked: Promise<string> = marked(md, { async: true });
-// @ts-expect-error marked can still be async if an extension sets `async: true`
-const notAsyncMarked: string = marked(md, { async: false });
-// @ts-expect-error marked can still be async if an extension sets `async: true`
+const notAsyncMarked: string = marked(md, { async: false, silent: true });
 const defaultMarked: string = marked(md);
-// as string can be used if no extensions set `async: true`
-const stringMarked: string = marked(md) as string;
```
UziTech, line 254: Marked calls that don't specify an async option should be `string | Promise<string>` since they could be either and they won't throw an error.

UziTech (changes requested):

test/types/marked.ts
vzakharov: > since they could be either

Just curious, how’s that? If no `async` is specified, it defaults to `false`, right?

UziTech: If no async is specified it should default to whatever async is set by extensions.

UziTech:
```js
const syncMarked = new Marked({ async: false });
syncMarked.parse(md) // this will be a string 

const asyncMarked = new Marked({ async: true });
asyncMarked.parse(md) // this will be a Promise<string>
```

Yimiprod: maybe there's a way to have typed returned instance ?
the default instance, without any modification, will be synch, right ?

UziTech: Maybe. Typescript would have to run your application to see what the return type is.

Yes default without any extensions is sync

src/Instance.ts
Yimiprod: Sorry for the hiatus, just fixed this point, for the second point i'll try to figure out how to do it, since it need to modify typing in the `.use()` function to output the correct instance type.

src/Instance.ts
```diff
@@ -265,20 +270,16 @@ export class Marked {
...
-      if (this.defaults.async === true && origOpt.async === false) {
-        if (!opt.silent) {
-          console.warn('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.');
-        }
+      const throwError = this.#onError(!!opt.silent, !!opt.async);
 
-        opt.async = true;
+      if (isAsyncOptions(this.defaults) && isSyncOptions(origOpt)) {
+        // Throw an error if an extension set async to true but the parse was called with async: false
+        return throwError(new Error('marked(): The async option was set to true by an extension. The async: false option sent to parse will be ignored.'));
```
UziTech, line 280: The error message should be:

```
marked(): The async option was set to true by an extension. Remove the async: false option to continue.
```

src/MarkedOptions.ts
```diff
@@ -150,3 +150,21 @@ export interface MarkedOptions extends Omit<MarkedExtension, 'hooks' | 'renderer
...
+}
+
+export type MarkedOptions = MarkedSyncOptions | MarkedAsyncOptions;
+
+export function isAsyncOptions(options: MarkedOptions): options is MarkedAsyncOptions {
+  return options.async === true;
+}
+
+export function isSyncOptions(options: MarkedOptions): options is MarkedSyncOptions {
+  return options.async === false;
```
UziTech, line 169: Should this be `false` or `undefined`?  Or maybe `!= true`?

UziTech (changes requested):

src/MarkedOptions.ts
Yimiprod: I pushed something like this:

```ts
return !isAsyncOptions(options);
```

It fail a lot of tests since it introduce a big breaking change.

Yimiprod:
Sorry for not having took the time to finish that merge request. I saw that it was resolved in another fix, thanks a lot to you and your library, it's so much easier to handle markdown content.

Issue #2355, opened 2022-01-12
Measured performance decrease between major releases of Marked, time for a performance pass?


alystair:
**What pain point are you perceiving?.**
Since MarkedJS's inception most major versions have increased time needed to complete parsing. The numbers shown below were captured on second page load /w 6x CPU slowdown - only Marked version being revised, nothing else. Marked processing time has increased more than 2x since v0.8.2

- Version 0.8.2
![image](https://user-images.githubusercontent.com/159109/149040782-3fe130ce-31e3-4c18-b7e3-850d08ee97a1.png)

- Version 2.0.3
![version 2.0.3](https://user-images.githubusercontent.com/159109/149039714-1805f3e5-54ed-41e4-a67a-49ebddadf2ff.png)
- Version 4.0.9
![version 4.0.9](https://user-images.githubusercontent.com/159109/149039645-a74a73b1-3468-45d0-8537-f1d5e2b79aa3.png)

**Describe the solution you'd like**
Perhaps it's time to look at an optimization pass and potential short circuits that could be taken to restore some previously lost performance? Performance should have remained somewhat consistent between versions.

UziTech:
It would be great to get something in the pipeline to measure performance. We tried that at one point but it wasn't reliable when run in GitHub actions.

PR are appreciated 😁👍

alystair:
Well, it's not something that has to be done often so this could be a manual operation by generating flamegraphs and attempting to optimize the functions that waste the most of time - just looking at the output above the lexer is a good place to start.

UziTech:
What are you using to produce those flame graphs?

alystair:
Chrome DevTools via "Performance" tab recordings

calculuschild:
Is there perhaps a good standard test Markdown document that would give Marked.js a thorough performance metric? I know we have our bench tests, but those use only our spec tests from Commonmark if I recall, which is heavily weighted toward specs with lots of odd edge cases (i.e., I think almost a fourth of our tests are for `em/strong`), and no tests at all for tables or other GFM features that might be contributing to slowdown as well.

alystair:
Safari also has some interesting tooling but DevTools should be sufficient. You can roll your own perf using the [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance) however it of course may add a few precious ns/ms to runtime ;)

UziTech:
Ya we could bench each test individually to see which ones are slowest relative to other packages.

[...]

PR #3146, opened 2023-12-28
refac: improve lexing inline elements step's performance

SpencerWhitehead7:
**Marked version:** 11.1.0

**Markdown flavor:** n/a

## Description

Improves performance of the lexing step, specifically lexing inline elements. Prompted by this issue https://github.com/markedjs/marked/issues/2355

## What was attempted

Create a benchmark file `bench-100.md` - this file copy pasted x 100 https://raw.githubusercontent.com/thephpleague/commonmark/2.2/tests/benchmark/sample.md (I saw that file linked in the issue and it looked good to me)

master:
- `npm run build`
- create a cpu profile with `node --cpu-prof bin/marked.js -i bench-100.md -o bench-100.html`

Apply my changes:
- `npm run build`
-  create an updated cpu profile with `node --cpu-prof bin/marked.js -i bench-100.md -o bench-100.html`

Load the two profiles in https://www.speedscope.app/ 
- before: 358ms doing marked stuff, 334ms lexing
- after: 242ms doing marked stuff, 218ms lexing
- the difference is even bigger with larger files
- I tried to attach the actual cpu profiles, but github wouldn't let me. I have screenshots of the speedscope though
before:
speedscope before
after:
speedscope after

The output file is identical, all the tests still pass, and as far as I can tell from reading the code, there shouldn't be any differences in behavior.

The reason this change speeds it up is because changing from `while loop + shift` to `for loop` is that `while + shift` is an n^2 operation; shift shuffles every remaining item in the array down by one, an n time operation, which is done once per item in the array (n * n). Running through the array with a for loop is just n time.

This optimization makes no difference in the automated benchmark. I haven’t read all the benchmark test cases, but I think they’re mostly tiny single item parses, which don’t accumulate a big queue to burn though. The benchmark file has lots of inline elements, so the optimization shows up strongly there.

I think a markdown file that had a good, statistically accurate representation of real world use patterns for all the different markdown features would be super helpful for benchmarking, although it seemed like overkill for this PR.

I’m not sure, and I don’t know if I have the time to do them if they do exist, but I think there might be other architecture level opportunities to improve performance, as opposed to micro level optimizations on things like the regexes. Is there interest in changes like that, even if they meant significant changes to the existing classes?

## Contributor

## Committer

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -93,6 +93,7 @@ export class _Lexer {
       const next = this.inlineQueue[i];
       this.inlineTokens(next.src, next.tokens);
     }
+    this.inlineQueue = [];
 
     return this.tokens;
   }
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -89,8 +89,8 @@ export class _Lexer {
 
     this.blockTokens(src, this.tokens);
 
-    let next;
-    while (next = this.inlineQueue.shift()) {
+    for (let i = 0; i < this.inlineQueue.length; i++) {
+      const next = this.inlineQueue[i];
       this.inlineTokens(next.src, next.tokens);
     }
 


calculuschild (approved):
Such a small change, but if it passes the tests this looks like an awesome improvement, even if it is mainly visible only on larger documents.

I think we are generally open to any performance boosts even if they require some retooling. As long as we remain compliant with the specs and extensions remain functional, we can always make a major version bump if signatures are significantly changed.

UziTech:
The only issue I can see is that `this.inlineQueue` will be an empty array at the end of the while loop and not at the end of the for loop. This could cause issues with some advanced uses of marked

SpencerWhitehead7:
As far as I could tell, nothing reads the array after the end of the while loop. I don't see Lexer using it anywhere else, or it being accessed elsewhere, and the lexer just returns the array of tokens at the end.

If there's cases I'm missing, it could be reset to an empty array after the for loop runs, or reversed and then popped instead of shifted.

UziTech:
I don't think it will break anything in marked if it isn't reset but if someone uses the lexer with `new marked.Lexer()` than it could cause issues when they reuse the same lexer.

I think we can just reset it to an empty array after the for loop

SpencerWhitehead7:
Oh, that makes sense. Thank you for explaining the case so clearly. I added reseting it to an empty array.

UziTech (approved):

PR #3165, opened 2024-01-18
Fix ENOENT error message in CLI

rzhade3:
**Marked version:** 11.1.1

**Markdown flavor:** n/a

## Description

This PR fixes a slight bug in the `marked` CLI in which the error message due to a `ENOENT` was bugging out the error output. 

I also updated the language in the `ENOENT` error to better reflect the fact that the `ENOENT` may have been due to any of the inputted files. 

## Expectation

When running the `marked` CLI with an input file that does not exist, I expect it to print out an error containing only the relevant information.

Screenshot 2024-01-18 at 1 58 05 AM

## Result

Marked was attempting to print out both error paths, and attempting to print an error object to `stderr`

Screenshot 2024-01-18 at 1 59 27 AM

## What was attempted

The stacktrace above was remediated.

## Contributor

## Committer

### DIFF
--- a/bin/main.js
+++ b/bin/main.js
@@ -270,7 +270,7 @@ export async function main(nodeProcess) {
     nodeProcess.exit(0);
   } catch (err) {
     if (err.code === 'ENOENT') {
-      nodeProcess.stderr.write('marked: ' + err.path + ': No such directory');
+      nodeProcess.stderr.write('marked: ' + err.path + ': No such file or directory');
     } else {
       nodeProcess.stderr.write(err.message);
     }
--- a/bin/main.js
+++ b/bin/main.js
@@ -271,9 +271,10 @@ export async function main(nodeProcess) {
     nodeProcess.exit(0);
   } catch (err) {
     if (err.code === 'ENOENT') {
-      nodeProcess.stderr.write('marked: output to ' + err.path + ': No such directory');
+      nodeProcess.stderr.write('marked: ' + err.path + ': No such directory');
+    } else {
+      nodeProcess.stderr.write(err.message);
     }
-    nodeProcess.stderr.write(err);
     return nodeProcess.exit(1);
   }
 }
--- a/bin/main.js
+++ b/bin/main.js
@@ -158,7 +158,7 @@ export async function main(nodeProcess) {
         input = files.pop();
       }
       if (!await fileExists(input)) {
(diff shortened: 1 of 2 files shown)


UziTech (changes requested):
Looks like some tests are failing. Can you also add a test for this so there is no regression in the future?

rzhade3:
@UziTech I've fixed the failing tests and attempted to write a test to prevent regressions, but it looked like this part of the code is currently not well tested. I seem to be running into this error whenever I attempt to pass a file as an input to the tests, that I'd appreciate some help debugging:

image

UziTech:
looks like the error was in the bin test framework

bin/main.js
```diff
@@ -157,6 +157,9 @@ export async function main(nodeProcess) {
         }
         input = files.pop();
       }
+      if (!await fileExists(input)) {
+        throw Error(`Cannot load input file '${input}'`);
+      }
```
styfle, line 162: Checking if a file exists right before reading it could lead to a race condition. The Node.js docs warn against this here:

> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

https://nodejs.org/api/fs.html#fsexistspath-callback

rzhade3: I've removed this exists call in https://github.com/markedjs/marked/pull/3165/commits/42bbb8604a0573814d464e11a000cf6d361dba58 to rely on the ENOENT error check instead.

bin/main.js
```diff
@@ -271,9 +270,10 @@ export async function main(nodeProcess) {
     nodeProcess.exit(0);
   } catch (err) {
     if (err.code === 'ENOENT') {
-      nodeProcess.stderr.write('marked: output to ' + err.path + ': No such directory');
+      nodeProcess.stderr.write('marked: ' + err.path + ': No such directory');
```
styfle, line 273:
```suggestion
      nodeProcess.stderr.write('marked: ' + err.path + ': No such file or directory');
```

Should this mention file or directory since its not always a directory?

styfle (approved):
Great, thanks!

Issue #3169, opened 2024-01-19
Incompatibility with esNext Syntax in marked.js Bundle


nbutmickey:
**What pain point are you perceiving?.**
1. we've observed that the output bundle from marked.js contains esNext syntax like ?. or ??. It's unreasonable and irresponsible for such a widely used library to include syntax incompatible with older browsers. [detail code](https://github.com/markedjs/marked/blob/6f4467ca9762f3a502c400ddc7687a27985f050d/src/Instance.ts#L258) 
2. upon inspecting the source code, it appears that the bundle is generated using Rollup with only the Terser plugin configured, without compatibility measures for advanced syntax.

**Describe the solution you'd like**
1. configuring the Buble plugin for Rollup to perform code transpilation.
2. remove the code involving the ?. or ?? syntax and implement suitable compatible code. 

Sure, I think option one can be a one-time solution, but it may lead to an increase in package size. Option two might be a bit more complicated as it requires replacing each one individually, and it's essential to ensure that subsequent code doesn't introduce similar syntax.

UziTech:
Option 1 is already available for your project. You can use Babel to compile marked to whatever version of es you need.

We decided to drop support a while ago for es5 since the es5 version of marked was about 30% slower than esNext.

It is unreasonable to believe marked knows what is best for all of its users. If you want speed you can use the esNext version. If you need wider support you can compile it with Babel.

nbutmickey:
> Option 1 is already available for your project. You can use Babel to compile marked to whatever version of es you need.
> 
> We decided to drop support a while ago for es5 since the es5 version of marked was about 30% slower than esNext.
> 
> It is unreasonable to believe marked knows what is best for all of its users. If you want speed you can use the esNext version. If you need wider support you can compile it with Babel.

1. Do you have relevant data supporting the conclusion that the ES5 version of 'marked' is 30% slower than the ESNext version?

2. I believe that as a library with such broad coverage, it should address compatibility issues. Assuming I create a library 'A' based on 'marked,' users of my library 'A' will install 'marked,' but if they don't use Babel to transpile ESNext code, issues may arise. We can't require all users to mandate Babel for compilation. This conclusion applies to library 'A' as well, and it's impractical to expect all authors of library 'A' to transpile your code.

In summary, I think it's crucial for an open-source library to handle compatibility issues well, or it can bring endless problems to developers.

UziTech:
[...]

PR #3171, opened 2024-01-20
feature:  Fix all products to support ES5.

nbutmickey:
**Marked version:**

11.1.1

**Markdown flavor:** n/a

## Description
This PR fixes the issue of the output not supporting ES5, which requires users to compile marked twice.

- Fixes https://github.com/markedjs/marked/issues/3169

## Expectation
The output in version 11.1.1 contains a significant amount of ES6 syntax, requiring users to compile marked twice. Otherwise, it won't run in environments that do not support ES6 syntax, as shown in the following image.

image

## Result
here is local build bundle
image

## What was attempted
none

## Contributor

## Committer

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -59,7 +59,7 @@ export default defineConfig([
       banner
     }],
     plugins: [
-      typescript()
+      typescript({tsconfig: 'tsconfig.build.json'}),
     ]
   }
 ]);
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -1,0 +1,7 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+      "target": "es5",
+    },
+  }
+  


UziTech:
Will this make the regular expressions that use unicode properties compatible with es5 browsers?

UziTech:
If we are going to support es5 browsers we will need tests to ensure no breaking changes. Can you write a test framework to run all the tests on old browsers as well. Thanks!

nbutmickey:
> Will this make the regular expressions that use unicode properties compatible with es5 browsers?
Compile it into ES5 code using only the TypeScript compilation engine.

nbutmickey:
> Will this make the regular expressions that use unicode properties compatible with es5 browsers?

Compile it into ES5 code using only the TypeScript compilation engine.

nbutmickey:
> If we are going to support es5 browsers we will need tests to ensure no breaking changes. Can you write a test framework to run all the tests on old browsers as well. Thanks!

hhh，it is really a huge work for me，I understand that code compiled using the TypeScript engine should be reliable.

UziTech:
Ya it is really huge work for anyone. Which is why we don't support older browsers

PR #3172, opened 2024-01-20
feat: flatten childToken arrays

### LINKED ISSUE #3170: Register Walktokens Behavior for extensions with different token structure
**Describe the feature**
In `walkTokens()`, traversing child tokens has to be handled differently for `table`s and `list`s, due to not holding a standard `tokens` object. 

https://github.com/markedjs/marked/blob/6f4467ca9762f3a502c400ddc7687a27985f050d/src/Instance.ts#L44C7-L61

I have recently noticed that my extension [marked-extended-tables](https://www.npmjs.com/package/marked-extended-tables) fails to be walked correctly, since it has *nearly* the same structure as `table` but a different token type `spanTable`. I would like some way to register the expected walk path for extensions that do not have the standard `tokens` object.

For compatibility with other extensions that use `walkTokens()`, because child tokens of `marked-extended-table` are not properly traversed.

Some kind of "on-the-fly" changing of the `spanTable` type back and forth between `table` and `spanTable` so it can still land in the correct `spanTable` Renderer, but this also fails due to slight differences in my token (`spanTable` supports multiple header rows for example).

UziTech:
**Marked version:** 11.1.1

## Description

Flatten custom extension childToken arrays before running them through walk tokens. This will allow custom extensions to have child tokens like the `table` token where the `rows` field is an array of arrays of tokens.

- Fixes #3170 

## Contributor

## Committer

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -63,7 +63,8 @@ export class Marked {
           const genericToken = token as Tokens.Generic;
           if (this.defaults.extensions?.childTokens?.[genericToken.type]) {
             this.defaults.extensions.childTokens[genericToken.type].forEach((childTokens) => {
-              values = values.concat(this.walkTokens(genericToken[childTokens], callback));
+              const tokens = genericToken[childTokens].flat(Infinity) as Token[] | TokensList;
+              values = values.concat(this.walkTokens(tokens, callback));
             });
           } else if (genericToken.tokens) {
             values = values.concat(this.walkTokens(genericToken.tokens, callback));
--- a/test/unit/marked.test.js
+++ b/test/unit/marked.test.js
@@ -395,6 +395,43 @@ describe('marked unit', () => {
                     + '\n<dt><strong>Topic 2 walked</strong> - unwalked</dt><dd><em>Description 2 walked</em></dd></p>\n');
     });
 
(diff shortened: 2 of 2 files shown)


UziTech:
@calculuschild this should work to walk arrays of arrays of tokens.

One question: Should we flatten with infinite depth? or should we only flatten 1 or 2 levels deep?

The only issue I found with doing Infinite depth is there can be a stack overflow if there is a circular dependency. I don't think that is a big issue but something to think about.

```
a = [1]
b = [2]
a.push(b)
b.push(a)
a.flat(Infinity) 
```

calculuschild (approved):
I *think* infinite depth shouldn't be a problem. It people are designing extensions in such a way to cause a cyclic dependency, that's a fault of the extension, not Marked, right?

calculuschild:
Github won't let me comment on the line, but with this change, does this let us simplify the special handling of Table tokens up on line 51?

UziTech:
I agree. This shouldn't affect core marked at all. Only extensions

styfle (approved):
I think you could also use a large-ish depth instead of Infinity, perhaps 1000, and still cover most use cases.

But I agree, its only for self-referencing extensions which is probably a small minority.

PR #3229, opened 2024-03-15
fix: punctuation regex syntax compatibility

musicq:
**Marked version:** 12.0.1

## Description

- Fix punctuation regex syntax compatibility

In some runtime environment, the punctuation regex syntax is not supported. Here's an example in Lark app.

image

## Expectation

Marked should work fine in most cases.

## Result

Host environment reports error due to unsupported regex syntax.

## What was attempted

Nothing can't be executed correctly.

## Contributor

## Committer

- [x] Squash and Merge PR following [conventional commit guidelines](https://www.conventionalcommits.org/).

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,8 +50,10 @@
     "html"
   ],
   "devDependencies": {
-    "@markedjs/testutils": "12.0.0-0",
     "@arethetypeswrong/cli": "^0.15.1",
+    "@babel/plugin-transform-unicode-regex": "^7.23.3",
+    "@markedjs/testutils": "12.0.0-0",
+    "@rollup/plugin-babel": "^6.0.4",
     "@rollup/plugin-terser": "^0.4.4",
     "@rollup/plugin-typescript": "^11.1.6",
     "@semantic-release/commit-analyzer": "^11.1.0",
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -1,5 +1,5 @@
 import {
-  edit, noopTest
+  edit, isSupportPunctuationRegex, noopTest
 } from './helpers.ts';
 
 /**
@@ -170,7 +170,8 @@ const br = /^( {2,}|\\)\n(?!\s*$)/;
 const inlineText = /^(`+|[^`])(?:(?= {2,}\n)|[\s\S]*?(?:(?=[\\<!\[`*_]|\b_|$)|[^ ](?= {2,}\n)))/;
 
 // list of unicode punctuation marks, plus any missing characters from CommonMark spec
-const _punctuation = '\\p{P}\\p{S}';
+// fallback to explicit punctuation list if runtime environment doesn't support punctuation regex syntax
+const _punctuation = isSupportPunctuationRegex ? '\\p{P}\\p{S}' : '!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~';
 const punctuation = edit(/^((?![*_])[\spunctuation])/, 'u')
   .replace(/punctuation/g, _punctuation).getRegex();
 
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -1,5 +1,5 @@
 import {
-  edit, isSupportPunctuationRegex, noopTest
+  edit, noopTest
 } from './helpers.ts';
 
 /**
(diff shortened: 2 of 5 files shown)


src/rules.ts
```diff
@@ -170,7 +170,8 @@ const br = /^( {2,}|\\)\n(?!\s*$)/;
 const inlineText = /^(`+|[^`])(?:(?= {2,}\n)|[\s\S]*?(?:(?=[\\<!\[`*_]|\b_|$)|[^ ](?= {2,}\n)))/;
 
 // list of unicode punctuation marks, plus any missing characters from CommonMark spec
-const _punctuation = '\\p{P}\\p{S}';
+// fallback to explicit punctuation list if runtime environment doesn't support punctuation regex syntax
+const _punctuation = isSupportPunctuationRegex ? '\\p{P}\\p{S}' : '!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~';
```
UziTech, line 174: These two regular expressions are very different which means markdown will render differently in different browsers.

You should be able to just change the string to a regular expression. Then babel can transpile it correctly.

```diff
- const _punctuation = '\\p{P}\\p{S}';
+ const _punctuation = /\p{P}\p{S}/u;
```

UziTech (changes requested):

package.json
```diff
@@ -50,8 +50,10 @@
     "html"
   ],
   "devDependencies": {
-    "@markedjs/testutils": "12.0.0-0",
     "@arethetypeswrong/cli": "^0.15.1",
+    "@babel/plugin-transform-unicode-regex": "^7.23.3",
+    "@markedjs/testutils": "12.0.0-0",
+    "@rollup/plugin-babel": "^6.0.4",
```
UziTech, line 56: We don't want to use babel for the output. This will make marked a lot slower and a lot bigger. If users need to support older browsers they can use babel themselves.

UziTech (changes requested):

src/rules.ts
```diff
@@ -170,44 +170,41 @@ const br = /^( {2,}|\\)\n(?!\s*$)/;
...
-+ '|[^punct\\s](\\*+)(?!\\*)(?=[punct\\s]|$)' // (2) a***#, a*** can only be a Right Delimiter
-+ '|(?!\\*)[punct\\s](\\*+)(?=[^punct\\s])' // (3) #***a, ***a can only be Left Delimiter
-+ '|[\\s](\\*+)(?!\\*)(?=[punct])' // (4) ***# can only be Left Delimiter
-+ '|(?!\\*)[punct](\\*+)(?!\\*)(?=[punct])' // (5) #***# can be either Left or Right Delimiter
-+ '|[^punct\\s](\\*+)(?=[^punct\\s])', 'gu') // (6) a***a can be either Left or Right Delimiter
-  .replace(/punct/g, _punctuation)
-  .getRegex();
+const emStrongLDelim = /^(?:\*+(?:((?!\*)[\p{P}\p{S}])|[^\s*]))|^_+(?:((?!_)[\p{P}\p{S}])|([^\s_]))/u;
+
+const emStrongRDelimAst = /^[^_*]*?__[^_*]*?\*[^_*]*?(?=__)|[^*]+(?=[^*])|(?!\*)[\p{P}\p{S}](\*+)(?=[\s]|$)|[^\p{P}\p{S}\s](\*+)(?!\*)(?=[\p{P}\p{S}\s]|$)|(?!\*)[\p{P}\p{S}\s](\*+)(?=[^\p{P}\p{S}\s])|[\s](\*+)(?!\*)(?=[\p{P}\p{S}])|(?!\*)[\p{P}\p{S}](\*+)(?!\*)(?=[\p{P}\p{S}])|[^\p{P}\p{S}\s](\*+)(?=[^\p{P}\p{S}\s])/gu;
```
UziTech, line 181: I don't think you need to change this regex since babel will replace the regex in `_punctuation` with a longer regex then `replace` will use that regex for `punct`

UziTech (changes requested):

src/rules.ts
musicq: I've tried that, but seems to not work. The `replace` method will malform the result and lead to the tests fail.

UziTech: hmm. Then it seems to me that it is not worth the extra work to support browsers that are > 50 versions behind

musicq: Then I'll just close this PR. Thanks

package.json
musicq: Actually I think babel here is just to transform the unicode regex, won't do other stuff 🤔

PR #3230, opened 2024-03-15
Allow generic type parameters for custom extensions

### LINKED ISSUE #3228: Improved TypeScript types for TokenizerAndRendererExtension
I've recently started writing custom some `TokenizerAndRendererExtension` code in a personal project, using TypeScript, and I've found it slightly annoying to work with the way the extension's `renderer` method receives a `Token.Generic` even though (I assume) it should be possible to know that it receives a token of the type returned by its `tokenizer` method.

Current solutions involve either using type assertions to tell the TypeScript compiler that I know more about the structure of these tokens than TypeScript is telling me, or validating the shape ourselves (which adds additional work to rendering that AFAIK shouldn't be necessary so isn't ideal). Both approaches require manually keeping type assertions or checks in the `renderer` updated whenever the shape of the token changes in the `tokenizer`.

On a related but less important note, it could be nice to also enforce the constraint that the extension's `name` string must be the same as its token's `name` string.

Currently, marked provides a `Token.Generic` type for general-use tokens, which looks like this:

```typescript
interface Generic {
	[index: string]: any;
	type: string;
	raw: string;
	tokens?: Token[] | undefined;
}
```

This allows access to any properties, but they are typed unsafely as `any` so it's up to the consuming code to decide how to deal with that.

I think this could be resolved by allowing `TokenizerAndRendererExtension` to take a generic type parameter, defining the type of token it uses internally. I believe this could be written in such a way as to allow that generic type parameter to be inferred through the value returned by the `tokenizer` function. The parameter could have a default type that would maintain the current `any` behaviour, so existing code that references this type without a generic type parameter would still work in the same way.

However, this approach would also necessarily make a few other internal types (in particular `TokenizerExtension` and `RendererExtension`) generic as well. Without having attempted to implement this, I'm not sure if there would be additional complexity spreading to other types. I don't have any knowledge currently of marked's codebase, aside perhaps from having viewed the type declaration files it ships with.

Cipscis:
**Marked version: 11.1.0**

**Markdown flavor:** n/a

## Description

- Fixes #3228

This PR adds optional generic type parameters to the types used for creating custom extensions. This allows marked to communicate more specific expectations to TypeScript for the token returned by a custom `tokenizer` function, and for the token passed to a custom `renderer` function.

Example usage:

```typescript
const myCustomExtension: TokenizerAndRendererExtension<
  'myCustomExtension',
  { example: string }
> = {
  name: 'myCustomExtension',
  level: 'inline',
  start(src) {
    // ...
  },
  tokenizer(src, tokens) {
    // ...

    return {
      type: 'myCustomExtension',
      raw: src,
      example: 'must be a string',
      tokens: [],
    };
  },
  renderer(token) {
    token.type; // <- 'myCustomExtension'
    token.example; // <- string
    // ...
  },
};
```

Unfortunately, it's not quite so easy to set this to allow the generic type parameters to be inferred as I had hoped. I'd forgotten when writing the issue that TypeScript is only able to infer generic types in function parameters, so that can't be done with type updates alone.

Enabling this sort of type inference would require a noop function wrapper to allow TypeScript to do inference, and I'm not sure if that would be a useful addition to marked's API:

```typescript
function createExtension<
  N extends string = string,
  T extends Record<string, unknown> = Record<string, any>
>(extension: TokenizerAndRendererExtension<N, T>): TokenizerAndRendererExtension<N, T> {
  return extension;
}

const myCustomExtension = createExtension({
  // extension code as in previous example
});
// ^ TokenizerAndRendererExtension<"myCustomExtension", {
//   type: "myCustomExtension";
//   raw: string;
//   example: string;
//   tokens: never[];
// }>
```

I'm also less certain about how valuable it is to include the name in the generic parameter. Particularly as it's necessary to declare `as const` if the token is not immediately returned, e.g.

```typescript
tokenizer(src, tokens) {
  // ...

  const token = {
    type: 'myCustomExtension' as const,
    raw,
    example: 'must be a string',
    tokens: [],
  };
  return token;
},
```

Because this PR only contains type changes, no actual functionality should have changed. And because the only changes have been adding generic type arguments *with defaults* that replicate the existing behaviour, I expect there wouldn't be any breaking changes with existing TypeScript code.

I'm not sure where best to document this feature's PR, so currently have not included any new documentation.

## Contributor

## Committer

### DIFF
--- a/src/MarkedOptions.ts
+++ b/src/MarkedOptions.ts
@@ -1,4 +1,4 @@
-import type { Token, Tokens, TokensList } from './Tokens.ts';
+import type { CustomToken, Token, TokensList } from './Tokens.ts';
 import type { _Parser } from './Parser.ts';
 import type { _Lexer } from './Lexer.ts';
 import type { _Renderer } from './Renderer.ts';
@@ -9,15 +9,21 @@ export interface TokenizerThis {
   lexer: _Lexer;
 }
 
-export type TokenizerExtensionFunction = (this: TokenizerThis, src: string, tokens: Token[] | TokensList) => Tokens.Generic | undefined;
+export type TokenizerExtensionFunction<
+  N extends string = string,
+  T extends Record<string, unknown> = Record<string, any>
+> = (this: TokenizerThis, src: string, tokens: Token[] | TokensList) => CustomToken<N, T> | undefined;
 
 export type TokenizerStartFunction = (this: TokenizerThis, src: string) => number | void;
 
-export interface TokenizerExtension {
-  name: string;
+export interface TokenizerExtension<
+  N extends string = string,
+  T extends Record<string, unknown> = Record<string, any>
+> {
+  name: N;
   level: 'block' | 'inline';
   start?: TokenizerStartFunction | undefined;
-  tokenizer: TokenizerExtensionFunction;
+  tokenizer: TokenizerExtensionFunction<N, T>;
   childTokens?: string[] | undefined;
 }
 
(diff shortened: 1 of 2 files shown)


UziTech:
You can create tests for types in https://github.com/markedjs/marked/blob/master/test/types/marked.ts

The best place to add docs would probably be https://github.com/markedjs/marked/blob/master/docs/USING_PRO.md#custom-extensions--extensions

UziTech:
@Cipscis are you still working on this?

Cipscis:
Sorry, no I haven't touched it since March

Issue #236, opened 2013-08-24
HTML block parsing in consistency


roydukkey:
Here is what I don't understand: Why does marked parse some features in HTML blocks and not others?
#### Example 1

```
 <div>
 ## Markdown header
 *Markdown header*
 **Markdown header**
 </div>
```

**Outputs**

```
 <div>
 ## Markdown header
 <em>Markdown header</em>
 <strong>Markdown header</strong>
 </div>
```
#### Example 2

```
 <div>
 ## Markdown header
 *Markdown header*
 **Markdown header**
 </div>
 <script>
 var nubbin;
 </script>
```

**Output**

```
 <p><div></p>
 <h2 id="markdown-header">Markdown header</h2>
 <p><em>Markdown header</em>
 <strong>Markdown header</strong>
 </div></p>
 <script>
 var nubbin;
 </script>
```

I would like, most, to see these issues being fixed, allowing HTML blocks to be truly untouched. Then, begin further discussion on other topics like #66, #133, #115.

donleyp:
Is there a way to disable this "extension"?

joshbruce:
See #956

Closing as stale, please open a new issue with reference is you are still using marked and have the same request. Thank you.

@UziTech: Here's one of those moments I'm curious about regarding the continued use of marked "in the wild". Marked looks unkempt according to the cross-project PR referenced above. Having said that, the PR is still open after a year. And that project is starting show signs of either be stale, completely stable, or unused.

Given what I assume to be low usage numbers for Elm, in general, I would not be surprised. Anyway, just considering the possibilities of false positives for marked's prevalence in the wild on active projects.

PR #3264, opened 2024-04-20
fix: blockquote code continuation

### LINKED ISSUE #3167: Lexer handling newlines incorrectly in some cases
**Marked version:**
11.1.1

**Describe the bug**
When using the lexer it seems to leave newlines at the end of some tokens instead of tokenizing them

**To Reproduce**
Input (hr):
```md
console.log(lexer.lex("---------------------------------\n\nhi"))
```
Output (hr):
```json
[
    {
        "type": "hr",
        "raw": "---------------------------------\n\n"
    },
    {
        "type": "paragraph",
        "raw": "hi",
        "text": "hi",
        "tokens": [
            {
                "type": "text",
                "raw": "hi",
                "text": "hi"
            }
        ]
    }
]
```

and input (blockquote):
```md
console.log(lexer.lex("> blockquote\n\nhi"))
```
Output (blockquote):
```json
[
    {
        "type": "blockquote",
        "raw": "> blockquote\n\n",
        "tokens": [
            {
                "type": "paragraph",
                "raw": "blockquote",
                "text": "blockquote",
                "tokens": [
                    {
                        "type": "text",
                        "raw": "blockquote",
                        "text": "blockquote"
                    }
                ]
            }
        ],
        "text": "blockquote"
    },
    {
        "type": "paragraph",
        "raw": "hi",
        "text": "hi",
        "tokens": [
            {
                "type": "text",
                "raw": "hi",
                "text": "hi"
            }
        ]
    }
]
```

For both of these examples you can see that the 2 newlines are being ignored and not tokenized by the lexer.
This is with `gfm: true` and `breaks: true`

**Expected behavior**
For hr input:
```json
[
    {
        "type": "hr",
        "raw": "---------------------------------"
    },
    {
        "type": "space",
        "raw": "\n\n"
    },
    {
        "type": "paragraph",
        "raw": "hi",
        "text": "hi",
        "tokens": [
            {
                "type": "text",
                "raw": "hi",
                "text": "hi"
            }
        ]
    }
]
```

For blockquote input:
```json
[
    {
        "type": "blockquote",
        "raw": "> blockquote",
        "tokens": [
            {
                "type": "paragraph",
                "raw": "blockquote",
                "text": "blockquote",
                "tokens": [
                    {
                        "type": "text",
                        "raw": "blockquote",
                        "text": "blockquote"
                    }
                ]
            }
        ],
        "text": "blockquote"
    },
    {
        "type": "br",
        "raw": "\n"
    },
    {
        "type": "paragraph",
        "raw": "hi",
        "text": "hi",
        "tokens": [
            {
                "type": "text",
                "raw": "hi",
                "text": "hi"
            }
        ]
    }
]
```

UziTech:
**Marked version:** 12.0.2

## Description

Fix blockquote continuation for code blocks

Fixes [CommonMark #236](https://spec.commonmark.org/0.31.2/#example-236)
Fixes [CommonMark #237](https://spec.commonmark.org/0.31.2/#example-237)

also adds space token after blockquote and hr if there are multiple newlines. fixes #3167.

I believe this will be a breaking change since tokens will change even though the output won't change.

## Contributor

## Committer

### DIFF
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -166,6 +166,7 @@ export class _Tokenizer {
         const currentLines = [];
 
         while (lines.length > 0) {
+          // get lines up to a continuation
           if (/^ {0,3}>/.test(lines[0])) {
             currentLines.push(lines.shift());
             inBlockquote = true;
@@ -178,16 +179,20 @@ export class _Tokenizer {
 
         const currentRaw = currentLines.join('\n');
         const currentText = currentRaw
-        // precede setext continuation with 4 spaces so it isn't a setext
+          // precede setext continuation with 4 spaces so it isn't a setext
           .replace(/\n {0,3}((?:=+|-+) *)(?=\n|$)/g, '\n    $1')
           .replace(/^ {0,3}>[ \t]?/gm, '');
         raw = raw ? `${raw}\n${currentRaw}` : currentRaw;
         text = text ? `${text}\n${currentText}` : currentText;
+
+        // parse blockquote lines as top level tokens
+        // merge paragraphs if this is a continuation
         const top = this.lexer.state.top;
         this.lexer.state.top = true;
         this.lexer.blockTokens(currentText, tokens, true);
         this.lexer.state.top = top;
 
+        // if there is no continuation then we are done
         if (lines.length === 0) {
           break;
         }
@@ -195,8 +200,10 @@ export class _Tokenizer {
         const lastToken = tokens[tokens.length - 1];
 
         if (lastToken?.type === 'code') {
+          // blockquote continuation cannot be preceded by a code block
           break;
         } else if (lastToken?.type === 'blockquote') {
+          // include continuation in nested blockquote
           const oldBlockquoteToken = lastToken as Tokens.Blockquote;
           const newText = oldBlockquoteToken.raw + '\n' + lines.join('\n');
           const newBlockquoteToken = this.blockquote(newText)!;
@@ -206,6 +213,7 @@ export class _Tokenizer {
           text = text.substring(0, text.length - oldBlockquoteToken.text.length) + newBlockquoteToken.text;
           break;
         } else if (lastToken?.type === 'list') {
+          // include continuation in nested list
           const oldListToken = lastToken as Tokens.List;
           const newText = oldListToken.raw + '\n' + lines.join('\n');
           const newListToken = this.list(newText)!;
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -165,17 +165,19 @@ export class _Tokenizer {
         let inBlockquote = false;
         const currentLines = [];
 
-        while (lines.length > 0) {
+        let i;
+        for (i = 0; i < lines.length; i++) {
           // get lines up to a continuation
-          if (/^ {0,3}>/.test(lines[0])) {
-            currentLines.push(lines.shift());
+          if (/^ {0,3}>/.test(lines[i])) {
+            currentLines.push(lines[i]);
             inBlockquote = true;
           } else if (!inBlockquote) {
-            currentLines.push(lines.shift());
+            currentLines.push(lines[i]);
           } else {
             break;
           }
         }
+        lines = lines.slice(i);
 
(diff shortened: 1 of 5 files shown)


src/Tokenizer.ts
```diff
@@ -148,24 +148,87 @@ export class _Tokenizer {
...
+          const oldListToken = lastToken as Tokens.List;
+          const newText = oldListToken.raw + '\n' + lines.join('\n');
+          const newListToken = this.list(newText)!;
+          tokens[tokens.length - 1] = newListToken;
+
+          raw = raw.substring(0, raw.length - lastToken.raw.length) + newListToken.raw;
+          text = text.substring(0, text.length - oldListToken.raw.length) + newListToken.raw;
+          lines = newText.substring(tokens[tokens.length - 1].raw.length).split('\n');
+          continue;
+        }
```
calculuschild, line 226: This section seems to have some code duplication. Would it be useful to combine the logic for `list` and `blockquote` here a bit?

Just playing around with an example without actually testing, I'm thinking something like this:

```
    if (lastToken?.type === 'blockquote' || lastToken?.type === 'list') {
      const oldToken = lastToken as Tokens.Blockquote | Tokens.List;
      const newText = oldToken.raw + '\n' + lines.join('\n');
      const newToken = this[lastToken.type]!(newText);
      tokens[tokens.length - 1] = newToken;

      raw = raw.substring(0, raw.length - oldToken.raw.length) + newToken.raw;
      text = text.substring(0, text.length - oldToken.text.length) + newToken.text;
      if (lastToken.type === 'list') {
        lines = newText.substring(tokens[tokens.length - 1].raw.length).split('\n');
      }
    }
 ```

File: src/Tokenizer.ts [hunk 2]
```diff
@@ -148,24 +148,87 @@ export class _Tokenizer {
...
+      const tokens: Token[] = [];
+
+      while (lines.length > 0) {
+        let inBlockquote = false;
+        const currentLines = [];
+
+        while (lines.length > 0) {
+          // get lines up to a continuation
+          if (/^ {0,3}>/.test(lines[0])) {
+            currentLines.push(lines.shift());
```
calculuschild, line 171: Is there any performance concern for `.shift()` here with potentially many lines?

src/Tokenizer.ts
```diff
@@ -148,24 +148,87 @@ export class _Tokenizer {
...
-      const top = this.lexer.state.top;
-      this.lexer.state.top = true;
-      const tokens = this.lexer.blockTokens(text);
-      this.lexer.state.top = top;
+      let lines = rtrim(cap[0], '\n').split('\n');
+      let raw = '';
+      let text = '';
+      const tokens: Token[] = [];
+
+      while (lines.length > 0) {
```
calculuschild, line 164: Also concerned that this seems to process *every* line in a document each time a blockquote rule is reached. Is there any other shortcut we should consider to break out of the loop early besides the check for `lastToken` at the end?

src/Tokenizer.ts
UziTech: The `Tokens.Blockquote | Tokens.List` will mean we can't use `oldToken.text` since Tokens.List doesn't have a `text` property. And `text = text.substring(...` line is different between them.

I figured rather than peppering if statements for their differences it would be cleaner to duplicate some of the code. 

If there were more than two duplicates I think it would be worth it to create a method but DRY typically only helps when code is duplicated more than 3 times.

src/Tokenizer.ts
UziTech: Good call! Probably better to not shift. I'll update that

src/Tokenizer.ts
UziTech: This only processes every line that could be part of the blockquote. We already break on multiple new lines and hr and stuff

src/Tokenizer.ts
calculuschild: In that case, what about reusing the same naming structure in each block for easier maintenance burden? The two are 90% identical, but this is difficult to see when `oldListToken` vs `oldBlockquoteToken` is making them appear more different than they are.

For me, it would be easier to visualize the differences between the two if the identical sections were also labeled identically.

PR #3291, opened 2024-05-15
feat: Send token objects to renderers

### LINKED ISSUE #2640: Pass token to renderer methods ?
**Describe the feature**
Currently renderer methods get arguments cooked by marked ~Renderer~ Parser, as in `code(code, infostring, escaped)`.
On the other hand, extensions rendered methods get a `token` parameter with all its properties.
I request that all renderer methods get a token.

For example this.renderer.list won't know if the list is loose.

I could have asked for this.renderer.list to get a `(body, ordered, start, loose)` signature.
But passing the token around systematically seems to be much simpler.

UziTech:
**Marked version:** v12.0.2

## Description

- Send token objects to renderers and move logic to parse tokens from the parser to the renderers. 
	- Fixes #2640
- Add `parser` as a property on the Renderer object
- Add space renderer that returns empty string by default
- Send block text tokens to the text renderer
- Add `header` and `align` properties to TableCell token
- Add TableRow token
- Add Checkbox token

## Contributor

## Committer

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -68,16 +68,16 @@ import { marked } from 'marked';
 
 // Override function
 const renderer = {
-  heading(text, level) {
+  heading({ text, depth }) {
     const escapedText = text.toLowerCase().replace(/[^\w]+/g, '-');
 
     return `
-            <h${level}>
+            <h${depth}>
               <a name="${escapedText}" class="anchor" href="#${escapedText}">
                 <span class="header-link"></span>
               </a>
               ${text}
-            </h${level}>`;
+            </h${depth}>`;
   }
 };
 
@@ -97,7 +97,7 @@ console.log(marked.parse('# heading+'));
   heading+
 </h1>
 ```
-**Note:** Calling `marked.use()` in the following way will avoid overriding the `heading` token output but create a new renderer in the process.
+**Note:** Calling `marked.use()` in the following way will avoid overriding the `heading` token output but create a new `heading` renderer in the process.
 
 ```js
 marked.use({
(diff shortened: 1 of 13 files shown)


docs/USING_PRO.md
```diff
@@ -192,33 +186,35 @@ console.log(marked.parse('$ latex code $\n\n` other code `'));
...
+- <code>**reflink**(src: *string*, links: *object*): *Tokens.Link | Tokens.Image | Tokens.Text*</code>
+- <code>**emStrong**(src: *string*, maskedSrc: *string*, prevChar: *string*): *Tokens.Em | Tokens.Strong*</code>
+- <code>**codespan**(src: *string*): *Tokens.Codespan*</code>
+- <code>**br**(src: *string*): *Tokens.Br*</code>
+- <code>**del**(src: *string*): *Tokens.Del*</code>
+- <code>**autolink**(src: *string*): *Tokens.Link*</code>
+- <code>**url**(src: *string*): *Tokens.Link*</code>
+- <code>**inlineText**(src: *string*): *Tokens.Text*</code>
+
+The Tokens.* properties can be found [here](https://github.com/markedjs/marked/blob/master/src/Tokens.ts).
```
styfle, line 217:
```suggestion
The *Tokens.* properties can be found [here](https://github.com/markedjs/marked/blob/master/src/Tokens.ts).
```
Same here

src/Renderer.ts
```diff
@@ -4,72 +4,131 @@ import {
...
 import type { MarkedOptions } from './MarkedOptions.ts';
+import type { Tokens } from './Tokens.ts';
+import type { _Parser } from './Parser.ts';
 
 /**
  * Renderer
  */
 export class _Renderer {
   options: MarkedOptions;
+  parser!: _Parser; // set by the parser
```
styfle, line 15: Can we set this in the constructor instead?

styfle (approved):

docs/USING_PRO.md
```diff
@@ -112,38 +112,32 @@ marked.use({
...
 
-`flags` has the following properties:
-
-```js
-{
-    header: true || false,
-    align: 'center' || 'left' || 'right'
-}
-```
+The Tokens.* properties can be found [here](https://github.com/markedjs/marked/blob/master/src/Tokens.ts).
```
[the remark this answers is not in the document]
UziTech: No the star is meant to mean any type in the `Tokens` namespace. Like ` Tokens.Code` etc.

src/Renderer.ts
UziTech: No because the Renderer is instantiated before the Parser any time a renderer extension is used. The tokenizer has the same problem with the lexer being set this way. 

This problem mostly exists since the Lexer and Parser  can be used in a static way `marked.lexer(options)`

Maybe we can require instance in the future

styfle: > Maybe we can require instance in the future

That sounds good to me 👍

UziTech:
There seems to be a problem with the docs for this change since we use the current marked but a version of marked-highlight that does not support this new renderer parameters.

I think I am going to have to find a way for both to be available for this to work so we can transition extensions.

UziTech:
I created a temporary option `useNewRenderer` so extensions can opt in to the new function signature. This should allow most renderer extensions to work without changing ("most" because there are some edge cases that we can't mock).

This temporary option will be removed in a future major version which will break renderer extensions that have not updated to accepting an object.

src/Instance.ts
```diff
@@ -146,15 +147,19 @@ export class Marked {
...
           }
-          const rendererProp = prop as Exclude<keyof _Renderer, 'options'>;
-          const rendererFunc = pack.renderer[rendererProp] as GenericRendererFunction;
+          const rendererProp = prop as Exclude<keyof _Renderer, 'options' | 'parser'>;
+          let rendererFunc = pack.renderer[rendererProp] as GenericRendererFunction;
           const prevRenderer = renderer[rendererProp] as GenericRendererFunction;
           // Replace renderer with func to run extension, but fall back if false
           renderer[rendererProp] = (...args: unknown[]) => {
+            if (!pack.useNewRenderer) {
+              // Remove this in next major version
```
styfle, line 160:
```suggestion
              // TODO: Remove this in next major version
```

src/Instance.ts
```diff
@@ -250,6 +255,250 @@ export class Marked {
     return this;
   }
 
+  // Remove this in next major release
```
styfle, line 258:
```suggestion
  // TODO: Remove this in next major release
```

styfle (approved):

PR #3317, opened 2024-06-09
Update Lexer section in USING_PRO.md

### LINKED ISSUE #3316: Unexpected behavior when using `marked.lexer` or `new marked.Lexer`
**Marked version:** v12.0.2

**Describe the bug**
Behavior is immutable (A) when using `marked.lexer` but it is somehow mutable (B) when using `new marked.Lexer().lex()`.

I'm using instance of a Lexer to call `lex()` with Markdown data and get tokens. When reusing the same instance, and calling `lex()` with different Markdown data, it returns the same tokens of the first call. This is not expected, or at least it is not [documented](https://github.com/markedjs/marked/blob/master/docs/USING_PRO.md#access-to-lexer-and-parser).

**To Reproduce**

1. See [Stackblitz](https://stackblitz.com/edit/marked-lexer-issue?file=src%2Fmain.ts) example (scroll at the end of `main.ts`).
2. Click option A, it shows unique content for car and plane.
2. Click option B, it shows unique content for car, but car + plane in second case.

**Expected behavior**
I would expect `lex()` to only work with the Markdown data used in method call. I wouldn't expect data from previous calls to pop in the returned value.

`marked.lexer` is a good candidate as a workaround. But maybe it is made like this by design? If so, it would be great to inform users about this in documentation.

poirierlouis:
**Marked version:** all

**Markdown flavor:** n/a

## Description

- Fixes #3316.

## Expectation

Documentation should be explicit for new users when using Lexer.

## Result

User is now informed of different behaviors when using `marked.lexer` or `new marked.Lexer`.

## What was attempted

n/a

## Contributor

## Committer

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -565,8 +565,6 @@ The parser takes tokens as input and calls the renderer functions.
 
 You also have direct access to the lexer and parser if you so desire. The lexer and parser options are the same as passed to `marked.setOptions()` except they have to be full options objects, they don't get merged with the current or default options.
 
-
-
 ``` js
 const tokens = marked.lexer(markdown, options);
 console.log(marked.parser(tokens, options));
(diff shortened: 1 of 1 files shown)


UziTech (approved):
thanks 💯

PR #3341, opened 2024-06-23
fix: allow async option to dictate type returned

UziTech:
**Marked version:** v13

## Description

BREAKING CHANGE: throw an error if `async: false` is sent with parse options when using an async extension.

This allows us to overload the parse method when the `async ` option is set.

```js
const html: Promise<string> = marked.parse(md, { async: true });
const html: string = marked.parse(md, { async: false});
const html: Promise<string> | string = marked.parse(md);
```

closes #3116

## Contributor

## Committer

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -21,8 +21,8 @@ export class Marked {
   defaults = _getDefaults();
   options = this.setOptions;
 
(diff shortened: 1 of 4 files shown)


UziTech:
@styfle @calculuschild Can I get an approval to merge this and release marked v14

Issue #3352, opened 2024-06-26
Integrate with Google OSS-Fuzz


LeoDog896:
**Describe the feature**

Add fuzzing to the core parser to help protect against availability attacks and constantly run them with Google's [OSS-Fuzz](https://github.com/google/oss-fuzz).

**Why is this feature necessary?**

`markedjs` has had a [few ReDoS vulnerabilities](https://github.com/markedjs/marked/security), which impact availability. If there are any other ReDoS vulnerabilities (or other availability-hindering vulnerabilities) that this repository doesn't know yet that can be caught using fuzzing, that can help secure dependents of this library.

If approved, I'll send a request to Google for integration.

UziTech:
Sounds good to me. 👍

LeoDog896:
Closed until https://github.com/google/oss-fuzz/issues/11652

PR #3353, opened 2024-06-27
test: add fuzzer

LeoDog896:
## Description

- Advances #3352 

## Result

Adds an `npm run fuzz` command that can be used to fuzz the `parse` and `parseInline` functions and their suboptions.

## Contributor

## Committer

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
       },
       "devDependencies": {
         "@arethetypeswrong/cli": "^0.15.3",
+        "@jazzer.js/core": "^2.1.0",
         "@markedjs/testutils": "12.0.0-0",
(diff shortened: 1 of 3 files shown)


UziTech:
looks like https://www.npmjs.com/package/@jazzer.js/core is no longer maintained

LeoDog896:
> looks like https://www.npmjs.com/package/@jazzer.js/core is no longer maintained

oh, yep! It seems that it's mentioned here - https://github.com/google/oss-fuzz/issues/11652

I assume this would be blocking for fuzzing integration, then.

UziTech:
I think that would be blocking. I don't want to introduce a dependency on something that could become a vulnerability.

The irony 🤣

PR #3367, opened 2024-07-12
chore: update eslint to v9

UziTech:
## Description

Update eslint to v9

This does mean removing standard since it is [no longer maintained](https://github.com/standard/standard/issues/1948#issuecomment-2138078249)

We could use [neostandard](https://github.com/neostandard/neostandard) which is the successor but I don't see why we wouldn't think the same thing will happen to it in the future. Just copying the config seems to be the best option to continue with the same rules.

Maybe a future PR can slim down the rules a bit. Personally I prefer only using `eslint:recommended` and `typescript-eslint:recommended` since it is simpler.

## Contributor

## Committer


test/specs/redos/cubic_def.cjs
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   markdown: `[x]:${' '.repeat(1500)}x ${' '.repeat(1500)} x`,
-  html: `<p>[x]:${' '.repeat(1500)}x ${' '.repeat(1500)} x</p>`,
+  html: `<p>[x]:${' '.repeat(1500)}x ${' '.repeat(1500)} x</p>`
```
styfle, line 3: Personally, I like trailing commas

package.json
```diff
@@ -83,30 +79,32 @@
...
-    "test:update": "node test/update-specs.js",
-    "rules": "node test/rules.js",
-    "bench": "npm run build && node test/bench.js",
-    "lint": "eslint --fix .",
-    "build:reset": "git checkout upstream/master lib/marked.cjs lib/marked.umd.js lib/marked.esm.js marked.min.js",
-    "build": "npm run rollup && npm run build:types && npm run build:man",
-    "build:docs": "npm run build && node docs/build.js",
-    "build:types": "tsc && dts-bundle-generator --project tsconfig.json -o lib/marked.d.ts src/marked.ts && dts-bundle-generator --project tsconfig.json -o lib/marked.d.cts src/marked.ts",
-    "build:man": "marked-man man/marked.1.md > man/marked.1",
-    "rollup": "rollup -c rollup.config.js"
```
styfle, line 109: Whats changing in the scripts here? Is it just the order?

styfle (approved):

test/specs/redos/cubic_def.cjs
UziTech: Me too. I was just kept the rules the same as before.

I will try changing the rules in a follow up pr. There will be a lot of src code changes

package.json
UziTech: Ya just reordering alphabetically

PR #3380, opened 2024-07-24
fix: fix recursion-like stack overflow error caused by the old render…

### LINKED ISSUE #3379: Version 13 , old rendering behavior, combined with a large number of execution calls, will lead to an "Maximum call stack size exceeded" error
**Marked version: 13**

**Describe the bug**

as title, code will throw "Maximum call stack size exceeded" error

**To Reproduce**

1. `npm install --save marked@13.0.2`
2. run code like below

```js

const content = '\n## 1. some title\n\nsimple content'

marked.use({
	renderer: {
		heading(text, level) {
			const escapedText = encodeURIComponent(text).replaceAll('%', '.');
			return `<h${level} id="${escapedText}">
									${text}
					</h${level}>`;
					
		},
	},
});

for (const _ of new Array(10000).fill(0)) {
	marked.parse(content);
}
```
4. code will throw error "RangeError: Maximum call stack size exceeded"
image

**some code dig**

As I read through the code, it seems that, in simple terms, reassignment `rendererFunc` leads to recursive-like calls (each execution results in an additional call).

image

yijian166:
fix #3379

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -158,7 +158,7 @@ export class Marked {
           renderer[rendererProp] = (...args: unknown[]) => {
             if (!pack.useNewRenderer) {
               // TODO: Remove this in next major version
-              rendererFunc = this.#convertRendererFunction(rendererFunc, rendererProp, renderer) as GenericRendererFunction;
+              rendererFunc = this.#convertRendererFunction(pack.renderer![rendererProp] as GenericRendererFunction, rendererProp, renderer) as GenericRendererFunction;
             }
             let ret = rendererFunc.apply(renderer, args);
             if (ret === false) {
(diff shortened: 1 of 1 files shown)


src/Instance.ts
```diff
@@ -158,7 +158,7 @@ export class Marked {
           renderer[rendererProp] = (...args: unknown[]) => {
             if (!pack.useNewRenderer) {
               // TODO: Remove this in next major version
-              rendererFunc = this.#convertRendererFunction(rendererFunc, rendererProp, renderer) as GenericRendererFunction;
+              rendererFunc = this.#convertRendererFunction(pack.renderer![rendererProp] as GenericRendererFunction, rendererProp, renderer) as GenericRendererFunction;
             }
```
UziTech, line 162: I feel like moving this out of the `renderer[rendererProp]` function would be better. There is no reason we have to assign rendererFunc every time the function is called.

UziTech (changes requested):
Nice catch!

src/Instance.ts
yijian166: yes, moving this out of make sense.

UziTech (approved):
Thanks for fixing this 💯

yijian166:
@UziTech sorry, wrong click ~~

PR #3424, opened 2024-08-21
feat: add provideLexer and provideParser hooks

### LINKED ISSUE #3315: Support for chunked rendering
Following the PR proposed in Marked-GFM-Headings: https://github.com/markedjs/marked-gfm-heading-id/pull/543, there is a notion of Marked sometimes being used in a "chunked" manner, i.e., parsing individual pages out of a longer document. This can be handy for live-editing services for example, where realtime updating of the rendered output may become slow for very large documents, instead re-parsing only the current page for speed.

Chunking can be done external to marked, by manually handling, say, an array of markdown source chunks and an array of output HTML chunks and updating each array entry as needed. However, difficulties arise when knowledge of the complete document is needed, for example for reflinks, where a link defined in a later chunk may be referenced in a prior one. The PR in Marked-GfM-Headings has a similar intent of ensuring globally-unique header IDs across chunks.

I don't know exactly how this would work (or if it is already possible), but I'm picturing some kind of method of marking certain data as persistent across chunks, with some knowledge of which chunk it came from so it can be updated as needed. Marked would also need some way to communicate which chunk is currently being rendered.

For example: the lexer.tokens.links object would need an option to be made persistent across chunks, and perhaps label each link entry with a chunk ID so if that same chunk is re-rendered (perhaps a reflink is removed), it can be properly updated for the global list.

It would be nice to have a set template for this type of object so extensions, etc. can have a common interface for cross-chunk support.

UziTech:
**Marked version:** 14.0.0

## Description

Add `provideLexer` and `provideParser` hooks to allow extensions that need to change the lexer or parser. This will allow adding links to the Lexer for chunked rendering (#3315). This will also allow extensions to create a new Parser to return something other than an html string. The Parser could return dom elements or react/vue/angular components.

- Fixes #3315

## Contributor

## Committer

### DIFF
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -261,6 +261,8 @@ Hooks are methods that hook into some part of marked. The following hooks are av
 | `preprocess(markdown: string): string` | Process markdown before sending it to marked. |
 | `postprocess(html: string): string` | Process html after marked has finished parsing. |
 | `processAllTokens(tokens: Token[]): Token[]` | Process all tokens before walk tokens. |
+| `provideLexer(): (src: string, options?: MarkedOptions) => Token[]` | Provide function to tokenize markdown. |
+| `provideParser(): (tokens: Token[], options?: MarkedOptions) => Token[]` | Provide function to parse tokens. |
 
 `marked.use()` can be called multiple times with different `hooks` functions. Each function will be called in order, starting with the function that was assigned *last*.
 
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -262,7 +262,7 @@ Hooks are methods that hook into some part of marked. The following hooks are av
 | `postprocess(html: string): string` | Process html after marked has finished parsing. |
 | `processAllTokens(tokens: Token[]): Token[]` | Process all tokens before walk tokens. |
 | `provideLexer(): (src: string, options?: MarkedOptions) => Token[]` | Provide function to tokenize markdown. |
-| `provideParser(): (tokens: Token[], options?: MarkedOptions) => Token[]` | Provide function to parse tokens. |
+| `provideParser(): (tokens: Token[], options?: MarkedOptions) => string` | Provide function to parse tokens. |
 
 `marked.use()` can be called multiple times with different `hooks` functions. Each function will be called in order, starting with the function that was assigned *last*.
 
--- a/docs/USING_PRO.md
+++ b/docs/USING_PRO.md
@@ -327,6 +327,45 @@ console.log(marked.parse(`
 <img src="x">
 ```
 
+**Example:** Save reflinks for chucked rendering
+
+```js
+import { marked, Lexer } from 'marked';
+
+let refLinks = {};
+
+// Override function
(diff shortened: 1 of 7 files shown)


docs/USING_PRO.md
```diff
@@ -261,6 +261,8 @@ Hooks are methods that hook into some part of marked. The following hooks are av
 | `preprocess(markdown: string): string` | Process markdown before sending it to marked. |
 | `postprocess(html: string): string` | Process html after marked has finished parsing. |
 | `processAllTokens(tokens: Token[]): Token[]` | Process all tokens before walk tokens. |
+| `provideLexer(): (src: string, options?: MarkedOptions) => Token[]` | Provide function to tokenize markdown. |
+| `provideParser(): (tokens: Token[], options?: MarkedOptions) => Token[]` | Provide function to parse tokens. |
```
calculuschild, line 265: The output of a parser should be a string, no?

src/Instance.ts
```diff
@@ -291,8 +291,12 @@ export class Marked {
 
       if (opt.hooks) {
         opt.hooks.options = opt;
+        opt.hooks.block = blockType;
```
calculuschild, line 294: Is this value used anywhere?

src/Instance.ts
```diff
@@ -291,8 +291,12 @@ export class Marked {
...
       }
 
+      const lexer = opt.hooks ? opt.hooks.provideLexer() : (blockType ? _Lexer.lex : _Lexer.lexInline);
+      const parser = opt.hooks ? opt.hooks.provideParser() : (blockType ? _Parser.parse : _Parser.parseInline);
```
calculuschild, line 298: I may be misunderstanding here: This falls back to default lexer and parser if no hook is provided. But the default `provideLexer` and `provideParser` hooks also perform a fallback to the default lexer and parser. Is one of these fallbacks redundant?

docs/USING_PRO.md
calculuschild: Could we add an example showing how these might be used for chunking? For example preserving the `links` between renders? Or add it to the tests?

src/Instance.ts
UziTech: Yes technically the default hooks are redundant, that is so someone can just provide one hook and the rest of the hooks provide the default behavior. Currently hooks is null by default so we don't have to call the functions unnecessarily if no hooks are changed.

docs/USING_PRO.md
UziTech: Nice catch! 

The default return value is a string but really it can be anything. The output of this function is returned as the return value for `marked.parse` so if someone wanted to create a new parser to parse the markdown into react elements like [marked-react](https://www.npmjs.com/package/marked-react) They could return a react element.

docs/USING_PRO.md
```diff
@@ -325,6 +327,45 @@ console.log(marked.parse(`
 <img src="x">
 ```
 
+**Example:** Save reflinks for chucked rendering
```
calculuschild, line 330: *chunked

styfle (approved):

Issue #3487, opened 2024-10-13
Another bug about #2733


linuxbckp:
**Marked version:**
14.1.2
**Describe the bug**
In #2733, someone gave out a function walkTokens(token). But this workaroound has another bug, it will repeatedly change URL.

**To Reproduce**
For example, if a file is ./d.efg located at C:/abc/, 
if I opened ./d.efg once, the result will become C:/abc/d.efg (correct),
if I opened ./d.efg twice, the result will become C:/abc/C:/abc/d.efg (incorrect).

**Expected behavior**
marker.use() ( here is marker.use( { walkTokens(token) } ) ) should be non-state.

UziTech:
Can you create a repo that can reproduce this? 

I don't know what you mean by "opened ./d.efg twice" marked does not do anything different if you open a file.

linuxbckp:
Here is a repo: https://github.com/CNOCTAVE/marked_reader
Press Ctrl + Shift + i to open Electron console, then you can see errors.
Click the left top button, choose a Markdown file. Do it twice, then reproduce the errors.

linuxbckp:
In Markdown, ./-style relative path is common.  When in Electron(a Node.js-based framework), Electron always parse ./ path as Node path, so ./ path should be rewritten to absolute path.  This repo is a Markdown reader, user should open a local Markdown file.  To make that bug clearer, you can write a picture path as relative path, if picture cannot correctly shows out, that's bug!

linuxbckp:
Like:

d

![d.png](./d.png)

where this ./d.png is at C:/abc/

UziTech:
You are adding the walkTokens function everytime they open a file. You should only be adding it once:

```js
import { marked } from "marked";
function walkTokens(token) {
    if (token.type === 'image' || token.type === 'link') {
        token.href = markdownDir + ("/" + token.href).replace("//", "/").replace("\\", "/").replace("%5C", "/").replace("./", "/"); // 修正相对路径
    }
}
marked.use({ walkTokens });
```

or you can use an instance to not have the extensions be global

```js
import { Marked } from "marked";
...
ipcMain.on('render-markdown', (event, filePath) => {
    ...
    function walkTokens(token) {
        if (token.type === 'image' || token.type === 'link') {
            token.href = markdownDir + ("/" + token.href).replace("//", "/").replace("\\", "/").replace("%5C", "/").replace("./", "/"); // 修正相对路径
        }
    }
    const marked = new Marked();
[...]

PR #3492, opened 2024-10-17
Update USING_PRO.md

linuxbckp:
Add document about #3487

### DIFF
--- a/docs/USING_ADVANCED.md
+++ b/docs/USING_ADVANCED.md
@@ -13,6 +13,31 @@ const marked = new Marked([options, extension, ...]);
 |:--------|:-------|:----------------------------------------------------------------------|
 | options |`object`|The same arguments that can be passed to [`marked.use`](/using_pro#use)|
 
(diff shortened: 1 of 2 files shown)


linuxbckp:
OK. I changed to Marked part.

docs/USING_ADVANCED.md
```diff
@@ -13,6 +13,31 @@ const marked = new Marked([options, extension, ...]);
 |:--------|:-------|:----------------------------------------------------------------------|
 | options |`object`|The same arguments that can be passed to [`marked.use`](/using_pro#use)|
 
+Another case is, if you want to rerun function everytime they open a file, or you want to make marked.run() non-state, you can create a new instance of Marked to ensure options and extensions are locally scoped.
```
UziTech, line 16: It doesn't seem like this is saying anything more than the documentation above. That the user should use Marked instance if they want extensions to be locally scoped.

I think the example should emphasize that `marked.use(...)` should not be used in a loop or function. It should only be used directly after `new Marked` is created or `marked` is imported

UziTech (changes requested):

linuxbckp:
OK. I changed to a note.

UziTech (approved):
Thanks! 💯

PR #3527, opened 2024-11-09
Remove unused plus typescript tightening

SpencerWhitehead7:
## Description

This is a bunch of fairly minor changes to remove unused files/packages and take advantage of some more recent typescript/js features, namely `?:` in parameter definitions, `?.` optional chaining, at `.at(-x)` instead of `[arr.length - x]`

There should not be any changes in functionality or type interfaces.

It will make the most sense looked one commit at a time.

## Contributor

## Committer


UziTech (commented):
Some of these changes could make marked slower like declaring a variable inside a loop instead of before the loop. I'm not sure if this is still the case but it would be nice to get an idea of how this change affects speed.

UziTech:
In terms of speed it looks like this may be faster. Running `npm run bench` gives us a benchmark for marked against commonmark and master is giving me ~35% slower than commonmark and this PR is giving ~30% slower.

src/MarkedOptions.ts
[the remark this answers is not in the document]
UziTech: Can you add some tests to `test\types\marked.ts` make sure undefined can still be passed in

Something like:

```js
marked.parse('', {
  async: undefined,
  breaks: undefined,
  extensions: undefined,
  gfm: undefined,
  hooks: undefined,
  pedantic: undefined,
  renderer: undefined,
  silent: undefined,
  tokenizer: undefined,
  walkTokens: undefined,
});

marked.use({
  async: undefined,
  breaks: undefined,
  extensions: undefined,
  gfm: undefined,
  hooks: undefined,
  pedantic: undefined,
  renderer: undefined,
  silent: undefined,
  tokenizer: undefined,
  walkTokens: undefined,
});
```

bower.json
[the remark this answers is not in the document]
UziTech: I guess we can always add it back if it is needed.

docs/demo/demo.js
```diff
@@ -185,7 +185,7 @@ function setOptions(opts) {
   $optionsElem.value = JSON.stringify(
     opts,
     (key, value) => {
-      if (value && typeof value === 'object' && Object.getPrototypeOf(value) !== Object.prototype) {
```
UziTech, line 188: I think this is needed since `null` will throw an error in `Object.getPrototypeOf(null) !== Object.prototype`

You can change it to `value !== null &&` if you want

UziTech (changes requested):

SpencerWhitehead7:
> In terms of speed it looks like this may be faster with Node v22. Running `npm run bench` gives us a benchmark for marked against commonmark and master is giving me ~35% slower than commonmark and this PR is giving ~30% slower.

I still have the bench-100 file from this PR https://github.com/markedjs/marked/pull/3146. It says main takes 122ms and this branch takes 117, which I think is kinda within margin of error. It's not a big degradation at least.

docs/demo/demo.js
SpencerWhitehead7: ah, nice catch, done

src/MarkedOptions.ts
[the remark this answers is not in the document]
SpencerWhitehead7: sure, done

src/Hooks.ts
```diff
@@ -6,7 +6,7 @@ import type { Token, TokensList } from './Tokens.ts';
 
 export class _Hooks {
   options: MarkedOptions;
-  block: boolean | undefined;
+  block?: boolean;
```
styfle, line 9: Note that these do not have the same meaning.

https://x.com/DavidKPiano/status/1856701854102557026

Although I think in this case that's okay.

src/Tokenizer.ts
```diff
@@ -414,8 +414,11 @@ export class _Tokenizer {
       }
 
       // Do not consume newlines at end of final item. Alternatively, make itemRegex *start* with any newlines to simplify/speed up endsWithBlankLine logic
-      list.items[list.items.length - 1].raw = list.items[list.items.length - 1].raw.trimEnd();
-      list.items[list.items.length - 1].text = list.items[list.items.length - 1].text.trimEnd();
+      const lastItem = list.items.at(-1);
+      if (lastItem) {
```
styfle, line 418: Thanks for making this one safe 👍

src/helpers.ts
```diff
@@ -72,7 +72,7 @@ export function splitCells(tableRow: string, count?: number) {
   if (!cells[0].trim()) {
     cells.shift();
   }
-  if (cells.length > 0 && !cells[cells.length - 1].trim()) {
+  if (!cells.at(-1)?.trim()) {
```
styfle, line 75: This looks like a bug.

Consider the case when `cells = []`.

```
console.log(cells.length > 0 && !cells[cells.length - 1].trim()); // false
console.log(!cells.at(-1)?.trim()) // true
```

SpencerWhitehead7: Thanks for the catch. Since all the unit tests still pass, I wonder if for some reason cells.length always happens to be > 0  on this line. I'll fix it though.

styfle (approved):

PR #3528, opened 2024-11-09
chore: fix update-specs script, use builtin fetch

SpencerWhitehead7:
## Description

Fixes the update specs script being broken, removes unnecessary dependency

In master, when you run update specs, it fails to update commonmark specs.

This is because the place it pulls from, `https://spec.commonmark.org/<version>/spec.json` seems like it doesn't accept versions numbers without a patch anymore, and the current script sends a request for `major.minor`. It also doesn't accept requests for `major.minor.0`, or, as far as I could tell, anything except the current version. Therefore, we can just use the version from commonmark's package.json instead of lowering it to `major.minor`.

This does mean the resulting spec file gets renamed, but I think having the full version is better if we're actually pulling from the full version.

The script was also using node-fetch, but fetch has existed natively in node since (I think) node 18 so it's within marked's support window.

-->

## Contributor

## Committer


UziTech (approved):
Thanks! 💯

PR #3584, opened 2025-01-12
fix: add missing TableCell token to MarkedToken

MrMarble:
**Marked version:** v15.0.6

**Markdown flavor:** n/a

## Description

- Fixes `TableCell` not exported as MarkedToken

## Contributor

## Committer

### DIFF
--- a/src/Tokens.ts
+++ b/src/Tokens.ts
@@ -20,6 +20,7 @@ export type MarkedToken = (
   | Tokens.Space
   | Tokens.Strong
   | Tokens.Table
+  | Tokens.TableCell
   | Tokens.Tag
   | Tokens.Text
 );


UziTech:
TableCell is not like the other tokens, it does not have `type` or `raw` properties, which is why the tests are failing.

Is there a reason you need TableCell exported in MarkedToken?

Checkbox and TableRow are similar tokens that are not exported in MarkedToken.

MrMarble:
When creating an extension that overrides the renderer, you might get a type error because Tokens.TableCell does not satisfy Token.
Honestly I thought it was a mistake that the type was left over, if it was on purpose then this can be closed

UziTech:
Can you give an example of an extension that doesn't work? I think it would be fine to add all of the tokens to MarkedToken but it might be a lot of work.

UziTech:
Looks like MarkedToken is only used in the Parser where the token won't ever be a TableCell

https://github.com/markedjs/marked/blob/5624378bd3dcc44f71dd0568deffe7e85f6f12c9/src/Parser.ts#L58

Anywhere where we need any token including those not run through the Parser we use `Token` which is any token type.

UziTech:
Closing this as stale. If you would like to continue this PR feel free to reopen

PR #3593, opened 2025-01-16
add support for strong tokens

### LINKED ISSUE #3592: error: Token with "strong" type was not found.
Run this code
```
import { type RendererObject, type Tokens, marked } from "marked";

const customRenderer: RendererObject = {
	heading(tokens: Tokens.Heading): string {
		const text = this.parser.parse(tokens.tokens);
		return `*${text.trim()}*\n`;
	},
};

const text = "### **Hello world**";

marked.use({ renderer: customRenderer });
const r = await marked.parse(text);
console.log(r);
```

Result:
```
$ bun run index.ts
2062 |                     if (this.options.silent) {
2063 |                         console.error(errMsg);
2064 |                         return '';
2065 |                     }
2066 |                     else {
2067 |                         throw new Error(errMsg);
                                     ^
error: Token with "strong" type was not found.
```

This is marked v15.0.6, ran it using bun.

preslavmihaylov:
Resolves #3592

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -90,6 +90,10 @@ export class _Parser {
           out += this.renderer.html(token);
           continue;
         }
+        case 'strong': {
+          out += this.renderer.strong(token);
+          continue;
+        }
         case 'paragraph': {
           out += this.renderer.paragraph(token);
           continue;


UziTech:
strong is an inline token not a block token. There are subtle differences that make adding it to the block tokens a bad idea.

PR #3594, opened 2025-01-18
feat: tokenizer extension position

### LINKED ISSUE #3590: Select Extension priority
**Describe the feature**
Custom extensions always take priority over default tokens. A way to specify where in the parsing sequence an extension should run so certain default tokens don't get overridden.

Sometimes there are ambiguous cases in Markdown where the result depends on which token gets parsed first. Example, tables without a starting pipe:

```
<div> header
|:---:|
cell
```

This is not be parsed as a table, because HTML is parsed first. However if someone writes a custom table extension for example, suddenly tables take priority and this is rendered as a table. The user would want the extension to execute before the default table tokenizer but after the HTML tokenizer.

It *is* possible to work around this by, in the extension tokenizer, pre-parsing the string and checking if any higher-priority tokens appear. However this adds extra steps and results in the same line potentially being parsed multiple times.

I'm picturing something like breaking the `parse()` and `parseInline()` functions into an array of tokenizers and calling each one in sequence, and a user making an extension would somehow be able to inject their extension into this array (added to the front by default), though I imagine calling each tokenizer from an array instead would cause some slowdown.

UziTech:
**Marked version:** 15.0.6

## Description

Add position property for tokenizer extension to it can pisk where to be ran in the lexer.

- Fixes #3590

### TODO:

- [ ] Check benchmark speed (initial check seems to be fine)
- [ ] Write tests
- [ ] Write docs

## Contributor

## Committer

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -6,7 +6,7 @@ import { _Renderer } from './Renderer.ts';
 import { _Tokenizer } from './Tokenizer.ts';
 import { _TextRenderer } from './TextRenderer.ts';
 import { escape } from './helpers.ts';
-import type { MarkedExtension, MarkedOptions } from './MarkedOptions.ts';
+import { tokenizerBlockPositions, tokenizerInlinePositions, type MarkedExtension, type MarkedOptions } from './MarkedOptions.ts';
 import type { Token, Tokens, TokensList } from './Tokens.ts';
 
 export type MaybePromise = void | Promise<void>;
@@ -74,7 +74,7 @@ export class Marked {
   }
 
   use(...args: MarkedExtension[]) {
-    const extensions: MarkedOptions['extensions'] = this.defaults.extensions || { renderers: {}, childTokens: {} };
+    const extensions: MarkedOptions['extensions'] = this.defaults.extensions || { renderers: {}, childTokens: {}, tokenizers: {} };
 
     args.forEach((pack) => {
       // copy options to new object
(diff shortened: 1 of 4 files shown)


calculuschild:
I haven't run any benchmarks on this, but I'm curious how much of a slowdown this adds, assuming no extensions. Do you have any measurements?

As an alternative that might be more flexible, but more restructuring... I wonder what you think about extracting each of the parser/renderer steps into an array of "extension-like objects".

I.e., instead of:

```
      beforeCode();
      // code
      if (token = this.tokenizer.code(src)) {
       ...
       beforeFences();

       // fences
      if (token = this.tokenizer.fences(src)) {
      ... 
```

something like (pseudocode):

```
const tokenizers = {
  code : ()=>{ if (token = this.tokenizer.code(src)) ... },
  fences : ()=>{ if (token = this.tokenizer.fences(src)) ... },
  ...
}

for ([tokenizerName, tokenizerFunction] in tokenizers) {
  runExtensionBefore(tokenizerName);
  tokenizerFunction();
}
```

I'm fairly certain there is some speed impact calling the tokenizers from an array/map/object like this but it might make the whole project more flexible for this type of "extension position" customization.

UziTech:
I did try to move the tokenizers to some sort of array but there were two problems.

1. Some tokenizers require a lot of extra logic that is not easy to move out of the lexer.
2. I couldn't think of an easy way to provide the array to the user so they can pick where to put their tokenizers

I think just adding a `position` property like this is the easiest for the user, and even though it is a lot of boilerplate we are not likely going to be adding or removing tokenizers from the lexer.

src/Instance.ts
```diff
@@ -105,23 +105,26 @@ export class Marked {
...
             }
-            const extLevel = extensions[ext.level];
-            if (extLevel) {
-              extLevel.unshift(ext.tokenizer);
-            } else {
-              extensions[ext.level] = [ext.tokenizer];
+            if (ext.position && ![...tokenizerBlockPositions, ...tokenizerInlinePositions].includes(ext.position)) {
+              throw new Error(`extension position must be one of '${tokenizerBlockPositions.join("', '")}',  '${tokenizerInlinePositions.join("', '")}'`);
+            }
+            if (!ext.level && !ext.position) {
```
calculuschild, line 114: Should we enforce compatible `level` and `position` entries? I.e., a `block` level doesn't make sense with an `inline` position.  Or, should these properties be mutually-exclusive to avoid unexpected behavior?

calculuschild (changes requested):
Implementation seems good to me assuming there's no alternative to the boilerplate.

Remaining comments:
1) Should this include a documentation update?
2) Does this need a test for an extension at one of the custom positions?
3) Is there any significant performance impact? Not sure if the extra function calls between each step are going to be an issue.

calculuschild:
> I did try to move the tokenizers to some sort of array but there were two problems.

Ok, fair enough. Looking back, this is what I tried to do back in 2021 and I think we concluded arrays just add too much slowdown anyway. https://github.com/markedjs/marked/pull/1872

UziTech:
> Remaining comments:

Ya, these 3 are in the TODOs section in this PRs description. I will work on them soon.

> Also looking back, we had discussed a "before" parameter 4 years ago. Hoping it wasn't also a lag issue that made us drop it

Looks like we just figured we would get it done when it was actually needed

src/Instance.ts
UziTech: They are mutually exclusive. They can both be specified in case an extension wants to support many versions of marked, but versions of marked that allow position will ignore level if position is provided.

UziTech:
After months of trying different things to see if there is a way this won't allow down marked I haven't been able to find one.

The only thing I can think of doing is making this an extension that uses the `provideLexer` hook to use a different lexer that changes the order. That way it won't slow down default marked and someone can opt in if they are ok with the slowdown.

PR #3661, opened 2025-04-16
fix: expose Lexer.rules.other

MichaelDoyle:
**Marked version:**

**Markdown flavor:** n/a

## Description

I'd like to extend the renderer, and I noticed there is no access to the "other" regexes.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -60,6 +60,7 @@ export class _Lexer {
    */
   static get rules() {
     return {
+      other,
       block,
       inline,
     };


UziTech:
Can you add a comment that these can have breaking changes in minor versions?

There have been many times we have needed to fix a bug by changing the number of capture groups in the regexps. This can cause breaking changes if the code that uses the regexp doesn't change as well.

It is usually better for that reason to just copy the regexp if you need to use it outside of marked.

MichaelDoyle:
I went a different direction for this -- your feedback makes sense.

PR #3670, opened 2025-04-25
fix: use esbuild for accurate sourcemaps

UziTech:
## Description

Sourcemaps don't work currently because of some strange thing between rollup and typescript.

This PR removes rollup and replaces it with esbuild. The sourcemaps work!

## Contributor

## Committer


styfle (approved):
Sounds good!

I also think we should consider releasing a semver major and drop support for CJS because all LTS versions of Node.js support `require(esm)`

https://nodejs.org/docs/latest-v20.x/api/process.html#processfeaturesrequire_module

We could also drop support for Node.js 18 since we still have it in package.json engines.node even though its been EOL since 2025-04-30.

https://github.com/nodejs/Release/blob/main/README.md#end-of-life-releases

Although technically Node.js 18 can load ESM just fine, but you would need `await import(esm)` from a CJS file.

UziTech:
Ya I think we can do that after this fix. We should just have two versions: esm and umd, and minify both. Now that the source maps work I don't see any reason not to minify the built files.

PR #3671, opened 2025-04-26
Fix: skip inline parsing when block tokens already parsed

Hongweiwa:
Marked version:
latest (master branch at commit [commit hash] if you know it)

Markdown flavor:
GitHub Flavored Markdown

Description
Improves the parsing behavior inside code blocks.

Prevents unnecessary inline parsing (such as emphasis, links, or special characters like @) inside fenced and indented code blocks.

No related GitHub issue — this is a proactive improvement based on observing markdown rendering inconsistencies.

Expectation
Inside code blocks (fenced by ``` or indented by 4 spaces), content should be treated as plain text,
without interpreting inline syntax such as @, _, *, etc.

Result
Before fix:

In some situations, special characters inside code blocks were incorrectly parsed.

After fix:

Code blocks now properly preserve all content literally, matching markdown specs.

What was attempted
Updated lexer and tokenizer logic to skip inline tokenization inside Tokens.Code nodes

This fix prevents Tokens.Code nodes from being parsed for inline formatting.
Code blocks must render content literally without interpreting @, *, or _ as special syntax.
This makes the parser behave according to markdown standards and avoids unexpected formatting issues.

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -299,6 +299,12 @@ export class _Lexer {
     let maskedSrc = src;
     let match: RegExpExecArray | null = null;
 
+    //Skip inline parsing for fenced code blocks
+  if ((tokens as any)?.isInCodeBlock) {
+    return tokens;
+  }
+
+
     // Mask out reflinks
     if (this.tokens.links) {
       const links = Object.keys(this.tokens.links);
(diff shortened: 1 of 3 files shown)


UziTech:
Would you be able to do the same thing by checking if `token.type === 'code'`?

Can you provide some markdown where marked renders something inside code blocks?

UziTech:
closing as stale. If you would like to continue working on this feel free to reopen it.

PR #3687, opened 2025-05-18
BREAKING: remove cjs build & update min node to 20

UziTech:
## Description

builds off of #3670

### BREAKING CHANGES

- remove `./marked.min.js` use `./lib/marked.umd.js` instead
- remove `./lib/marked.cjs` 
- update minimum supported node version to 20 to support `require('marked.esm.js')`. see https://nodejs.org/docs/latest-v20.x/api/process.html#processfeaturesrequire_module

## Contributor

## Committer


UziTech:
test failure should be fixed by https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/248

docs/demo/worker.js
```diff
@@ -152,6 +153,7 @@ function loadVersion(ver) {
   }
   return promise.then(() => {
     currentVersion = ver;
+    console.log(`Using version '${ver}'`);
```
styfle, line 156: Might not need this

test/bench.js
styfle, line 1: How was the performance here?

Curious if there is any difference between mode versions.

UziTech: cjs was always slightly slower than esm. Not by much though.

styfle (approved):
Thanks!

UziTech:
I think this is ready to go, but I am going to wait to merge this for a month so we can bundle any other breaking changes together.

PR #3722, opened 2025-07-05
fix: add generic types for parser and renderer output

UziTech:
**Marked version:** 16.0.0

## Description

Add Generic types for parser and renderer output.

## Contributor

## Committer

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -7,39 +7,39 @@ import type { MarkedOptions } from './MarkedOptions.ts';
 /**
  * Parsing & Compiling
  */
-export class _Parser<P = string, R = string> {
-  options: MarkedOptions<P, R>;
-  renderer: _Renderer<P, R>;
-  textRenderer: _TextRenderer<R>;
-  constructor(options?: MarkedOptions<P, R>) {
+export class _Parser<ParserOutput = string, RendererOutput = string> {
+  options: MarkedOptions<ParserOutput, RendererOutput>;
+  renderer: _Renderer<ParserOutput, RendererOutput>;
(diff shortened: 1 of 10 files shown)


src/Parser.ts
```diff
@@ -96,10 +96,10 @@ export class _Parser {
         }
         case 'text': {
           let textToken = token;
-          let body = this.renderer.text(textToken);
+          let body = this.renderer.text(textToken) as string;
```
styfle, line 99: This `as string` seems like the types are wrong. Is there a way to improve it so we don't need the assertion 🤔

src/defaults.ts
```diff
@@ -17,9 +17,9 @@ export function _getDefaults(): MarkedOptions {
     walkTokens: null,
   };
 }
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export let _defaults: MarkedOptions<any, any> = _getDefaults();
```
styfle, line 21: Why `any`? Is that because we don't want to pass in generic types to _getDefaults()?

UziTech: If we mark this as anything else then _defaults would have to stay that type. But because we can specify different types for some options this has to be any because it can be anything.

src/Parser.ts
UziTech: This is needed because the renderer is generic so at this point `this.renderer.test()` could return any type. We are assuming that the RendererOutput applies to all renderer functions so the user will have to change every function (including this one) if they want to return something other than a string. The correct way to type this would be to have a generic param for every renderer function. I feel like that is a bit overkill.

styfle (approved):

PR #3745, opened 2025-08-13
feat: add def token in tokens and renderer

UziTech:
**Marked version:** 16.1.2

## Description

Add definition token and renderer as an empty string.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -206,6 +206,7 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
             href: token.href,
             title: token.title,
           };
+          tokens.push(token);
         }
         continue;
       }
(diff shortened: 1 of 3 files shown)


styfle (approved):

calculuschild:
This looks straightforward enough.

My one comment is if this would work better using the preprocess hook or something.

Defs can occur after the line where the corresponding link is rendered. This currently works fine because Defs sre parsed first as part of the "block" pass, and links later during the "inline" pass.

However if this is to support extensions, I could see users wanting to have access to the set of Def tokens as a precursor to everything everything else so they could also be referenced in any block-level extensions. (I actually do this myself for one of my extensions that extends links) That would be a bigger refactor though.

UziTech:
Ya that would be a much bigger refactor. Def tokens are already parsed they just aren't added to the tokens and there is no renderer for them (since they usually aren't rendered anyways)

PR #3761, opened 2025-09-10
fix: tokens.at is not a function

### LINKED ISSUE #3546: [bug] `tokens.at` is not a function
**Marked version:**
15.0.2

**Describe the bug**
After upgrading from v14, our app started getting error reports with the following text:

```
t.at is not a function. (In 't.at(-1)', 't.at' is undefined)
```

... so I am doing as instructed 😛 

The immediate stack trace reads:

```
assets/node_modules/marked/lib/marked.umd.js:1522:0:in `at'
```

```js
                }
                if (this.state.top && (token = this.tokenizer.paragraph(cutSrc))) {
                    const lastToken = tokens.at(-1);
                    if (lastParagraphClipped && lastToken?.type === 'paragraph') {
                        lastToken.raw += '\n' + token.raw;
```

**To Reproduce**
Unclear as this only arose through our error logging. We ended up rolling back the `marked` upgrade, but if/when we find out more information about a specific reproduction path, we will update.

However our logging indicates this error only occurs on a limited range of older browser versions, including Safari 15.3 and earlier, and Chrome 125 and earlier, and it [appears](https://caniuse.com/mdn-javascript_builtins_array_at) the `.at()` method is not supported on Safari before v15.4 (though supposedly it's supported in Chrome 92+ making the error there a bit more curious). It's not clear whether Marked still supports these versions as the readme currently only says "not IE11".

**Expected behavior**
The parsing works as expected / without error.

xingyixiang:
**Marked version: 16.2.1

## Description

- Fixes #3546

The Array.prototype.at() method, introduced in ES2022, provides a concise way to access array elements using both positive and negative indices.

**‌Modern Browsers‌:**
Fully supported in Chrome 92+, Firefox 90+, Safari 15.4+, and Edge 92+

## Expectation
I hope the official can be compatible with lower versions of browsers. 🙏

## Contributor

## Committer

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2022",
+    "target": "es2019",
     "module": "NodeNext",
     "isolatedModules": true,
     "strict": true,


UziTech:
We cannot support older browsers because we cannot ensure they won't be broken by another PR in the future. In order for us to support older browsers we would need to setup automated tests on older browsers. If you would like to setup and maintain that test framework then I am ok with supporting older browsers, but until then this change wouldn't help.

xingyixiang:
@UziTech 
Thank you, it really makes sense.  I have modified **tsconfig**, and the minimum requirement is to support **ES2019**

xingyixiang:
Description of compatibility in **README**



This revision enhances the project's compatibility and makes it more stable and reliable.

UziTech:
Unless we can test this on actual browsers all this does is increase the size of marked npm package.

The best fix for this is to use polyfills when needed. See https://github.com/markedjs/marked/issues/3546#issuecomment-2545886874

xingyixiang:
> Unless we can test this on actual browsers all this does is increase the size of marked npm package.
> 
> The best fix for this is to use polyfills when needed. See [#3546 (comment)](https://github.com/markedjs/marked/issues/3546#issuecomment-2545886874)

This modification did not increase the size of marked npm package.

After I changed it to `"target": "es2019"`, there is only one compatibility issue with `Arrays.prototype.at`. 

It would be better if the official makes compatibility modifications, so that most browsers can achieve out-of-the-box effects, and some application scenarios are not convenient to use polyfills

UziTech:
> It would be better if the official makes compatibility modifications

I think this is where we disagree. There is a tradeoff between being as fast as we can and supporting older browsers. Not all marked users need old browsers. Marked should be fast and small by default since there are ways to support older browsers if needed with polyfills and babel. If we optimize for older browsers there is no way (that I know of) to automatically make it faster for newer browsers.

Unfortunately we can't be great for everyone by default. We try to support most users. It would be great if someone would release a package that takes marked and polyfills for older browsers and packages them together for people to use. Unfortunately I have not found a person who has time to maintain something like that.

PR #3769, opened 2025-09-26
fix: Handle case-insensitive `www.` prefix

vaskkey:
**Marked version:**

16.3.0

## Expectation

When a link starts with `www.` (in any case combination like `Www.`, `WWW.`, etc.), it should be automatically converted to a clickable link with `http://` prepended.

## Result

Before the fix, only lowercase `www.` was being prepended with `http://`. Links starting with `Www.` or other case variations were being processed as autolinks but did not have `http://` prepended, which caused the links to redirect to `http://mainwebsite/Www.foo.com`.

After the fix, both `www.foo.com` and `Www.foo.com` are now properly converted to `<a href="http://www.foo.com">www.foo.com</a>` and `<a href="http://Www.foo.com">Www.foo.com</a>` respectively.

## What was attempted

The fix was implemented in `src/Tokenizer.ts:843` by changing the strict equality check from:
```ts
if (cap[1] === 'www.') {
```
to a case-insensitive comparison:
```ts
if (cap[1]?.toLowerCase() === 'www.') {
```

## Contributor

- [X] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR): Added cases to autolinks.md

## Committer

### DIFF
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -840,7 +840,7 @@ export class _Tokenizer<ParserOutput = string, RendererOutput = string> {
           cap[0] = this.rules.inline._backpedal.exec(cap[0])?.[0] ?? '';
         } while (prevCapZero !== cap[0]);
         text = cap[0];
-        if (cap[1] === 'www.') {
+        if (cap[1]?.toLowerCase() === 'www.') {
           href = 'http://' + cap[0];
         } else {
           href = cap[0];
--- a/test/specs/new/autolinks.html
+++ b/test/specs/new/autolinks.html
@@ -8,6 +8,10 @@
 
 <p><a href="hTtP://fOo.CoM">hTtP://fOo.CoM</a></p>
 
+<p><a href="http://www.foo.com">www.foo.com</a></p>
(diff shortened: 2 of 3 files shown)


UziTech:
[Autolink](https://github.github.com/gfm/#autolinks-extension-) is a GFM extension so we want it to act the same as GitHub and it looks like GitHub doesn't allow the www to be case insensitive.

www.example.com
Www.example.com
WWW.example.com

If this is something you want you can create a [custom extension](https://marked.js.org/using_pro#extensions)

vaskkey:
@UziTech In that case I could probably configure the autolink to be a bit stricter than it is right now. because currently for Www.example.com `marked` generates `<a href="Www.example.com">Www.example.com</a>` which is neither a good link nor a GitHub way of parsing links

UziTech:
Good call. Ya that sounds good. We want it to be the same as GitHub

vaskkey:
Closing this PR, Github-like implementation here: https://github.com/markedjs/marked/pull/3770

PR #3817, opened 2025-10-28
fix: Avoid RegExp lookbehind assertions (#3816)

### LINKED ISSUE #3816: Marked is incompatible with Safari < 16.5
`marked` is incompatible with Safari on iOS/iPadOS prior to 16.5 due to a `SyntaxError`.

The issue stems from two instances of **RegExp negative lookbehind assertions (`?<!...`)** within the `blockSkip` constant.

These appear to be the only instances of this incompatible syntax in the entire codebase. Because the issue is isolated to this single constant, a targeted fix could restore compatibility for a large number of users.

The problematic lines are within the `link` and `code` replacements:
```javascript
const blockSkip = edit(/link|code|html/, 'g')
  .replace('link', /\[(?:[^\[\]`]|(?<!`)(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:...)\)/) // <-- OFFENSE #1
  .replace('code', /(?<!`)(?<b>`+)[^`]+\k<b>(?!`)/)                                // <-- OFFENSE #2
  .replace('html', /<(?! )[^<>]*?>/)
  .getRegex();
```

Thank you for your consideration and for maintaining this library.

ziadkh0:
RegExp lookbehind assertions are supported in Safari >= 16.5, avoiding them keeps support for older Safari versions.

**Marked version:** 16.4.1

**Markdown flavor:** n/a

## Description

- Fixes #3816

## Contributor

## Committer

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -21,6 +21,8 @@ try {
   // eslint-disable-next-line prefer-regex-literals
   return !!new RegExp('(?<=1)(?<!1)');
 } catch {
+  // See browser support here:
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion
   return false;
 }
 })();


src/rules.ts
```diff
@@ -267,9 +267,9 @@ const _punctuationOrSpaceGfmStrongEm = /(?!~)[\s\p{P}\p{S}]/u;
 const _notPunctuationOrSpaceGfmStrongEm = /(?:[^\s\p{P}\p{S}]|~)/u;
 
 // sequences em should skip over [title](link), `code`, <html>
-const blockSkip = edit(/link|code|html/, 'g')
-  .replace('link', /\[(?:[^\[\]`]|(?<!`)(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)/)
-  .replace('code', /(?<!`)(?<b>`+)[^`]+\k<b>(?!`)/)
+const blockSkip = edit(/link|code|html/, 'dg')
+  .replace('link', /\[(?:[^\[\]`]|(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)/)
+  .replace('code', /(?:^^|[^`])((?<b>`+)[^`]+\k<b>(?!`))/)
```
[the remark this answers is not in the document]
ziadkh0: `^^` is required because `edit().replace()` remove one `^`.
And the `^` is needed to handle the case of a line starting with backticks.

src/rules.ts
```diff
@@ -267,9 +267,9 @@ const _punctuationOrSpaceGfmStrongEm = /(?!~)[\s\p{P}\p{S}]/u;
 const _notPunctuationOrSpaceGfmStrongEm = /(?:[^\s\p{P}\p{S}]|~)/u;
 
 // sequences em should skip over [title](link), `code`, <html>
-const blockSkip = edit(/link|code|html/, 'g')
-  .replace('link', /\[(?:[^\[\]`]|(?<!`)(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)/)
-  .replace('code', /(?<!`)(?<b>`+)[^`]+\k<b>(?!`)/)
+const blockSkip = edit(/link|code|html/, 'dg')
+  .replace('link', /\[(?:[^\[\]`]|(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)/)
```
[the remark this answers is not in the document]
ziadkh0: It appears you are hallucinating 😕

ziadkh0:
I noticed `npm run test:redos` uses `recheck` which doesn't appear to support the `d` flag.
Should I update `test/recheck.js` to ignore this flag?

UziTech:
ya updating recheck.js would be helpful. I don't think the d flag would change whether or not the regexp is vulnerable to redos.

UziTech:
the full regexp looks safe

```
/\[(?:[^\[\]`]|(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)|(?:^|[^`])((?<b>`+)[^`]+\k<b>(?!`))|<(?! )[^<>]*?>/dg
```

UziTech:
It looks like this does slow marked down significantly.

`npm run bench` for this PR:
```
marked completed in 5407ms and passed 97.70%
commonmark completed in 3798ms and passed 100.00%
markdown-it completed in 4457ms and passed 88.96%
42% slower than commonmark
```

`npm run bench` for master:
```
marked completed in 5102ms and passed 97.70%
commonmark completed in 3787ms and passed 100.00%
markdown-it completed in 4452ms and passed 88.96%
35% slower than commonmark
```

UziTech:
@styfle @calculuschild is it worth it to slow marked down to support ~0.25% of users?

I'm on the fence

image

calculuschild:
I would say not worth it. The core marked should prioritize speed for the 99.75% of users.

If somebody really needs to support old browsers, I recommend an older version or making a fork.

ziadkh0:
I can feature detect support and make the lookbehind-less version a fallback, that way it's win-win. What do you think?

ziadkh0:
Only marginal difference on a M1 MacBook Air. 🤷‍♂️

`npm run bench` for this PR:
```
marked completed in 1911ms and passed 97.70%
commonmark completed in 1410ms and passed 100.00%
markdown-it completed in 1670ms and passed 88.96%
36% slower than commonmark
```

`npm run bench` for master:
```
marked completed in 1914ms and passed 97.70%
commonmark completed in 1407ms and passed 100.00%
markdown-it completed in 1678ms and passed 88.96%
36% slower than commonmark
```

ziadkh0:
I'm getting big run to run variance when running `npm run bench` 😢
@UziTech are your bench results consistent from run to run?

UziTech:
The times are not consistent due to other things taking up the processor during part of it but the time for marked relative to the other in one run should be fairly consistent. Mine is usually within 2% so a 7% change seems significant

ziadkh0:
I have seen a difference of 10% between slowest run and fastest run (e.g. `36% slower than commonmark` and `46% slowerthan commonmark`)!

UziTech:
> I can feature detect support and make the lookbehind-less version a fallback, that way it's win-win. What do you think?

That seems like unnecessary maintenance burden to make sure the two version of the regexp + code are in sync.

ziadkh0:
For what it's worth, in the usage stats you forgot to include iOS Safari ~1% of users

Image

calculuschild:
Any chance this could just be made as an extension?

ziadkh0:
I ran `npm run bench` for master on an idling machine 11 times and I still got a big variance from run to run

```
41% slower than commonmark
47% slower than commonmark
51% slower than commonmark
41% slower than commonmark
47% slower than commonmark
38% slower than commonmark
36% slower than commonmark
31% slower than commonmark
43% slower than commonmark
40% slower than commonmark
42% slower than commonmark
```

Am I the only one unable to get a consistent run to run results?

ziadkh0:
I made the lookbehind-less version a fallback for engines that don't support them, and changed the implementation a bit to improve performance.

@UziTech, can you please rerun the benchmark on the most recent commit? 🤞

As for the maintenance burden, I'm willing to maintain support for Safari 15.6. 😄

src/rules.ts
```diff
@@ -16,6 +16,16 @@ function edit(regex: string | RegExp, opt = '') {
   return obj;
 }
 
+const supportsLookbehind = (() => {
+try {
+  // eslint-disable-next-line prefer-regex-literals
+  return !!new RegExp('(?<=1)(?<!1)');
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+} catch(error) {
```
UziTech:
```suggestion
} catch {
```

You should be able to remove the catch argument if it is not used

ziadkh0: Good catch! 😉

ziadkh0:
@UziTech Thank you for testing and approving this PR! 😄

1. Would you prefer I squash the two commits into a single one? (I'm not quite sure what the commit guidelines are 😅)
2. The new version no longer uses the `d` RegExp flag, should I update `recheck.js` in another PR or there's no need?

UziTech:
1. We use squash and merge whenever we merge a PR. You don't need to do anything. 
2. No need, we can do it later if needed.

src/rules.ts
```diff
@@ -267,9 +276,10 @@ const _punctuationOrSpaceGfmStrongEm = /(?!~)[\s\p{P}\p{S}]/u;
 const _notPunctuationOrSpaceGfmStrongEm = /(?:[^\s\p{P}\p{S}]|~)/u;
 
 // sequences em should skip over [title](link), `code`, <html>
-const blockSkip = edit(/link|code|html/, 'g')
-  .replace('link', /\[(?:[^\[\]`]|(?<!`)(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)/)
-  .replace('code', /(?<!`)(?<b>`+)[^`]+\k<b>(?!`)/)
+const blockSkip = edit(/link|precode-code|html/, 'g')
+  .replace('link', /\[(?:[^\[\]`]|(?<a>`+)[^`]+\k<a>(?!`))*?\]\((?:\\[\s\S]|[^\\\(\)]|\((?:\\[\s\S]|[^\\\(\)])*\))*\)/)
+  .replace('precode-', supportsLookbehind ? '(?<!`)()' : '(^^|[^`])')
```
styfle, line 283: My only thought now is about testing. Are we testing both possibilities (with lookbehind and without)?

ziadkh0: No, only with lookbehind as Node supports it since version 8.10.
I don't know which approach is best suited to test without lookbehind support. 🤔
I think ideally, tests should be run in an environment that doesn't support lookbehind (like Safari < 16.4) but I'm not sure its possible...

styfle: Thanks! Mind adding a comment with the browser support so we don't forget in the future?

Maybe a comment with a link to this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion

src/rules.ts
```diff
@@ -16,6 +16,15 @@ function edit(regex: string | RegExp, opt = '') {
   return obj;
 }
 
+const supportsLookbehind = (() => {
+try {
+  // eslint-disable-next-line prefer-regex-literals
+  return !!new RegExp('(?<=1)(?<!1)');
+} catch {
```
styfle, line 23:
```suggestion
} catch {
  // See browser support here:
  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion
```

PR #3814, opened 2025-10-21
Modernize Docs UI with Tailwind, Dark Mode, and Improved Layout

samarJ19:
Fixes #3796 

Marked version:

Markdown flavor: CommonMark | GitHub Flavored Markdown | Markdown.pl | n/a

Description
This PR modernizes the documentation page for the MarkedJS website:

Refactors the docs UI using Tailwind CSS for a clean, modern look.
Improves layout spacing, section alignment, and overall readability.
Implements dark mode support with proper styling for code blocks and prose.
Simplifies the responsive design, removing unnecessary complexity while maintaining accessibility.
Ensures the GitHub corner banner is always visible and correctly positioned.
Refactors JavaScript for navigation highlighting, theme toggling, and copy-to-clipboard functionality.
Fixes all ESLint errors and passes linting and style checks.
Preserves the original docs structure while enhancing the user experience.
Adds screenshots of the new UI for visual reference.
Result
The documentation page now features a visually appealing, modern design with improved usability and accessibility. All code changes pass existing tests and linting requirements.

Contributor
 No new tests required for this PR (UI/UX changes only).
 All code changes are documented and follow project conventions.
Committer
 CI is green.
 Ready for squash and merge following conventional commit guidelines.

Screenshots: 

Screenshot 2025-10-21 210753
Screenshot 2025-10-21 210802

PR #3840, opened 2025-11-16
chore(docs): prevent theme flash by respecting system preference on initial load

samarJ19:
**Marked version:**
(same version used in PR #3814)

**Markdown flavor:**
GitHub Flavored Markdown

## Description

* Fixes regression introduced after **Modernize Docs UI with Tailwind, Dark Mode, and Improved Layout (#3814)**.
* The website was not consistently respecting **system theme preference** on the first load.
* The issue occurred because:

  * The inline `<script>` in the `<head>` correctly applied dark/light mode before render.
  * `index.js` later ran on `DOMContentLoaded` and **reapplied the theme unconditionally** via `applyTheme(initialTheme)`.
  * `applyTheme()` always removed the `dark` class first → causing:

    * A momentary flash
    * Theme being overridden incorrectly
    * Race condition between inline script and DOMContentLoaded logic

### Solution

This PR updates the theme initialization logic so that **applyTheme() only runs when necessary**.

Specifically:

* It now **checks whether the current DOM theme already matches the resolved theme** (system preference or saved preference).
* `applyTheme()` is called **only if a mismatch exists**.
* Prevents unnecessary DOM manipulation and eliminates:

  * Flash of incorrect theme
  * Double theme updates
  * Race conditions between inline script and JS loader

### Result

With this fix:

* System preference is applied correctly on first visit.
* User's saved preference (localStorage) still correctly overrides system settings.
* Theme toggle button continues to work as expected.
* No visual flashing between light/dark.

---

## Contributor

* [x] No new tests required (the fix only adjusts initialization logic; existing theme behavior tests cover expected output).
* [x] Not a new feature — continuation/fix to prior merged work (#3814).

Screenshot 2025-11-16 154856

### DIFF
--- a/docs/demo/demo.js
+++ b/docs/demo/demo.js
@@ -93,7 +93,7 @@ function getSystemTheme() {
 }
 
 function sanitisePreference(value) {
-  return value === 'dark' || value === 'light' || value === 'system' ? value : null;
+  return THEME_ORDER.includes(value) ? value : null;
 }
 
 function readStoredPreference() {


styfle:
@samarJ19 Please rebase from the main branch. I'm seeing 17 commits and 16 of those are merged already. I would expect to see only one commit with your change.

styfle (changes requested):
Please fix the merge conflicts

```
<<<<<<< HEAD
```

UziTech:
The demo page is always in light mode now

docs/demo/demo.js
```diff
@@ -52,70 +52,133 @@ $clearElem.addEventListener('click', handleClearClick, false);
...
+    writeStoredPreference(preference);
+  }
+}
+
+applyPreference(currentPreference, true);
+
+if ($themeToggle) {
+  $themeToggle.addEventListener('click', function() {
+    const index = THEME_ORDER.indexOf(currentPreference);
+    const nextIndex = index === -1 ? 0 : (index + 1) % THEME_ORDER.length;
```
styfle:
```suggestion
    const nextIndex = (index + 1) % THEME_ORDER.length;
```

(-1 + 1) is the same as `0`

docs/js/index.js
```diff
@@ -1,84 +1,122 @@
...
-  const initialTheme = getPreferredTheme();
-  applyTheme(initialTheme);
-
-  // Theme toggle click handler
-  if (themeToggle) {
-    themeToggle.addEventListener('click', function() {
-      const currentTheme = document.documentElement.classList.contains('dark') ? 'dark' : 'light';
-      const newTheme = currentTheme === 'dark' ? 'light' : 'dark';
+  function sanitisePreference(value) {
+    return value === 'dark' || value === 'light' || value === 'system' ? value : null;
```
styfle:
```suggestion
    return THEME_ORDER.includes(value) ? value : null;
```

Theres already a valid array so lets reuse that

docs/demo/demo.js
```diff
@@ -52,70 +52,133 @@ $clearElem.addEventListener('click', handleClearClick, false);
...
-applyTheme(initialTheme);
-
-if ($themeToggle) {
-  $themeToggle.addEventListener('click', function() {
-    const currentTheme = document.body.classList.contains('dark')
-      ? 'dark'
-      : 'light';
-    const newTheme = currentTheme === 'dark' ? 'light' : 'dark';
+function sanitisePreference(value) {
+  return value === 'dark' || value === 'light' || value === 'system' ? value : null;
```
styfle:
```suggestion
  return THEME_ORDER.includes(value) ? value : null;
```

docs/demo/demo.js
```diff
@@ -52,70 +52,133 @@ $clearElem.addEventListener('click', handleClearClick, false);
...
 
-    applyTheme(newTheme);
-    localStorage.setItem('theme', newTheme);
+function readStoredPreference() {
+  try {
+    const stored = sanitisePreference(localStorage.getItem(THEME_STORAGE_KEY));
+    if (stored) {
+      return stored;
+    }
+    return sanitisePreference(localStorage.getItem(LEGACY_STORAGE_KEY));
```
styfle, line 105: Why do we need legacy?

Isn't the old key compatible since the new key is adding a new value, not changing the others.

samarJ19: We keep the legacy key because:

The value set changed semantics: old code that reads theme only knows light/dark. If we started writing system into theme, that older code could misbehave or show the wrong thing, because it doesn’t know what system means.
There are multiple consumers: other pages/scripts (see [index.js](vscode-file://vscode-app/c:/Users/Asus/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and the built files under public/) still look at theme. Using a new key (theme-preference) lets new code understand system while old code keeps seeing only light/dark.
The migration pattern is intentional:
readStoredPreference prefers the new key but falls back to theme, so existing users keep their old choice.
writeStoredPreference mirrors light/dark into the legacy key, but removes it for system, so old pages treat “system” as “no explicit theme set” instead of a value they don’t understand.
So the legacy key isn’t fully compatible once you introduce system; the split new/legacy keys keeps both old and new behavior working safely.

PR #3925, opened 2026-03-23
fix: prevent heading and def tokens from greedily capturing multiple newlines

Puxhkar:
…nitions

**Marked version:**

## Description

## Contributor

## Committer


UziTech:
I will have to test this against a couple extensions. This could be a breaking change.

Puxhkar:
> I will have to test this against a couple extensions. This could be a breaking change.

Good point, thanks for calling that out.

I’ll test this with a few extensions and edge cases (especially ones that rely on `raw` and newline handling) to make sure nothing breaks. If I find any issues, I can either adjust the implementation or document it properly as a breaking change.

Let me know if there are any specific extensions you’d like me to check as well.

UziTech:
I looked at a few extensions that use the tokens and I don't see any that would break due to this change, but I still think we should consider this a breaking change since it is not really a fix.

If we want to be consistent we should also make html blocks and indented code blocks do the same thing.

UziTech:
I created #3939 to include html blocks and indented code blocks. We will try to get this out for the next major change.

UziTech (approved):
Thanks for fixing this 💯

Puxhkar:
> Thanks for fixing this 💯

can you please merge the PR

styfle (approved):

chocolateboy:
> I looked at a few extensions that use the tokens and I don't see any that would break due to this change, but I still think we should consider this a breaking change since it is not really a fix.

FYI: [this](https://github.com/markedjs/marked/releases/tag/v17.0.5) *is* a breaking change for [toc-md](https://www.npmjs.com/package/toc-md) and its (slightly less unmaintained) derivatives:

- [toc-md-alt](https://www.npmjs.com/package/toc-md-alt) 
- [@chocolateboy/toc-md](https://www.npmjs.com/package/@chocolateboy/toc-md)

<details>

## Input

```markdown
<!-- TOC -->

  h1
===

h2
--

### h3

```

## Expected

```markdown
<!-- TOC -->

- [  h1](#--h1)
  - [h2](#h2)
    - [h3](#h3)

<!-- TOC END -->

  h1
===

h2
--

### h3

```

## Actual

```markdown
<!-- TOC -->

- [h1](#h1)
  - [h2](#h2)
    - [h3](#h3)

<!-- TOC END -->

  h1
===

h2
--

### h3

```
</details>

PR #3945, opened 2026-04-09
fix(rules): ensure lookbehind regex is evaluated correctly by minifiers

### LINKED ISSUE #3944: SyntaxError: Invalid regular expression: invalid group specifier name in Safari < 16.4
**Marked version:**
18.0.0
**Describe the bug**

Very similar to #3816 - still experiencing an `invalid group specifier name`  in Safari 15. 

The regex that is produced for blockSkip (I think as I can only test remotely on compiled code) ends up to be: 

```
r = "\\[(?:[^\\[\\]`]|(?<a>`+)[^`]+\\k<a>(?!`))*?\\]\\((?:\\\\[\\s\\S]|[^\\\\\\(\\)]|\\((?:\\\\[\\s\\S]|[^\\\\\\(\\)])*\\))*\\)|(?<!`)()(?<b>`+)[^`]+\\k<b>(?!`)|<(?! )[^<>]*?>"
```

running the regex in safari 15 `new RegExp(r)`, produces the above error

**To Reproduce**
as above

**Expected behavior**
marked works with older version of Safari

Thank you for this lib and the efforts maintaining it !

christophe-g:
Use a default variable in the regex to prevent removal by minifiers. 

This is a very minor change that fix  #3944 downstream

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -19,7 +19,6 @@ function edit(regex: string | RegExp, opt = '') {
 // use a default variable to ensure the regex is evaluated on client an not removed by minifiers
 const supportsLookbehind = ((a = '') => {
 try {
-  // eslint-disable-next-line prefer-regex-literals
   return !!new RegExp('(?<=1)(?<!1)' + a);
 } catch {
   // See browser support here:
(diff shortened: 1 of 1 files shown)


UziTech (approved):
Thanks 💯

wyhaya:
Hi, can this PR be merged? I'm using Vite 8 and it seems there's no way to disable RegExp compression, so this PR is very helpful.

styfle (approved):

PR #3947, opened 2026-04-18
fix: fix infinite loop for indented code blank line

UziTech:
**Marked version:** v18.0.1

## Description

Fix infinite loop in indented code with only whitespace.

In v18 we introduced a change to indented code that removes trailing blank lines, but the way we detect blank lines as any whitespace we can create empty token if the code is only whitespace (e.g. '\v' vertical tab). Currently we will just keep creating empty indented code tokens until we run out of memory. 

Thanks for finding this @MaanVader

This PR checks for infinite loops and throws an error if found. It also defines a blank line in indented code tokens as only containing space or tab like it should. If the indented code block only has tabs or spaces it is not an indented code block according to CommonMark.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -110,7 +110,15 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
       src = src.replace(other.tabCharGlobal, '    ').replace(other.spaceLine, '');
     }
 
+    let srcLength = Infinity;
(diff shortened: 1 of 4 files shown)


styfle (approved):

emwhipple30-svg (commented):
d

PR #3969, opened 2026-05-08
[codex] Cache list indentation regexes

astelmach20:
**Marked version:**

18.0.3 / `69257e45`

## Description

This caches the indentation-sensitive regexes used while scanning list items. These regexes are rebuilt for each list item today even though the effective indentation is bounded to `Math.min(3, indent - 1)`, so the same few regex shapes are reused repeatedly during list parsing.

A CPU profile of a CommonMark parse loop showed samples in `nextBulletRegex`, `fencesBeginRegex`, `headingBeginRegex`, and `htmlBeginRegex`. After caching the bounded indentation regexes, those helper functions no longer showed self samples in the same profile pass.

Local benchmark notes from my machine:

- `npm run bench`: `marked` improved from 1655 ms to 1513 ms on the CommonMark benchmark run.
- A list-heavy parse benchmark improved from roughly 488-502 ms to 376-466 ms across seven rounds.
- A repeated CommonMark parse loop improved from roughly 244-260 ms to 229-245 ms across seven rounds.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression: existing spec/unit coverage for list parsing, plus full `npm test` passed locally.
- [x] If submitting new feature, it has been documented in the appropriate places. n/a, this is not a new feature.

## Committer

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -3,8 +3,8 @@ const noopTest = { exec: () => null } as unknown as RegExp;
 function cachedIndentRegex(createRegex: (indent: number) => RegExp) {
   const cache: RegExp[] = [];
   return (indent: number) => {
-    const cappedIndent = Math.min(3, indent - 1);
-    const cacheIndex = cappedIndent + 1;
+    const cappedIndent = Math.max(0, Math.min(3, indent - 1));
+    const cacheIndex = cappedIndent;
     let regex = cache[cacheIndex];
     if (!regex) {
       regex = createRegex(cappedIndent);
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -1,5 +1,19 @@
 const noopTest = { exec: () => null } as unknown as RegExp;
 
+function cachedIndentRegex(createRegex: (indent: number) => RegExp) {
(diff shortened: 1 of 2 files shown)


UziTech (approved):
Thanks for helping make marked faster! 💯

src/rules.ts
```diff
@@ -1,5 +1,19 @@
...
+  const cache: RegExp[] = [];
+  return (indent: number) => {
+    const cappedIndent = Math.max(0, Math.min(3, indent - 1));
+    const cacheIndex = cappedIndent;
+    let regex = cache[cacheIndex];
+    if (!regex) {
+      regex = createRegex(cappedIndent);
+      cache[cacheIndex] = regex;
+    }
+    return regex;
```
styfle:
```suggestion
    const cacheIndex = Math.max(0, Math.min(3, indent - 1));
    let regex = cache[cacheIndex];
    if (!regex) {
      regex = createRegex(cacheIndex);
      cache[cacheIndex] = regex;
    }
    return regex;
```

Seems like you could eliminate one of these variables since the indent and index are the same value.

PR #3984, opened 2026-06-02
fix: parse empty list item with trailing space

sarathfrancis90:
**Marked version:** 18.0.4 (master)

## Description

A bare list marker followed only by a space (e.g. `- `, `* `, `1. `, `1) `) was not recognized as a list. It fell through to the paragraph rule and rendered as a paragraph instead of an empty list item.

This is internally inconsistent: a marker with **no** trailing space already worked, and an empty item that continues an existing list also worked — only a marker followed by a single trailing space was broken.

## Expectation

`- ` (and `* `, `1. `, `1) `) should produce an empty list item, matching the CommonMark reference implementation (commonmark 0.31.2) and markdown-it 14.2.0 in CommonMark mode:

```html
<ul>
<li></li>
</ul>
```

## Result

marked produced a paragraph containing the literal marker:

```html
<p>- </p>
```

## Root cause

The block `list` rule required at least one content character after the marker's whitespace:

```
^(bull)([ \t][^\n]+?)?(?:\n|$)
```

The `+?` means a marker followed only by whitespace cannot match, so it falls through to the paragraph rule. A previous change (PR #3890) updated the marker prefix but left this content group as `+?`, so the trailing-space case stayed broken (the fixture there had no trailing space). This PR is the complementary fix.

## Fix

Relax the content group from `[ \t][^\n]+?` to `[ \t][^\n]*?` so a marker followed only by whitespace still starts a list:

```
^(bull)([ \t][^\n]*?)?(?:\n|$)
```

To preserve CommonMark's rule that an *empty* list item cannot interrupt a paragraph, the paragraph (and GFM block) list-interrupt patterns now require a non-blank character after the marker, so `foo\n+ ` stays a single paragraph while `foo\n- bar` still interrupts. The GFM table interrupt pattern is intentionally left unchanged, since a bare bullet still ends table rows.

## Test evidence

- Added `test/specs/new/list_item_empty_trailing_space.{md,html}` covering `- `, `* `, `1. `, `1) `, the paragraph non-interrupt case `foo\n+ `, and the still-interrupting case `bar\n1. `.
- The new fixture fails on unmodified master and passes with this change (TDD).
- Full suite green locally: `test:specs` 1743 pass / 0 fail, `test:unit` 188 pass / 0 fail, plus `test:cjs`, `test:umd`, `test:types` (attw) and `test:lint`. CommonMark "List items" (48/48) and "Lists" (26/26) remain at 100% with no regression.

## Contributor

## Committer


UziTech (approved):
Thanks! 💯

sarathfrancis90:
Thanks for the quick review, @UziTech! 🙏 Glad it lines up with the CommonMark reference and markdown-it behavior — happy to adjust anything if needed.

styfle (approved):
Confirmed this matches Commonmark, thanks!

PR #3989, opened 2026-06-07
docs: add @webc.site/math-marked to known extensions

x-at-01:
Hi!

As discussed in #3988, this PR adds `@webc.site/math-marked` to the list of known extensions in `docs/USING_ADVANCED.md`.

`@webc.site/math-marked` is a 3.5KB zero-dependency TeX-to-MathML compiler extension for Marked that compiles inline and block LaTeX/TeX math formulas directly to MathML at compile time.

Thank you!


UziTech (approved):
Thanks 💯

PR #3991, opened 2026-06-12
fix: keep trailing text on HTML block close line for PI, declarations, and CDATA

sarathfrancis90:
## Description

While checking marked against the CommonMark spec I noticed that text following the closing delimiter of an HTML block on the *same line* is dropped from the block for type 3 (`<?...?>`), type 4 (`<!...>`), and type 5 (`<![CDATA[...]]>`).

CommonMark ends these blocks on the line containing the closing delimiter, and the entire line — including anything after the delimiter — is part of the raw HTML block. marked already handles this for type 2 comments (spec [example 177](https://spec.commonmark.org/0.31.2/#example-177) shows `<!-- foo -->*bar*` keeping `*bar*` as raw text), but rules 3–5 stopped right at the delimiter, so the trailing text got split off and reparsed.

For example:

```
<?php echo '>'; ?>*not emphasis*
```

rendered as:

```html
<?php echo '>'; ?><p><em>not emphasis</em></p>
```

instead of leaving `*not emphasis*` inside the HTML block like the reference does.

## Fix

The comment rule (2) already consumes the rest of the closing line with a trailing `[^\n]*`. I applied the same to rules 3, 4, and 5 so they consume to the end of the closing line before the newline.

## Tests

Added `test/specs/new/html_block_trailing_text.md` covering all three block types. It fails on `master` and passes with this change. Full CommonMark/GFM spec suites, unit tests, lint, types, and the ReDoS check are all green, and the CommonMark conformance score is unchanged.

## Contributor checklist

- [x] Tests pass (`npm test`)
- [x] Lint passes (`npm run lint`)
- [x] Added a test for the fixed behavior


UziTech (approved):
Nice catch! Thanks for fixing this 💯

styfle (approved):

PR #3997, opened 2026-06-17
fix: apply Unicode full case fold in link reference label normalization

JSap0914:
## Bug

JavaScript's `String.prototype.toLowerCase()` implements Unicode simple case mapping, not full case folding. The CommonMark spec §5.6 requires Unicode case folding when matching link reference labels:

> One label matches another just if their normalized forms are equal. To normalize a label, perform the Unicode case fold on the label, and collapse consecutive spaces, tabs, and line endings to a single space.

The key difference: full case folding allows a single character to expand to multiple characters. For example:

| Character | `toLowerCase()` | Full case fold |
|-----------|-----------------|----------------|
| ẞ (U+1E9E) | ß | ss |
| ß (U+00DF) | ß | ss |
| ﬁ (U+FB01) | ﬁ | fi |

This means **CommonMark spec example 540** fails:

```markdown
[ẞ]

[SS]: /url
```

Expected: `<p><a href="/url">ẞ</a></p>`
Got: `<p>[ẞ]</p>` (unresolved reference)

## Fix

Add `normalizeLabel()` to `src/helpers.ts` that chains `toLowerCase()` with replacements for the Unicode F-status (full) case fold pairs that expand to multiple characters. Update `Tokenizer.def` and `Tokenizer.reflink` to use it instead of plain `toLowerCase()`.

## Verification

```
npm test
```

- CommonMark Links: 77/90 → 78/90 (example 540 now passes)
- GFM Links: same improvement
- All 1743 spec tests pass; all 188 unit tests pass; lint clean

> **AI assistance disclosure:** This fix was developed with AI-assisted tooling.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -164,3 +164,25 @@ export function expandTabs(line: string, indent = 0) {
 
   return expanded;
 }
+
+/**
+ * Normalize a link reference label per CommonMark spec:
+ * perform Unicode full case fold (covering multi-char expansions that
+ * String.prototype.toLowerCase() misses, e.g. ẞ→ss, ligatures→letters),
+ * then collapse runs of whitespace to a single space.
+ *
+ * @see https://spec.commonmark.org/0.31.2/#matches
+ */
+export function normalizeLabel(label: string): string {
+  return label
+    // Apply Unicode full case fold before toLowerCase so that
+    // ẞ (U+1E9E) → ss and ß (U+00DF) → ss (simple fold maps ẞ to ß first).
+    .toLowerCase()
+    .replace(/ß/g, 'ss') // U+00DF + folded U+1E9E
+    .replace(/ﬀ/g, 'ff') // U+FB00
+    .replace(/ﬁ/g, 'fi') // U+FB01
+    .replace(/ﬂ/g, 'fl') // U+FB02
+    .replace(/ﬃ/g, 'ffi') // U+FB03
+    .replace(/ﬄ/g, 'ffl') // U+FB04
+    .replace(/ﬅ|ﬆ/g, 'st'); // U+FB05, U+FB06
+}
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -5,6 +5,7 @@ import {
   findClosingBracket,
   expandTabs,
   trimTrailingBlankLines,
+  normalizeLabel,
 } from './helpers.ts';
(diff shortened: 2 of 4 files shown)


UziTech (approved):
Thanks! 💯

src/helpers.ts
```diff
@@ -164,3 +164,25 @@ export function expandTabs(line: string, indent = 0) {
...
+    // Apply Unicode full case fold before toLowerCase so that
+    // ẞ (U+1E9E) → ss and ß (U+00DF) → ss (simple fold maps ẞ to ß first).
+    .toLowerCase()
+    .replace(/ß/g, 'ss') // U+00DF + folded U+1E9E
+    .replace(/ﬀ/g, 'ff') // U+FB00
+    .replace(/ﬁ/g, 'fi') // U+FB01
+    .replace(/ﬂ/g, 'fl') // U+FB02
+    .replace(/ﬃ/g, 'ffi') // U+FB03
+    .replace(/ﬄ/g, 'ffl') // U+FB04
+    .replace(/ﬅ|ﬆ/g, 'st'); // U+FB05, U+FB06
```
styfle, line 187: Is this list exhaustive?

I'm wondering if we can use a built-in API like [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) or perhaps [`label.normalize('NFKC')`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) instead?

src/helpers.ts
```diff
@@ -164,3 +164,25 @@ export function expandTabs(line: string, indent = 0) {
...
+ * then collapse runs of whitespace to a single space.
+ *
+ * @see https://spec.commonmark.org/0.31.2/#matches
+ */
+export function normalizeLabel(label: string): string {
+  return label
+    // Apply Unicode full case fold before toLowerCase so that
+    // ẞ (U+1E9E) → ss and ß (U+00DF) → ss (simple fold maps ẞ to ß first).
+    .toLowerCase()
+    .replace(/ß/g, 'ss') // U+00DF + folded U+1E9E
```
styfle, line 181: I see the test for this first one, but do we have tests for the others?

UziTech (changes requested):
@JSap0914 can you fix the comments from @styfle ?

UziTech:
This is fixed in #4077

PR #4003, opened 2026-06-23
fix: Fix ordered lists after blockquotes

cyphercodes:
**Marked version:**

18.0.5

**Markdown flavor:** Markdown.pl|CommonMark|GitHub Flavored Markdown|n/a

CommonMark / GitHub Flavored Markdown

## Description

- Fixes #4002.
- Keeps ordered lists that follow a blockquote as sibling blocks instead of lazy blockquote continuations.
- Adds a regression spec for an ordered list after a blockquote.

## Tests

- `npm test`

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR); or,
- [ ] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

## Committer

In most cases, this should be a different person than the contributor.

- [ ] CI is green (no forced merge required).
- [ ] Squash and Merge PR following [conventional commit guidelines](https://www.conventionalcommits.org/).

Issue #4002, opened 2026-06-23
Ordered list following a blockquote is incorrectly nested inside the blockquote


Deuscx:
**Marked version:** 18.0.5

**Describe the bug**

When an **ordered list item** immediately follows a blockquote line (no blank line, list marker at column 0), marked’s `blockquote` tokenizer **lazy-continues** and absorbs the list into the blockquote. The list is parsed as a nested ordered list inside the blockquote instead of a sibling block.

This affects **ordered lists only**. An unordered list after a blockquote is parsed correctly as a sibling. A blockquote followed by a blank line and then a list is also parsed correctly.

The root cause appears to be in the `blockquote()` tokenizer: when the last parsed child inside a blockquote is a `list`, marked merges remaining lines into that list (`h?.type === "list"` branch), which pulls in subsequent ordered-list lines that should end the blockquote.

**To Reproduce**

1. [Marked Demo](https://marked.js.org/demo/) — paste:

```markdown
1. test
> blockquote
2. test
> blockquote
```

2. [CommonMark Dingus](https://spec.commonmark.org/dingus/) — same input.

Or run locally:

```bash
npm install marked
node -e "
const { marked } = require('marked');
const md = \`1. test
> blockquote
2. test
> blockquote\`;
console.log(marked.parse(md, { gfm: true, breaks: true }));
"
```

**Actual output (marked):**

```html
<ol>
<li>test</li>
</ol>
<blockquote>
<p>blockquote</p>
<ol start="2">
<li>test<br>blockquote</li>
</ol>
</blockquote>
```

Lexer tokens show the second list item and the second `> blockquote` line are trapped inside one blockquote token:

```js
// Simplified token tree
[
  { type: 'list', items: [{ text: 'test' }] },
  {
    type: 'blockquote',
    tokens: [
      { type: 'paragraph', text: 'blockquote' },
      { type: 'list', ordered: true, start: 2, items: [{ text: 'test\nblockquote' }] }
    ]
  }
]
```

**Expected behavior**

Per CommonMark (and micromark), the blockquote should close after its content, and the following ordered list items and blockquotes should be **sibling blocks**:

```html
<ol>
<li>test</li>
</ol>
<blockquote>
<p>blockquote</p>
</blockquote>
<ol start="2">
<li>test</li>
</ol>
<blockquote>
<p>blockquote</p>
</blockquote>
```

Minimal case:

```markdown
> blockquote
2. test
```

[...]

PR #4004, opened 2026-06-25
fix: keep empty list after blockquote as a sibling block

spokodev:
## Description

A blockquote immediately followed by a bare list marker line nests an empty list inside the blockquote instead of starting a sibling list.

**Minimal repro**

```js
marked.parse('> foo\n-');
```

Expected (matches commonmark.js and markdown-it):

```html
<blockquote>
<p>foo</p>
</blockquote>
<ul>
<li></li>
</ul>
```

Actual:

```html
<blockquote>
<p>foo</p>
<ul>
<li></li>
</ul>
</blockquote>
```

Reproduces for `-`, `*`, `+`, `1.`, `1)` and for multi line blockquotes (`> foo\n> bar\n-`). Non empty lists (`> foo\n- bar`) already behave correctly.

## Root cause

The `blockquote` rule reuses the paragraph list interrupt clause:

```
 {0,3}(?:[*+-]|1[.)])[ \t]+[^ \t\n]
```

The trailing `[ \t]+[^ \t\n]` requires content after the marker, which is correct for a top level paragraph (an empty list cannot interrupt a paragraph). But the same clause also controls when a blockquote stops its lazy continuation. A bare marker line does not match it, so the marker line is lazily continued into the blockquote paragraph and then re lexed into a nested empty list.

Per CommonMark 0.31.2 (5.1 block quotes laziness, 5.2 list items) a list begun by a bare marker is a sibling block after the blockquote, not a lazy continuation of it. commonmark.js and markdown-it both produce the sibling.

## Fix

Give the blockquote its own paragraph variant (`blockquoteParagraph`) whose list interrupt clause also matches a bare marker:

```
 {0,3}(?:[*+-]|1[.)])(?:[ \t]|\n|$)
```

The top level `paragraph` rule is left untouched, so `foo\n-` (setext heading), `foo\n*` and `foo\n1.` (no interruption) keep their current behavior. The change only affects when a blockquote ends.

## Relation to #4003

#4003 also routes the `blockquote` rule through a `blockquoteParagraph` const, but for a different bug (#4002, non empty ordered lists with a start other than 1). It keeps the `[ \t]+[^ \t\n]` content requirement, so it does not fix the empty marker case verified here. The two changes touch the same region; I can rebase or fold the two interrupt clauses into one if a maintainer prefers a single edit.

## Tests

- Added `test/specs/new/empty_list_after_blockquote.{md,html}` covering the bare unordered and bare ordered cases including a multi line blockquote.
- `npm test` green, including the CommonMark and GFM conformance specs (1745 spec assertions, 188 unit tests).

## Contributor

## Committer


UziTech (changes requested):
looks like this has a merge conflict because of #4003

Other than that this looks good

spokodev:
Rebased onto master and resolved the conflict with #4003. The two fixes are compatible: I kept #4003's `createParagraph` helper and its any-number blockquote interruption, and folded in the bare-marker case so an empty list marker after a blockquote still starts a sibling list. Both cases pass together (`> foo` + `2. bar` gives `<ol start="2">`, `> foo` + `-` gives an empty `<ul>`). Full spec suite green (1749) plus unit (188).

UziTech (approved):
Thanks for fixing this! 💯

styfle (approved):

PR #4007, opened 2026-07-01
fix: treat a line of only tabs as a blank line between paragraphs

sarathfrancis90:

A line that's only tabs (or tabs mixed with spaces) is a blank line in CommonMark, so it should separate paragraphs. Marked only treats a *spaces*-only line as blank, so when the blank line between two paragraphs happens to contain a tab, that line gets swallowed as a lazy continuation and the two paragraphs merge into one.

With a single tab on the middle line:

```
foo
	
bar
```

marked currently returns one paragraph (`<p>foo\tbar</p>`) instead of `<p>foo</p>\n<p>bar</p>`. A space-only middle line already works, which is what made this easy to miss.

The paragraph rule's continuation lookahead was checking for `` +\n`` (spaces only); I widened it to `[ \t]+\n` so a tab-only line counts as blank too. GFM and pedantic paragraphs build on the same base rule, so they pick up the fix as well. Added a spec test under `test/specs/new`.

## Contributor

- [x] Test added (`blank_line_with_tabs`)
- [x] No new feature to document

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -128,7 +128,7 @@ const lheadingGfm = edit(lheadingCore)
   .replace(/html/g, / {0,3}<[^\n>]+>\n/) // block html can interrupt
   .replace(/table/g, / {0,3}\|?(?:[:\- ]*\|)+[\:\- ]*\n/) // table can interrupt
   .getRegex();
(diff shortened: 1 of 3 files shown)


UziTech (approved):
Nice catch! 💯

styfle (approved):

PR #4008, opened 2026-07-02
fix: an unmatched strong run before emphasis stays literal

spokodev:
The input `**a*b*c` produced malformed output.

- marked: `<p>*<em>a</em>b*c</p>`
- CommonMark 0.31.2 reference and markdown-it: `<p>**a<em>b</em>c</p>`

When the `**` opener finds no closer, the lexer reprocesses from the second `*`
of the run. That mid-run `*` greedily pairs with the `*` before `b`, even though
that `*` is itself the opener of `*b*`. The unmatched `**` should stay literal
while the inner `*b*` forms its own emphasis.

The fix marks an opener that is reprocessed from inside its own delimiter run
(`prevChar` equals the delimiter character) and stops it from closing against an
ambiguous delimiter that can also open, so that delimiter opens its own span
instead. This is distinct from the valid cases `**foo*`, `***foo**` and
`****foo*`, which close against unambiguous right delimiters and are preserved.

All CommonMark and marked spec tests pass. Added a spec fixture for the case.


UziTech (approved):
Thanks! 💯

UziTech:
@styfle any objections with merging this?

PR #4014, opened 2026-07-12
fix: Avoid O(n^2) backtracking in HTML block close and tilde interrupt regexes

hong4rc:
Follow-up to #4013 — two more spots in `src/rules.ts` where a block-level regex backtracks quadratically on a long single line. Split into one commit each.

The first is the HTML block close. #3991 rewrote the close of the closing-tag, processing-instruction, declaration, and CDATA branches to `[^\n]*\n+` so trailing text on the close line is kept. Keeping the trailing text is right, but the `\n+` isn't: it requires at least one newline, so when the input ends without a trailing newline the close can never match and the engine retries every split point of the preceding lazy `[\s\S]*?` before finally falling through to `$`. That's O(n²):

```js
marked.parse('<!x ' + 'a>'.repeat(80000));            // ~5s
marked.parse('<script>' + 'a</script>'.repeat(40000)); // ~6.5s
```

Switching `\n+` back to `\n*` closes on the first match. Whenever a trailing newline is present (the normal case) `\n*` and `\n+` consume exactly the same text, so output is unchanged — #3991's own trailing-text fixtures still pass.

The second is the paragraph/table/blockquote interrupt check. Its fence sub-pattern is `` (?:`{3,}(?=[^`\n]*\n)|~{3,})[^\n]*\n ``. The backtick branch is guarded by a lookahead, but `~{3,}` isn't, and it overlaps with the following `[^\n]*`, so a long run of tildes with no newline backtracks quadratically:

```js
marked.parse('intro\n' + '~'.repeat(80000));           // ~5s
```

Because `~{3,}` is always immediately followed by `[^\n]*`, `~~~[^\n]*` matches exactly the same strings, so replacing `~{3,}` with `~~~` removes the overlap without changing what matches. The real fenced-code tokenizer (which captures the fence length via a group and a `$` alternative) is left untouched — only the three interrupt copies change.

Both changes are behavior-preserving: the full spec and unit suites pass with identical output, and I added a `quadratic_*.cjs` guard next to #4013's for each.

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -169,7 +169,7 @@ const createParagraph = (listInterrupt: RegExp) => edit(_paragraph)
   .replace('|lheading', '') // setext headings don't interrupt commonmark paragraphs
   .replace('|table', '')
   .replace('blockquote', ' {0,3}>')
-  .replace('fences', ' {0,3}(?:`{3,}(?=[^`\\n]*\\n)|~{3,})[^\\n]*\\n')
+  .replace('fences', ' {0,3}(?:`{3,}(?=[^`\\n]*\\n)|~~~)[^\\n]*\\n')
   .replace('list', listInterrupt)
   .replace('html', '</?(?:tag)(?: +|\\n|/?>)|<(?:script|pre|style|textarea|!--)')
   .replace('tag', _tag) // pars can be interrupted by type (6) html blocks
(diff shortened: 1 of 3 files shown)


UziTech (approved):
Thanks for fixing theses! 💯

styfle (approved):

PR #4015, opened 2026-07-12
fix: Recognize setext headings whose first line starts with #

hong4rc:
A line like `#1 Goals` or `#hashtag` followed by a setext underline is rendered as a paragraph plus an `<hr>` instead of a heading:

```js
marked.parse('#1 Goals\n---------');  // <p>#1 Goals</p>\n<hr>
```

Both commonmark.js and markdown-it produce `<h2>#1 Goals</h2>` here, and that matches the CommonMark spec: a line only starts an ATX heading when the `#`s are followed by a space, a tab, or the end of the line, so `#1` is ordinary paragraph text — which means the `---------` beneath it makes a setext heading.

The `lheading` rule already knows other blocks can interrupt a setext heading, but its ATX check is ` {0,3}#{1,6}`, which matches any line that merely *starts* with `#` and misses the "followed by whitespace" requirement. marked's own paragraph rules already get this right (` {0,3}#{1,6}(?:\s|$)`), as does the ATX `heading` rule itself (`(?=\s|$)`); this just brings `lheading`/`lheadingGfm` in line with them.

Valid ATX headings are unaffected — `# Real\n===` still renders `<h1>Real</h1>` followed by `<p>===</p>` — and the full spec suite passes with the added fixture.


UziTech (approved):
Nice catch! 💯

styfle (approved):

PR #4023, opened 2026-07-18
fix: fall back to default checkbox renderer when extension returns false

sarathfrancis90:

A renderer extension named after a built-in token can return `false` to fall back to the default renderer — that's the documented contract in USING_PRO, and it works for `code`, `heading`, `text` and the rest. It doesn't work for `checkbox`: the checkbox silently disappears instead.

```js
const marked = new Marked({
  extensions: [{
    name: 'checkbox',
    renderer(token) {
      return token.checked ? false : '<span class="todo"></span> ';
    },
  }],
});

marked.parse('- [x] one\n- [ ] two\n');
```

I expected the checked item to fall back to the default `<input checked="" disabled="" type="checkbox">`, but it renders as just `<li>one</li>`. The unchecked item never returns `false`, and it renders fine, so only the fallback path is broken.

`Parser` decides whether a `false` return should fall through to the built-in renderer by testing the token type against a hardcoded list of built-in type names. `checkbox` was added to both parse switches in #3755 but never to either list, so the fallback branch is unreachable for it and the `ret || ''` path wins.

Added `checkbox` to the block and inline lists. Both matter in practice: a tight list puts the checkbox token at block level while a loose list puts it inline, so there's a test for each.

## Contributor

- [x] Tests added (checkbox extension fallback, tight and loose list)
- [x] No new feature to document

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -50,7 +50,7 @@ export class _Parser<ParserOutput = string, RendererOutput = string> {
       if (this.options.extensions?.renderers?.[anyToken.type]) {
         const genericToken = anyToken as Tokens.Generic;
         const ret = this.options.extensions.renderers[genericToken.type].call({ parser: this }, genericToken);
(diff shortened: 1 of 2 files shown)


UziTech (approved):
Nice catch! 💯

PR #4024, opened 2026-07-19
fix: Avoid O(n^2) backtracking in GFM email autolink regexes

hong4rc:
Same shape of bug as #4013, this time in the GFM email autolink path.

The `url` rule (used for autolinking bare emails) and the `text` rule (used to find where the next inline token starts) each retry at every character position in the input, and each one contains an unbounded `[A-Za-z0-9._+-]+` run that scans forward looking for an `@`. When the input has a long run of local-part-looking characters with no `@` anywhere in it, every one of those retries rescans the same tail of the string, so tokenizing goes quadratic in the length of that run.

The fix bounds both runs to `{1,64}` instead of `+`, matching RFC 5321's 64-character limit on the email local part. That caps how far each scan can walk, so tokenizing stays linear, and it doesn't lose any valid input — a local part longer than 64 characters was never a valid email to begin with.

Checked with the full spec suite (1753/1753 at the 18.0.6 base this branch started from, still all green after rebasing onto current master) plus a new `test/specs/redos/quadratic_email_autolink.cjs` case that's red before the fix and green after. Timing-wise, `parse('a_'.repeat(32000))` went from ~2420ms to ~33ms.

While in there I noticed the sibling `<email>`-bracket autolink rule around rules.ts:379 has the same unbounded-run shape — didn't touch it here since it's a separate code path, but it's a plausible follow-up.


UziTech:
While it is correct that email addresses can't have a local name longer than 64 characters, marked is not trying to parse valid email addresses. The goal of marked with GFM is to copy GitHub's markdown parsing. For GitHub there is no limit to this length.

example:

abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz@example.com

PR #4030, opened 2026-07-23
fix: Fix spurious deeper nesting in a blockquote continuation

spokodev:
### Problem

In a nested blockquote, a paragraph line that lazily continues and is then
followed by a line restating the `>` markers is parsed into an extra, spurious
blockquote level (and a split paragraph) instead of continuing the same
paragraph.

Input:

```
> > a
b
> > c
```

`marked` (before this PR):

```html
<blockquote>
<blockquote>
<p>a
b</p>
<blockquote>
<p>c</p>
</blockquote>
</blockquote>
</blockquote>
```

Expected — identical output from both `commonmark` 0.31.2 and
`micromark` + `micromark-extension-gfm`:

```html
<blockquote>
<blockquote>
<p>a
b
c</p>
</blockquote>
</blockquote>
```

`> > a\nb\n> > > c` shows the same off-by-one-level (4 blockquotes in marked vs
the correct 3). Plain lazy nesting without a restated marker
(`> > > foo\nbar`, CommonMark example 250) is unaffected, which is why no
existing spec caught it.

### Root cause

The nested-blockquote continuation branch re-lexes the inner blockquote from
`oldToken.raw + '\n' + lines.join('\n')`. `oldToken.raw` already had one `>`
marker stripped for the inner frame, but the appended continuation `lines` still
carry the outer markers, so a restated `> >` keeps one surplus `>` when recursed
and is read as a fresh deeper blockquote that interrupts the open paragraph. The
single-level path works only because the marker is stripped there via
`blockquoteSetextReplace2`; this branch bypassed that strip.

### Fix

Strip one `>` marker from the continuation lines (same `blockquoteSetextReplace2`
the normal path uses) so they align with the inner frame, and rebuild `raw` from
the actual consumed source. Deeper markers (`> > >`) simply recurse one further
level, which is correct.

### Test

Added `test/specs/new/blockquote_nested_lazy_continuation` (`.md`/`.html`).
Fails on `master` (2 failing spec cases), passes with the fix. Full suite green:
`test:specs` 1769 pass / 0 fail, `test:unit` 188 pass / 0 fail.

Reference: CommonMark 0.31.2 §6.3 (block quotes — laziness): a line beginning
with the blockquote's own markers followed by paragraph text continues that
paragraph, not a new deeper quote.


UziTech (approved):
Nice catch! 💯

PR #4033, opened 2026-07-30
docs: explain partial extensions in use

### LINKED ISSUE #4016: Comment source code functions for intellisense
In trying to supply a custom `processAllTokens` hook:
```js
marked.marked.lexer(stringToParse, { hooks: { processAllTokens: tokens => tokens } })
```

Typescript wanted `hooks` to include all the other *possible* functions the Hooks object can have.
> Type '{ processAllTokens: (tokens: TokensList | Token[]) => Token[]; }' is missing the following properties from type '_Hooks<string, string>': options, preprocess, postprocess, emStrongMask, and 2 more.

I eventually determined `marked.use({ hooks: {` works because it combines the hooks object, making it partial. However, this would've been easier if `marked.use` mentioned how this works in the intellisense, or if other options did. It's harder to connect the meaning of functions across the app. I think I could've saved a lot of time if the documentation included the source code functions.

(Very excellent software, by the way. I will be using this in my own project.)

ColumbusLabs:
**Marked version:** 58ed4af62e7383ca770ac68aa987ca39887f33f3

**Markdown flavor:** n/a

## Description

- Fixes #4016.
- Adds source-level documentation to `Marked#use`, `marked.use`, and the named `use` export.
- Explains that `renderer`, `tokenizer`, and `hooks` can be partial when registered through `use`, because they are merged with built-in and previously registered behavior.
- Documents the default, instance, and inline parse APIs with the recommended per-call pattern: create separate `Marked` instances and configure each through `use` instead of passing hooks, renderers, or tokenizers on every call.

The named `use` export now aliases the documented implementation function so generated declarations expose the same IntelliSense text for both `marked.use` and `import { use }`.

## Contributor

- [x] no tests required for this documentation-only PR.
- [x] If submitting new feature, it has been documented in the appropriate places. (Not a new feature.)

Verification:

- `npm test` (build, 1,767 spec tests, 190 unit tests, UMD, CJS, package/type checks, and lint)
- `npm run test:types`
- `npm run test:lint`
- Generated-declaration inspection for default, instance, inline, and named APIs
- `git diff --check`
- `npm run build:reset`

Note: `docs/CONTRIBUTING.md` still mentions `npm run test:all`, but that script is not present in `package.json`; `npm test` is the current full pipeline.

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -18,7 +18,20 @@ export class Marked<ParserOutput = string, RendererOutput = string> {
   defaults = _getDefaults<ParserOutput, RendererOutput>();
   options = this.setOptions;
 
+  /**
+   * Compiles markdown to HTML.
+   *
+   * To configure hooks, a renderer, or a tokenizer, create a separate `Marked`
+   * instance and use `Marked#use` instead of passing them on each call.
+   */
   parse = this.parseMarkdown(true);
+
+  /**
+   * Compiles inline markdown to HTML.
+   *
+   * To configure hooks, a renderer, or a tokenizer, create a separate `Marked`
+   * instance and use `Marked#use` instead of passing them on each call.
+   */
   parseInline = this.parseMarkdown(false);
 
   Parser = _Parser<ParserOutput, RendererOutput>;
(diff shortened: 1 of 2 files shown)


UziTech:
Shouldn't the per-call docs be on the parse function? I don't think people who use the per-call options are going to be looking at the `use` docs for instructions.

Also we want to discourage hooks, renderer, tokenizer in the per-call options. They are still there for legacy reasons but there shouldn't be any reason they need to be used. If you need different options per-call it is best to create different instances of `Marked` and use `use` to set them.

ColumbusLabs:
Addressed in 7208a550: the per-call caveat is no longer placed in the use docs. The parse and parseInline docs now recommend separate Marked instances configured through use, covering the default, instance, and named APIs. I verified the generated declarations, the full npm test pipeline, test:types, lint, build reset, and a clean diff check.

UziTech (approved):
Thanks!

PR #4039, opened 2026-08-02
fix: parse EOF backtick fences after paragraphs

### LINKED ISSUE #4038: Backtick code fence at EOF does not interrupt a paragraph
### Marked version

Current `master` (18.0.7 at time of testing)

### Describe the bug

A fenced code block at the end of the document is parsed as literal paragraph text when it immediately follows a paragraph. This affects both backtick and tilde fences. CommonMark allows fenced code blocks to interrupt paragraphs, and an unclosed fenced code block continues until the end of the containing block.

### To reproduce

```js

console.log(marked.parse('foo\n```', { gfm: false }));
console.log(marked.parse('foo\n~~~', { gfm: false }));
```

Actual:

```html
<p>foo
```</p>
```

and

```html
<p>foo
~~~</p>
```

Expected for both fence characters:

```html
<p>foo</p>
<pre><code></code></pre>
```

### Compatibility

This only affects EOF fence recognition when checking whether a paragraph/table row should be interrupted.

deepakganesh78:
Fixes #4038

## Reproduction

On current `master`, EOF fenced code blocks are swallowed into the preceding paragraph for both fence characters:

```js
import { marked } from './lib/marked.esm.js';

marked.parse('foo\n```', { gfm: false }); // before: "<p>foo\n```</p>\n"
marked.parse('foo\n~~~', { gfm: false }); // before: "<p>foo\n~~~</p>\n"
```

CommonMark treats backtick and tilde fences the same for paragraph interruption, and an unclosed fence is closed by EOF. With this change, `foo\n````, `foo\n```js`, `foo\n~~~`, and `foo\n~~~js` all end the paragraph and produce a fenced code block.

## Root cause

The block fence tokenizer already accepts EOF after an opening fence, but the paragraph/table interruption regexes only recognized a fence when the opening fence line ended with `\n`. At EOF, paragraph parsing consumed the fence as literal text before the fence tokenizer could run. The first version of this PR only allowed EOF in the backtick branch; the tilde branch had the same bug.

## Fix

Allow the paragraph/table interruption checks to match either a newline or EOF after the opening fence line for both backtick and tilde fences. The backtick-specific lookahead is kept so inline code spans are not treated as fences; tildes share the same trailing line/EOF handling.

## Compatibility notes

This changes parsing only for fenced code blocks at EOF when they interrupt a paragraph or GFM table row. Newline-terminated fences and inline code span cases such as ``a `b` c`` and ``foo\n`bar` `` were verified unchanged. The existing long-tilde ReDoS fixture was updated to the now-correct EOF-fence output while still covering the same large input.

## Validation

Empirical outputs after rebuild (`npm run build:esbuild`):

- `foo\n```` -> `<p>foo</p>\n<pre><code>\n</code></pre>\n`
- `foo\n```js` -> `<p>foo</p>\n<pre><code class="language-js">\n</code></pre>\n`
- `foo\n~~~` -> `<p>foo</p>\n<pre><code>\n</code></pre>\n`
- `foo\n~~~js` -> `<p>foo</p>\n<pre><code class="language-js">\n</code></pre>\n`
- newline-terminated fence inputs still produce the same fenced code blocks.
- `a \`b\` c` -> `<p>a <code>b</code> c</p>\n`
- `foo\n\`bar\`` -> `<p>foo\n<code>bar</code></p>\n`

Regression verification:

- With unmodified `src/rules.ts`, `node --test --test-reporter=spec test/run-spec-tests.js` reported `tests 1771`, `pass 1765`, `fail 6`; both `backtick_fence_eof_interrupts_paragraph` and `tilde_fence_eof_interrupts_paragraph` failed (plus the updated long-tilde ReDoS fixture).
- With the fix, `npm run test:specs` reported `tests 1771`, `pass 1771`, `fail 0`.
- `npm run test:types`: passed (exit 0; `attw` reports the existing ignored CJS-to-ESM warning only).
- `npm run test:lint`: passed (exit 0).
- `npm run test:umd`: passed (exit 0).
- `npm run test:cjs`: passed (exit 0).
- Pre-existing Windows path-with-spaces bin unit failures remain separate: `npm run test:unit` reported `tests 190`, `pass 185`, `fail 5` (`exec`, `stdin`, `string`, `input`, `input file positional`, `input last file positional` fail with `Cannot find module 'E:\Work\Code'`).


UziTech (approved):
Nice catch! 💯

styfle (approved):

PR #4046, opened 2026-08-07
fix: place task checkboxes after list loose is finalized

### LINKED ISSUE #4045: Task list checkbox rendered outside the <p> in loose nested lists, inconsistently within the same list
### Marked version

18.0.9 (browser, UMD build). Not present in 15.0.12.

### Describe the bug

In a **loose** task list that contains a nested list, the checkbox is sometimes emitted *outside* the paragraph :

```html
<li><input type="checkbox"> <p>Text</p>
```

instead of inside it :

```html
<li><p><input type="checkbox"> Text</p>
```

Since `<p>` is a block element, the checkbox ends up alone on its own line and the label drops below it.

What makes me think this is a bug rather than a deliberate change :

**it is inconsistent within a single list. ** In the sample below, three of the four task items get the checkbox outside the paragraph, and one gets it inside, for identical Markdown structure.

### To Reproduce

```markdown
* [ ] Prepare the project

  * [ ] Install dependencies
  * [x] Update Marked
  * [ ] Check the rendering
* [x] Run the tests

  * [x] Test lists
  * [ ] Test tables

    * [ ] Test nested cases
```

Output on 18.0.9 :

```html
<ul>
<li><input disabled="" type="checkbox"> <p>Prepare the project</p>
<ul>
<li><input disabled="" type="checkbox"> Install dependencies</li>
<li><input checked="" disabled="" type="checkbox"> Update Marked</li>
<li><input disabled="" type="checkbox"> Check the rendering</li>
</ul>
</li>
<li><p><input checked="" disabled="" type="checkbox"> Run the tests</p>
<ul>
<li><input checked="" disabled="" type="checkbox"> <p>Test lists</p>
</li>
<li><input disabled="" type="checkbox"> <p>Test tables</p>
<ul>
<li><input disabled="" type="checkbox"> Test nested cases</li>
</ul>
</li>
</ul>
</li>
</ul>
```

`Run the tests` is the odd one out : same structure as the others, but its checkbox stays inside the `<p>`.

Same input on 15.0.12, for comparison

```html
<ul>
<li><p><input disabled="" type="checkbox"> 
Prepare the project</p>
<ul>
<li><input disabled="" type="checkbox"> Install dependencies</li>
<li><input checked="" disabled="" type="checkbox"> Update Marked</li>
<li><input disabled="" type="checkbox"> Check the rendering</li>
</ul>
</li>
<li><p><input checked="" disabled="" type="checkbox"> 
Run the tests</p>
<ul>
<li><p><input checked="" disabled="" type="checkbox"> 
Test lists</p>
</li>
<li><p><input disabled="" type="checkbox"> 
Test tables</p>
<ul>
<li><input disabled="" type="checkbox"> Test nested cases</li>
</ul>
</li>
</ul>
</li>
</ul>
```

15.0.12 always keeps the checkbox inside the `<p>`, so it renders on one line. The stray newline after the `<input>` collapses to whitespace and makes no visual difference.

### Expected behavior

The checkbox stays inside the paragraph, as in 15.0.12, and the placement is at least consistent across items of the same list.

### Note

I bisected by testing every version from 13.0.0 to 18.0.9 on https://marked.js.org/demo/ (the version dropdown makes this easy).

Output is identical to 15.0.12 all the way through 16.X, then changes starting exactly at 17.0.0, and stays the same from 17.0.0 through 18.0.9.

So the regression was introduced in 17.0.0, though I haven’t looked at the commits to say which change is responsible : the checkbox-token/loose-list-paragraph changes mentioned in the 17.0.0 release notes look like the right area, but that’s a guess.

I ran the same corpus (~30 list variations) through both versions and this is the only case where the rendered output differs visually.
Everything else was whitespace-only changes in the source.

Hashim1999164:
## Summary
- Finalize `list.loose` (including spacer-based detection) before placing task checkboxes in `Tokenizer.list()`.
- Nested loose task lists no longer leave early item checkboxes outside `<p>` while later items nest correctly.
- Adds a regression fixture for the nested loose task list case from #4045.

Fixes #4045

## Test plan
- [x] `npm run build`
- [x] `npm run test:specs` (1781 passed, including new `list_loose_nested_tasks` and existing task/list fixtures)
- [x] Manually confirmed loose items render checkboxes inside `<p>`; tight nested items stay without `<p>`

### DIFF
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -416,9 +416,22 @@ export class _Tokenizer<ParserOutput = string, RendererOutput = string> {
       list.raw = list.raw.trimEnd();
 
       // Item child tokens handled here at end because we needed to have the final item to trim it first
+      // First pass: tokenize items and finalize list.loose from spacers before placing checkboxes
       for (const item of list.items) {
         this.lexer.state.top = false;
         item.tokens = this.lexer.blockTokens(item.text, []);
+
+        if (!list.loose) {
+          // Check if list should be loose
+          const spacers = item.tokens.filter(t => t.type === 'space');
+          const hasMultipleLineBreaks = spacers.length > 0 && spacers.some(t => this.rules.other.anyLine.test(t.raw));
+
+          list.loose = hasMultipleLineBreaks;
+        }
+      }
+
+      // Second pass: place task checkboxes using the final list.loose
+      for (const item of list.items) {
         const itemToken = item.tokens[0];
         if (item.task && (itemToken?.type === 'text' || itemToken?.type === 'paragraph')) {
           // Remove checkbox markdown from item tokens
@@ -459,14 +472,6 @@ export class _Tokenizer<ParserOutput = string, RendererOutput = string> {
           }
         } else if (item.task) {
           item.task = false;
-        }
-
-        if (!list.loose) {
-          // Check if list should be loose
-          const spacers = item.tokens.filter(t => t.type === 'space');
-          const hasMultipleLineBreaks = spacers.length > 0 && spacers.some(t => this.rules.other.anyLine.test(t.raw));
-
-          list.loose = hasMultipleLineBreaks;
         }
       }
 
--- a/test/specs/new/list_loose_nested_tasks.html
+++ b/test/specs/new/list_loose_nested_tasks.html
@@ -1,0 +1,20 @@
+<ul>
+<li><p><input disabled="" type="checkbox"> Prepare the project</p>
+<ul>
+<li><input disabled="" type="checkbox"> Install dependencies</li>
+<li><input checked="" disabled="" type="checkbox"> Update Marked</li>
+<li><input disabled="" type="checkbox"> Check the rendering</li>
+</ul>
+</li>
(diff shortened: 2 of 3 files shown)


UziTech (approved):
Thanks for fixing this! 💯

src/Tokenizer.ts
```diff
@@ -416,9 +416,22 @@ export class _Tokenizer<ParserOutput = string, RendererOutput = string> {
...
       for (const item of list.items) {
         this.lexer.state.top = false;
         item.tokens = this.lexer.blockTokens(item.text, []);
+
+        if (!list.loose) {
+          // Check if list should be loose
+          const spacers = item.tokens.filter(t => t.type === 'space');
+          const hasMultipleLineBreaks = spacers.length > 0 && spacers.some(t => this.rules.other.anyLine.test(t.raw));
+
+          list.loose = hasMultipleLineBreaks;
```
styfle, line 429: In a follow up, we could change this one be one line without 

```js
list.loose = item.tokens.some(t => t.type === 'space' && this.rules.other.anyLine.test(t.raw));
```

Though no need to change here since its just moving existing code

styfle (approved):

Issue #4050, opened 2026-08-10
CommonMark spec gaps grouped by root cause, and a question about lookup tables


luantaraschi:
I ran the spec suite on `main` (18.0.9, 681373c) and went through the cases carrying `shouldFail` to see what is actually left. Since `update-specs.js` sets that flag from whatever fails at the time it runs, the set is a snapshot of current behavior rather than a list of decisions, so I grouped it by root cause. Posting it in case it is useful for triage, and because two of the groups run into the same wall and that seems worth asking about before anyone starts on them.

CommonMark 0.31.2, `gfm: false`, 15 cases:

| Root cause | Examples | Count |
|---|---|---:|
| Character references in link destination and title | 32, 33, 503 | 3 |
| Nested brackets in link text | 512, 518, 519, 520, 528, 532, 533 | 7 |
| Raw HTML and autolinks taking precedence over links | 524, 526, 536, 538 | 4 |
| Unicode case folding of link labels | 540 | 1 |

Running the same file with `gfm: true` flags 4 more (602, 608, 611, 612), but those look like the autolink extension doing its job against a spec file written before it existed. For instance 611 expects `https://example.com` to stay plain text and marked links it, which is what GFM asks for. I would not count those as gaps.

`gfm.0.29` has 4: 633, 634 and 635 are `mailto:` and `xmpp:` autolinks, 657 is the disallowed raw HTML filter.

## The wall

The two smallest groups look approachable until you try them, and both need a lookup table, which sits badly with marked having no runtime dependencies.

Character references, current output on the left:

```
[foo](/f&ouml;&ouml; "f&ouml;&ouml;")
  <p><a href="/f&ouml;&ouml;" title="f&ouml;&ouml;">foo</a></p>
  <p><a href="/f%C3%B6%C3%B6" title="föö">foo</a></p>

[link](foo%20b&auml;)
  <p><a href="foo%20b&auml;">link</a></p>
  <p><a href="foo%20b%C3%A4">link</a></p>
```

Leaving the entity in place works for text content because the browser resolves it, but a destination has to be decoded first and then percent encoded, and getting there means the HTML5 named reference table.

Case folding is the same shape. Example 540 wants `[ẞ]` to match `[SS]`, which needs full folding to `ss`. I tried the usual shortcuts and neither gets there:

```js
'ẞ'.toLowerCase()                    // 'ß'
'SS'.toLowerCase()                   // 'ss'
[...]

PR #4040, opened 2026-08-02
fix: inlineTokens rebuilds the reflink-mask preamble per call

susyimes:
## Summary

inlineTokens rebuilds the reflink-mask preamble per call (Object.keys over all link defs), making n refs + n defs O(n^2) (measured exponent 3.0, 24s at n=13000)

## Evidence

- Defect reproduced on the true upstream source at release tag `18.0.7` (`reproduced_at_tag: True`).
- HEAD triage at time of verification: `unfixed_at_head` (ref `HEAD`).
- Pristine suite baseline: 1955 passed, 0 failed.
- Patched suite (this change + the regression test): 1957 passed, 0 failed.

The regression test `inlineTokens-masking.test.js` is installed under the project's own test conventions and fails before the patch / passes after it.

## Reproduction

A minimized standalone repro (`footnote_defs.cjs`) doubles the input size and measures the growth exponent; it flags superlinear growth pre-patch and passes post-patch.

## Provenance and disclosure

This pull request was prepared by an autonomous stewardship agent ([blackhole-agent](https://github.com/susyimes/blackhole-agent)). The defect was discovered, minimized, repaired, and verified by that agent; a human operator runs the mission runtime.

Sealed evidence bundle digests (sha256):

- `contribution.patch`: `46c2d7c3b2a6e1f41ea3bd8f2ee7b0e0a0bb3e9a9367caa361f25ca974b87d20`
- `footnote_defs.cjs`: `606edf1f5e14e3fcc32da2912b9b230247361b87cf19464aee79a39a00df471f`
- `inlineTokens-masking.test.js`: `af5a9ef040cb19b55f9ea6a774ed54ab11d7149200314c1c310989b1f5ffd3a4`

These digests seal the exact patch, regression test, and repro this PR carries, so the evidence chain can be re-checked byte-for-byte.

PR #4051, opened 2026-08-15
fix: do not nest a link inside a link

hdimer:
**Marked version:** `18.0.9` (current `master`, 9552b6b)

## Description

No existing issue for this one. It is the largest of the groups @luantaraschi catalogued in #4050 — or rather, half of it: "nested brackets in link text" is really two unrelated root causes, and this PR is the one that does not need a lookup table or a regex-depth change.

## Expectation

CommonMark §6.3: **"Links may not contain other links, at any level of nesting."** When a link's text contains a link, the outer brackets are literal text and the inner link wins.

```html
<p>[foo <a href="/uri">bar</a>](/uri)</p>
```

Both bundled reference implementations agree — `commonmark` 0.31.2 and `markdown-it` 15 (already dev-dependencies here) return exactly that.

## Result

```js
marked.parse('[foo [bar](/uri)](/uri)');
// <p><a href="/uri">foo <a href="/uri">bar</a></a></p>

marked.parse('[foo [bar](/uri)][ref]\n\n[ref]: /uri');
// <p><a href="/uri">foo <a href="/uri">bar</a></a></p>
```

The output is **nested `<a>` elements**. That is not merely off-spec, it is invalid HTML: `<a>` has no permitted `<a>` descendant, so an HTML parser closes the outer anchor at the inner one and re-parents the rest. The visible result in a browser is a link whose clickable region and destination are both wrong, and there is no workaround from the caller's side because the trigger is ordinary document text.

Fixes CommonMark examples **518, 519, 532** (`shouldFail` cleared in both `test/specs/commonmark/` and `test/specs/gfm/`).

## What was attempted

`outputLink` builds the token and lexes the label in one expression, so nothing ever looked at what the label turned into. The fix makes it look: if the label tokenized into a link, return `undefined` instead of a link token. `Lexer.inlineTokens` already guards both call sites with `if (token = this.tokenizer.link(src))` / `reflink(...)`, and both tokenizers already declared `| undefined`, so the bail falls through to `inlineText`, which emits the `[` literally and re-scans. The inner link is then tokenized normally.

Nesting is tracked with a `linkEmitted` lexer state flag saved and restored around the recursive `inlineTokens` call — the same shape `blockquote()` already uses for `state.top`. A flag rather than a walk of the returned tokens, because the rule has to see through `em`/`strong` (example 519 depends on it) while **not** firing on autolinks, and autolinks are `type: 'link'` too.

Three things I deliberately did not do:

- **Autolinks do not trigger the rule.** `[foo <https://example.com/>](/uri)` still produces nested anchors, because that is what both reference implementations do. Pinned as a test so it cannot drift.
- **Images are exempt** — their text is flattened into an `alt` attribute, so `[foo](uri2)` stays an image. Also pinned.
- **`state.inLink` still resets unconditionally**, exactly as before. Making it save/restore would change unrelated behaviour and is not needed here.

### Not fixed, same neighbourhood

Examples **512, 520, 528** stay red. They are the *other* root cause in that #4050 group: `_inlineLabel` tolerates only one level of bracket nesting, so `[link [foo [bar]]](/uri)` never matches in the first place. Nothing to do with link-in-link, and worth splitting out.

Example **533** (`[foo *bar [baz][ref]*][ref]`) goes from `<a href="/uri">foo <em>bar <a href="/uri">baz</a></em></a>` to `[foo *bar <a href="/uri">baz</a>*]<a href="/uri">ref</a>` — the nested anchor is gone, but the `<em>` does not form, because `reflinkSearch` masked the whole span before emphasis ran and the mask is now stale. That is the area #4040 and #4048 are already in, so I left it alone rather than collide.

## Verification

Built `master` and the patch side by side and diffed their output against `commonmark` 0.31.2.

- **40,256 parses** (structured nested-link templates × 8 wrappers × 2 depths, plus 20,000 pseudo-random documents over an alphabet of link/reflink/image/autolink/emphasis/code/bracket atoms, each under `gfm:false` and `gfm:true`): **154 fixed, 0 broken**, 19 changed-but-still-wrong — and all 19 are the example-533 masking shape above, where the patch is strictly closer to the reference than `master` was.
- **Escaping oracle**, 7,056 parses over link texts seeded with `<pre>`/`<code>`/`<script>`/`<kbd>` and `&`/`<`/`"` filler: **0 cases** where the patch emits fewer HTML entities than `master`. See the note below for why that oracle exists.
- `test:specs` (1779), `test:unit` (191), `test:umd`, `test:cjs`, `test:types` and `test:lint` all green. Same caveat others have reported: `quadratic_emstrong_delim[0]` intermittently blows its 1s budget on my laptop, but it does so on a clean unpatched `master` build too, so it is this machine and not the patch.
- **Perf**: nested-link-heavy input 1.06–1.13×, plain links within noise, pathological nested reflinks ~1.5× (the rejected label is tokenized once and thrown away). No blowup with depth — `_inlineLabel`'s one-level nesting limit caps the discard-and-rescan at ~2 frames; measured flat from depth 4 to 22.

The `<pre>` unit test is not padding. The discarded tokenization pass runs `tag()` over the whole label, and `tag()` sets `state.inRawBlock`. Restoring `linkEmitted` alone left that set on the bail path, so the re-scan stamped `escaped: true` on text *preceding* the raw-block opener and `[a & b <pre> [x](/1)](/2)` came out with a bare `&`. With a malformed tag in the label it turned text that `master` escapes into live markup. Hence `state.inRawBlock = outerInRawBlock` before the `return`, and an exact-string unit test — the spec harness normalizes entities through cheerio, so it cannot see this class of bug at all.

### Two things to flag rather than bury

**Escaped brackets in a label.** `outputLink` unescapes `cap[1]` *before* lexing it (`outputLinkReplace`), so a label containing `\[q\](/e)` is handed to the lexer as `[q](/e)` and tokenizes as a link that the source never contained. The rejection path turns that latent mis-tokenization into visible damage in one shape:

```
![[\[q\](/e)][ref]](/i)   with   [ref]: /r
  master:  <img src="/i" alt="[q](/e)">   (matches commonmark)
  patched: <img src="/i" alt="[q]ref">
```

On 2,560 parses from a generator built specifically to be adversarial here, it is **764 improved against 14 regressed**, and all 14 are that one shape — a reflink inside an image label. I tried two local fixes: lexing `cap[1]` instead of `text` swaps these 14 for 220 worse ones (raw backslashes leak into `alt`), and gating on `text === cap[1]` is inert because the unescaping is lossy one frame up. It is a label-handling defect, not a nesting one, and it belongs with 512/520/528. Happy to be told it blocks this PR.

**Stateful inline extensions fire twice** on a rejected link, since the label is tokenized once for the probe and again on the re-scan. Extensions that count things (footnote numbering, ID generation) will double-count inside `[foo @@ [x](/1)](/2)`. Inherent to probe-then-rescan; flagging it because the extension ecosystem is the constituency that eats it.

**Type note:** `state` is public in `lib/marked.d.ts`, so `linkEmitted` is an additive field. Extensions that read state are unaffected; anything assigning a whole state object literal would need the new key.

## Contributor

Ticket type: **L1 - broken** by the CONTRIBUTING table, I think — wrong output against a supported spec, and no caller-side workaround. Happy to be corrected to L2.

## Committer

---
Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -13,6 +13,8 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
   state: {
     inLink: boolean;
     inRawBlock: boolean;
+    /** a link was produced in the inline run currently being scanned */
+    linkEmitted: boolean;
     top: boolean;
   };
 
@@ -33,6 +35,7 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
     this.state = {
       inLink: false,
       inRawBlock: false,
+      linkEmitted: false,
       top: true,
     };
 
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -11,22 +11,42 @@ import type { _Lexer } from './Lexer.ts';
 import type { Links, Tokens, Token } from './Tokens.ts';
 import type { MarkedOptions } from './MarkedOptions.ts';
 
-function outputLink(cap: string[], link: Pick<Tokens.Link, 'href' | 'title'>, raw: string, lexer: _Lexer, rules: Rules): Tokens.Link | Tokens.Image {
+function outputLink(cap: string[], link: Pick<Tokens.Link, 'href' | 'title'>, raw: string, lexer: _Lexer, rules: Rules): Tokens.Link | Tokens.Image | undefined {
   const href = link.href;
   const title = link.title || null;
   const text = cap[1].replace(rules.other.outputLinkReplace, '$1');
+  const isImage = cap[0].charAt(0) === '!';
 
   lexer.state.inLink = true;
-  const token: Tokens.Link | Tokens.Image = {
-    type: cap[0].charAt(0) === '!' ? 'image' : 'link',
+  const outerLinkEmitted = lexer.state.linkEmitted;
+  const outerInRawBlock = lexer.state.inRawBlock;
+  lexer.state.linkEmitted = false;
+  const tokens = lexer.inlineTokens(text);
+  const textHasLink = lexer.state.linkEmitted;
+  lexer.state.linkEmitted = outerLinkEmitted;
+  lexer.state.inLink = false;
+
+  if (!isImage) {
+    // CommonMark: "Links may not contain other links, at any level of nesting."
+    // Bail so the caller falls through to text and the inner link is the one kept.
+    // Images are exempt: their text is flattened into an alt attribute.
+    if (textHasLink) {
+      // these tokens are discarded, so undo the raw-block state they opened;
+      // leaving it set would suppress escaping for the text that is re-scanned
+      lexer.state.inRawBlock = outerInRawBlock;
+      return;
+    }
+    lexer.state.linkEmitted = true;
+  }
+
+  return {
+    type: isImage ? 'image' : 'link',
     raw,
     href,
     title,
     text,
-    tokens: lexer.inlineTokens(text),
+    tokens,
   };
-  lexer.state.inLink = false;
-  return token;
 }
 
 function indentCodeCompensation(raw: string, text: string, rules: Rules) {
--- a/test/specs/commonmark/commonmark.0.31.2.json
+++ b/test/specs/commonmark/commonmark.0.31.2.json
@@ -4145,8 +4145,7 @@
     "example": 518,
     "start_line": 7886,
     "end_line": 7890,
-    "section": "Links",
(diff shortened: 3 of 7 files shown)


UziTech (approved):
Thanks! 💯

styfle (approved):

Issue #33, opened 2012-02-17
sequence of <em>'s is mis-rendered in some cases.


leeoniya:
by itself,

``` md
#### *expression<sub>1</sub>* *expression<sub>2</sub>* ... *expression<sub>n</sub>*
```

renders correctly as

``` html
<h4><em>expression<sub>1</sub></em> <em>expression<sub>2</sub></em> ... <em>expression<sub>n</sub></em></h4>
```

but in the context of https://raw.github.com/dmajda/pegjs/master/README.md renders as

``` html
<h4><em>expression<sub>1</sub></em> / <em>expression<sub>2</sub></em> / ... / <em>expression<sub>n</sub></em></h4>
```

maybe a parser issue?
thanks!

Issue #32, opened 2012-02-16
Marked crash Firefox (Windows XP)


alejandro:
I've been using marked for a while, but today I was doing cross-platform test and I noticed that Firefox crashed when marked tried to parse my *.md file. Firefox just crash. I test it in chrome (still on windows XP) and it works fine. Also I test it on Firefox (Windows 7) and works fine. I don't know what is the main problem in here. 

ps. Firefox v6.0

PR #4053, opened 2026-08-15
fix: escape character references in autolink destinations

### LINKED ISSUE #4052: `&` not HTML-escaped in link/autolink href, silently changing the link target
**Marked version:** 18.0.9 (also reproduces on every version tested, back to v12)

**Repro**
```js
marked('<https://example.com/?x=1&lt;2>');
```

**Expected** (commonmark.js and CommonMark spec [example 595](https://spec.commonmark.org/0.31.2/#example-595) — `&` in the destination is escaped to `&amp;`):
```html
<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
```

**Actual** (raw `&` in the href):
```html
<p><a href="https://example.com/?x=1&lt;2">https://example.com/?x=1&lt;2</a></p>
```

Because `&lt;` is a valid HTML character reference, a browser decodes it inside the attribute, so the link actually points to `https://example.com/?x=1<2` — a different URL than was written. Confirmed by parsing marked's output with parse5:
```
marked href, parsed back: "https://example.com/?x=1<2"     (wrong)
commonmark href, parsed:  "https://example.com/?x=1&lt;2"   (correct)
```
Note marked already escapes the same URL in the anchor *text* (`&amp;`) but not in the `href`, so the output is internally inconsistent. Any URL containing a valid entity sequence (`&lt;`, `&amp;`, `&copy;`, …) is affected; this is silent link-target corruption for user- or attacker-supplied URLs.

**Cause**
`Renderer.link()` writes `cleanUrl(href)` straight into the attribute without `escapeHtmlEntities()` — while the `title` attribute one line below *is* escaped. `cleanUrl` only runs `encodeURI`, which never encodes `&`, so the raw `&` reaches the attribute. (`Renderer.image()` has the same pattern for `src`.)

Kjubikstronk:
Fixes #4052.

Following up on [my comment there](https://github.com/markedjs/marked/issues/4052#issuecomment-5302818703) — the one-line change the issue suggests regresses inline links, so this takes the narrower route instead.

## Why it is autolink-specific

CommonMark resolves character references in [link destinations](https://spec.commonmark.org/0.31.2/#link-destination) but [not in autolinks](https://spec.commonmark.org/0.31.2/#autolinks) ("backslash-escapes and entity references do not work in autolinks"). marked never resolves them anywhere — `token.href` keeps the literal source text — and that produces two different outcomes:

- **Inline link** `[t](…?x=1&lt;2)` — the destination value should be `?x=1<2`, which is written back into an attribute as `&lt;`. The unresolved source text is already exactly that, so the current output is right.
- **Autolink** `<…?x=1&lt;2>` — the destination value is the literal text `?x=1&lt;2`, which has to be written as `&amp;lt;`. Emitting it raw lets the browser decode it, so the link points at `?x=1<2`.

That is why escaping every destination breaks the first case while fixing the second:

| Input | before | escape everything | this PR |
| --- | --- | --- | --- |
| `<https://example.com/?x=1&lt;2>` | `?x=1&lt;2` ❌ | `?x=1&amp;lt;2` ✅ | `?x=1&amp;lt;2` ✅ |
| `[t](https://example.com/?a=1&amp;b=2)` | `?a=1&amp;b=2` ✅ | `?a=1&amp;amp;b=2` ❌ | `?a=1&amp;b=2` ✅ |

## The change

`Tokenizer.autolink()` and `Tokenizer.url()` now set `autolink: true` on the token, and `Renderer.link()` escapes the destination and the text with `escapeHtmlEntities(…, true)` when it is set. The flag is an optional addition to `Tokens.Link`, so existing custom renderers and extensions are unaffected.

The text needed it too, not just the href — for `<…?x=1&lt;2>` the anchor text was also rendering as `&lt;`, because the default text escaping deliberately skips anything that already looks like a reference. Both halves now match the expected output in the issue.

## This also fixes CommonMark example 595

[Example 595](https://spec.commonmark.org/0.31.2/#example-595) is in `test/specs/commonmark` with `shouldFail: false`, so it is expected to pass — but marked's actual output for it was:

```html
<a href="https://foo.bar.baz/test?q=hello&id=22&boolean">…&amp;id=22&amp;boolean</a>
```

against an expected `href` of `…?q=hello&amp;id=22&amp;boolean`. The href and the link text disagreed with each other.

It was reported as passing because the spec runner compares with `HtmlDiffer`, which parses both sides — and in an attribute, `&id=22` and `&amp;id=22` parse to the same value. That normalisation hides the difference whenever the text is not a valid reference. It only becomes visible with something like `&lt;`, where the two parse to *different* values, which is what the issue reporter demonstrated with parse5.

So the spec suite could not have caught this, and still cannot. I've written the new tests as exact string comparisons in `test/unit/marked.test.js` rather than adding spec fixtures, for that reason.

I have deliberately not touched the unresolved-reference behaviour itself — examples 32 and 33 are still marked `shouldFail`, and `[foo](/f&ouml;&ouml;)` still renders unchanged. That is the larger gap tracked in #4050 and wants its own discussion.

## Verification

- `test:specs` — 1779 pass, 0 fail, unchanged before and after.
- `test:unit` — 189 pass, up from 185; the 5 `exec` failures are pre-existing on `main` and fail identically with this change stashed.
- `tsc --noEmit` and `eslint` clean on the touched files.
- Removing the source change makes 3 of the 4 new tests fail; the fourth is the inline-link guard, which passes either way by design.

The four `Lexer` token assertions for autolinks and urls were updated for the new field.

### DIFF
--- a/test/specs/new/link_destination_character_references.html
+++ b/test/specs/new/link_destination_character_references.html
@@ -1,7 +1,15 @@
 <p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
+<p><a href="https://example.com/?y=1&amp;amp;2">https://example.com/?y=1&amp;amp;2</a></p>
+<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
 <p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
-<p><a href="https://example.com/?a=1&amp;b=2">t</a></p>
-<p><a href="http://example.com?foo=1&amp;bar=2">example</a></p>
-<p><img src="http://example.com?foo=1&amp;bar=2" alt="i"></p>
-<p><a href="http://example.com?a=1&lt;2">t</a></p>
-<p><a href="https://foo.bar.baz/test?q=hello&amp;id=22&amp;boolean">https://foo.bar.baz/test?q=hello&amp;id=22&amp;boolean</a></p>
+<p><a href="https://example.com/?y=1&amp;amp;2">https://example.com/?y=1&amp;amp;2</a></p>
+<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
+<p><a href="https://example.com/?x=1&lt;2">https://example.com/?x=1&lt;2</a></p>
+<p><a href="https://example.com/?y=1&amp;2">https://example.com/?y=1&amp;2</a></p>
+<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
+<p><a href="https://example.com/?x=1&lt;2">https://example.com/?x=1&lt;2</a></p>
+<p><a href="https://example.com/?y=1&amp;2">https://example.com/?y=1&amp;2</a></p>
+<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
+<p><img src="https://example.com/?x=1&lt;2" alt="https://example.com/?x=1&lt;2"></p>
+<p><img src="https://example.com/?y=1&amp;2" alt="https://example.com/?y=1&amp;2"></p>
+<p><img src="https://example.com/?a=1&amp;b=2" alt="https://example.com/?a=1&amp;b=2"></p>
--- a/test/specs/new/link_destination_character_references.html
+++ b/test/specs/new/link_destination_character_references.html
@@ -1,0 +1,7 @@
+<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
+<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
+<p><a href="https://example.com/?a=1&amp;b=2">t</a></p>
+<p><a href="http://example.com?foo=1&amp;bar=2">example</a></p>
+<p><img src="http://example.com?foo=1&amp;bar=2" alt="i"></p>
+<p><a href="http://example.com?a=1&lt;2">t</a></p>
+<p><a href="https://foo.bar.baz/test?q=hello&amp;id=22&amp;boolean">https://foo.bar.baz/test?q=hello&amp;id=22&amp;boolean</a></p>
--- a/test/unit/marked.test.js
+++ b/test/unit/marked.test.js
@@ -10,64 +10,6 @@ describe('marked unit', () => {
     setOptions(getDefaults());
   });
 
-  describe('Character references in link destinations', () => {
-    // Character references are not resolved inside an autolink, so a `&` there
-    // is literal and has to be escaped. These assert on the exact output
-    // because the spec runner compares parsed HTML, where `&lt;` and `&amp;lt;`
-    // in an attribute are indistinguishable.
-    // https://github.com/markedjs/marked/issues/4052
-
-    it('should escape an ampersand in an autolink', () => {
-      assert.strictEqual(
-        marked.parse('<https://example.com/?x=1&lt;2>').trim(),
-        '<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>',
-      );
-    });
-
-    it('should escape an ampersand in an extended url', () => {
-      assert.strictEqual(
-        marked.parse('https://example.com/?x=1&lt;2').trim(),
-        '<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>',
-      );
-    });
-
-    it('should keep the destination of an inline link unchanged', () => {
-      assert.strictEqual(
-        marked.parse('[t](https://example.com/?a=1&amp;b=2)').trim(),
-        '<p><a href="https://example.com/?a=1&amp;b=2">t</a></p>',
-      );
-    });
-
-    it('should escape an ampersand that cannot start a reference', () => {
-      assert.strictEqual(
-        marked.parse('[example](http://example.com?foo=1&bar=2)').trim(),
-        '<p><a href="http://example.com?foo=1&amp;bar=2">example</a></p>',
-      );
-    });
-
-    it('should escape an ampersand in an image source', () => {
-      assert.strictEqual(
-        marked.parse('![i](http://example.com?foo=1&bar=2)').trim(),
-        '<p><img src="http://example.com?foo=1&amp;bar=2" alt="i"></p>',
-      );
-    });
-
-    it('should leave a reference in an inline destination alone', () => {
-      assert.strictEqual(
-        marked.parse('[t](http://example.com?a=1&lt;2)').trim(),
-        '<p><a href="http://example.com?a=1&lt;2">t</a></p>',
-      );
-    });
-
-    it('should match the CommonMark autolink example', () => {
-      // https://spec.commonmark.org/0.31.2/#example-595
-      assert.strictEqual(
-        marked.parse('<https://foo.bar.baz/test?q=hello&id=22&boolean>').trim(),
-        '<p><a href="https://foo.bar.baz/test?q=hello&amp;id=22&amp;boolean">https://foo.bar.baz/test?q=hello&amp;id=22&amp;boolean</a></p>',
-      );
-    });
-  });
-
   describe('Test paragraph token type', () => {
     it('should use the "paragraph" type on top level', () => {
       const md = 'A Paragraph.\n\n> A blockquote\n\n- list item\n';
--- a/test/unit/marked.test.js
+++ b/test/unit/marked.test.js
@@ -10,6 +10,43 @@ describe('marked unit', () => {
     setOptions(getDefaults());
   });
 
+  describe('Character references in link destinations', () => {
+    // Character references are not resolved inside an autolink, so a `&` there
+    // is literal and has to be escaped. These assert on the exact output
+    // because the spec runner compares parsed HTML, where `&lt;` and `&amp;lt;`
+    // in an attribute are indistinguishable.
+    // https://github.com/markedjs/marked/issues/4052
+
+    it('should escape an ampersand in an autolink', () => {
+      assert.strictEqual(
+        marked.parse('<https://example.com/?x=1&lt;2>').trim(),
+        '<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>',
+      );
+    });
+
+    it('should escape an ampersand in an extended url', () => {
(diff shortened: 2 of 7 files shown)


test/unit/marked.test.js
```diff
@@ -10,6 +10,43 @@ describe('marked unit', () => {
...
   });
 
+  describe('Character references in link destinations', () => {
+    // Character references are not resolved inside an autolink, so a `&` there
+    // is literal and has to be escaped. These assert on the exact output
+    // because the spec runner compares parsed HTML, where `&lt;` and `&amp;lt;`
+    // in an attribute are indistinguishable.
+    // https://github.com/markedjs/marked/issues/4052
+
+    it('should escape an ampersand in an autolink', () => {
```
UziTech: Any tests that test the output should be in https://github.com/markedjs/marked/tree/master/test/specs/new

UziTech (changes requested):

UziTech:
Looks like this also doesn't follow CommonMark for

```markdown
[example](http://example.com?foo=1&bar=2)
```

[CommonMark demo](https://spec.commonmark.org/dingus/?text=%5Bexample%5D(http%3A%2F%2Fexample.com%3Ffoo%3D1%26bar%3D2))

```html
<p><a href="http://example.com?foo=1&amp;bar=2">example</a></p>
```

Kjubikstronk:
You're right, and my reasoning in the description was too narrow. I checked that a destination containing a *reference* round-trips correctly and concluded inline destinations were fine, but a bare `&` is a different case and it was still being written out raw.

The two need different treatment, and `escapeHtmlEntities` already has both modes:

- **inline destination** — the text is kept as written, so `&lt;` is already right in an attribute and must not be touched, while `&bar=2` cannot begin a reference and has to become `&amp;bar=2`. That is the default mode.
- **autolink** — references are not resolved, so every `&` is literal and all of them are escaped. That is the `encode` mode.

Which is a one-line change now that the autolink flag exists:

```js
href = escapeHtmlEntities(cleanHref, autolink);
```

Image sources had the same gap, so they are escaped too.

```
[example](http://example.com?foo=1&bar=2)   ->  href="http://example.com?foo=1&amp;bar=2"
![i](http://example.com?foo=1&bar=2)        ->  src="http://example.com?foo=1&amp;bar=2"
[t](http://example.com?a=1&lt;2)            ->  href="http://example.com?a=1&lt;2"        (unchanged)
[t](http://example.com?a=1&amp;b=2)         ->  href="http://example.com?a=1&amp;b=2"     (unchanged)
<https://example.com/?x=1&lt;2>             ->  href="https://example.com/?x=1&amp;lt;2"
```

Tests added for the bare-ampersand case in both a link and an image, plus one pinning that a reference in an inline destination is left alone.

`test:specs` 1779 pass / 0 fail, unchanged. `test:unit` 192 pass, up from 189; the 5 `exec` failures are pre-existing on `main` and fail identically with this change stashed.

test/unit/marked.test.js
Kjubikstronk: I tried that first and the spec runner can't see this class of bug. I put the autolink case in `test/specs/new` with `&amp;` in the expected href, reverted the fix, and it still passed — the runner compares parsed HTML, so `href="...&amp;id=22"` and `href="...&id=22"` are the same attribute value by the time it looks at them.

The unit tests do catch it. With the fix reverted, four of them fail. That's why they assert on the raw output string, and why I left the comment above them.

Happy to add a spec fixture too if it's worth having as documentation, but I didn't want to add a test that passes either way. Which would you prefer?

UziTech: You can use renderExact to not use the html differ
https://github.com/markedjs/marked/blob/master/test/specs/new/tabs_code.md

Kjubikstronk: `renderExact` was exactly what I was missing, thanks. All seven cases are now in `test/specs/new/link_destination_character_references.md` and I've checked it fails with the fix reverted, so it does catch the regression.

I've left the unit tests in place for now — say the word and I'll drop them, since the spec fixture covers the same ground.

UziTech: Ya lets remove the unit tests so they are cleaner

test/specs/new/link_destination_character_references.html
```diff
@@ -0,0 +1,7 @@
+<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
+<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
+<p><a href="https://example.com/?a=1&amp;b=2">t</a></p>
+<p><a href="http://example.com?foo=1&amp;bar=2">example</a></p>
+<p><img src="http://example.com?foo=1&amp;bar=2" alt="i"></p>
+<p><a href="http://example.com?a=1&lt;2">t</a></p>
```
UziTech: According to this [CommonMark demo](https://spec.commonmark.org/dingus/?text=%5Bt%5D(http%3A%2F%2Fexample.com%3Fa%3D1%26lt%3B2)) this line should be

```
<p><a href="http://example.com?a=1%3C2">t</a></p>
```

test/unit/marked.test.js
Kjubikstronk: Done — unit tests removed, the spec fixture covers all seven cases. Suite is green: 1781 spec tests, 44 unit tests, 0 failures.

UziTech (changes requested):

test/specs/new/link_destination_character_references.html
UziTech: Can you use these as the tests:

```
---
renderExact: true
---
https://example.com/?x=1&lt;2

https://example.com/?y=1&amp;2

https://example.com/?a=1&b=2

<https://example.com/?x=1&lt;2>

<https://example.com/?y=1&amp;2>

<https://example.com/?a=1&b=2>

[https://example.com/?x=1&lt;2](https://example.com/?x=1&lt;2)

[https://example.com/?y=1&amp;2](https://example.com/?y=1&amp;2)

[https://example.com/?a=1&b=2](https://example.com/?a=1&b=2)

[https://example.com/?x=1&lt;2][link1]

[https://example.com/?y=1&amp;2][link2]

[https://example.com/?a=1&b=2][link3]

![https://example.com/?x=1&lt;2](https://example.com/?x=1&lt;2)

![https://example.com/?y=1&amp;2](https://example.com/?y=1&amp;2)

![https://example.com/?a=1&b=2](https://example.com/?a=1&b=2)

[link1]: https://example.com/?x=1&lt;2
[link2]: https://example.com/?y=1&amp;2
[link3]: https://example.com/?a=1&b=2
```

```
<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
<p><a href="https://example.com/?y=1&amp;amp;2">https://example.com/?y=1&amp;amp;2</a></p>
<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
<p><a href="https://example.com/?x=1&amp;lt;2">https://example.com/?x=1&amp;lt;2</a></p>
<p><a href="https://example.com/?y=1&amp;amp;2">https://example.com/?y=1&amp;amp;2</a></p>
<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
<p><a href="https://example.com/?x=1%3C2">https://example.com/?x=1&lt;2</a></p>
<p><a href="https://example.com/?y=1&amp;2">https://example.com/?y=1&amp;2</a></p>
<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
<p><a href="https://example.com/?x=1%3C2">https://example.com/?x=1&lt;2</a></p>
<p><a href="https://example.com/?y=1&amp;2">https://example.com/?y=1&amp;2</a></p>
<p><a href="https://example.com/?a=1&amp;b=2">https://example.com/?a=1&amp;b=2</a></p>
<p><img src="https://example.com/?x=1%3C2" alt="https://example.com/?x=1&lt;2"></p>
<p><img src="https://example.com/?y=1&amp;2" alt="https://example.com/?y=1&amp;2"></p>
<p><img src="https://example.com/?a=1&amp;b=2" alt="https://example.com/?a=1&amp;b=2"></p>
```

UziTech:
Interestingly GitHub doesn't parse this one correctly

```
<https://example.com/?y=1&amp;2>
```

[CommonMark](https://spec.commonmark.org/dingus/?text=%3Chttps%3A%2F%2Fexample.com%2F%3Fy%3D1%26amp%3B2%3E)

```
<p><a href="https://example.com/?y=1&amp;amp;2">https://example.com/?y=1&amp;amp;2</a></p>
```

GitHub <https://example.com/?y=1&amp;2>

```
<p><a href="https://example.com/?y=1&amp;2">https://example.com/?y=1&amp;2</a></p>
```

test/specs/new/link_destination_character_references.html
UziTech: updated the tests to add reflinks and show the difference between the href and text of some

Kjubikstronk: Swapped in your cases, thanks — the reflinks and the href/text split make the autolink-vs-inline difference much clearer than what I had.

On the `%3C` one: you're right that CommonMark gives `http://example.com?a=1%3C2` there, but that line comes out the same on `master` as on this branch. I diffed the whole fixture both ways — this PR only changes the six autolink lines and the `?a=1&b=2` cases, where `master` emits a raw `&` in the href. The three `&lt;` inline/reference/image destinations are byte-identical before and after.

So the missing percent-encoding looks like a separate gap rather than something I've introduced. Happy to take it on, either here or as its own PR — just say which you'd prefer, since it's a different code path from the entity escaping.

UziTech: Can you add a comment in the file about the lines that don't match CommonMark

Kjubikstronk: Added it to the frontmatter — `front-matter` runs the block through YAML, so `#` comments are stripped and never reach the rendered output. The fixture still passes byte for byte.

It calls out that the `&lt;` inline links, reference links and images are the ones that differ: CommonMark resolves the reference and percent-encodes it to `?x=1%3C2` where marked keeps `?x=1&lt;2`. I noted that this comes from URL encoding rather than from the escaping in this PR, and that it is unchanged on `master`, so nobody reads those lines later as something this PR introduced.

UziTech (approved):
Nice work! 💯

Kjubikstronk:
This has been green and approved since yesterday — anything else you'd like on it before it goes in?

Asking because a few CommonMark link cases are still marked `shouldFail` in the spec files (#32, #33 and #503, the ones where `&ouml;` in a destination should be decoded and percent-encoded). I'd like to look at those next, but they go through `outputLink` in `Tokenizer.ts` too, so I'd rather not stack a second PR on this one while it's pending.

UziTech:
Just waiting on additional reviews. We typically like getting at least 2 reviews from MarkedJS team members.

PR #4060, opened 2026-08-19
test: add test case for parseInline with empty input

aoright:
### Summary
- In `test/unit/marked.test.js`, add a unit test verifying that `parseInline('')` returns an empty string.


UziTech:
why?

PR #4063, opened 2026-08-20
fix: reject GFM email autolink when the domain ends in _ or -

contactjawad:
The GFM email autolink rule mangles an address when its domain is immediately followed by `_` or `-`.

The `email` sub-pattern in the GFM inline `url` rule ends with a `(?![-_])` lookahead. Since that only forbids `-`/`_` *after* the final domain character, the regex backtracks and gives that character back to satisfy the lookahead — truncating the last domain label and producing a wrong `mailto:` target plus a visibly split word (e.g. `foo@bar.com-` autolinks `foo@bar.co` and leaves `m-` behind).

Per the GFM spec a domain label must not end in `_` or `-`. Changing the lookahead to `(?![\w-])` also forbids a following word character, forcing the final label to match maximally; when the maximal domain is then followed by `_`/`-`, the whole autolink fails and the text is rendered as-is — which is the correct behaviour.

Added a unit test covering `foo@bar.com_` and `foo@bar.com-`, both of which now tokenize as plain text. It fails on `master` and passes with this change.

### DIFF
--- a/test/unit/Lexer.test.js
+++ b/test/unit/Lexer.test.js
@@ -2118,6 +2118,36 @@ paragraph
             ],
           });
         });
+
+        it('url email domain ending in _ or - is not autolinked (GFM)', () => {
+          // A domain must not end in `_` or `-`; the whole email autolink is
+          // rejected and rendered as plain text, with the domain kept intact.
+          expectInlineTokens({
+            md: 'foo@bar.com_',
+            options: { gfm: true },
+            tokens: [
+              {
+                type: 'text',
+                raw: 'foo@bar.com_',
+                text: 'foo@bar.com_',
+                escaped: false,
+              },
+            ],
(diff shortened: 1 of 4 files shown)


test/unit/Lexer.test.js
UziTech, line 1: Can you remove this test and add a test in https://github.com/markedjs/marked/tree/master/test/specs/new?

UziTech (changes requested):

test/unit/Lexer.test.js
contactjawad: Done — moved it to a spec-file pair (test/specs/new/gfm_email_autolink_trailing.md / .html) and removed the unit test. It fails on master and passes with the fix.

UziTech (approved):

PR #4065, opened 2026-08-20
fix: escape HTML entities in link href and image src attributes

### LINKED ISSUE #4052: `&` not HTML-escaped in link/autolink href, silently changing the link target

madebysaira:
## Summary

Fixes #4052

The renderer wrote `cleanUrl(href)` directly into `href`/ attributes without escaping `&` to `&amp;`. This caused silent link-target corruption for URLs containing valid HTML entity sequences (`&lt;`, `&amp;`, `&copy;`, etc.).

### The bug

```js
marked('<https://example.com/?x=1&lt;2>')
// Actual:   href="...?x=1&lt;2"    (browser decodes to ...?x=1<2 — WRONG)
// Expected: href="...?x=1&amp;lt;2" (browser decodes to ...?x=1&lt;2 — CORRECT)
```

Any URL containing a valid entity sequence silently changes the link target. This also affects CommonMark example 595.

### Root cause

Two issues:

1. **Renderer** — `cleanUrl()` runs `encodeURI()` which never encodes `&`, so the raw `&` reaches the HTML attribute unescaped.
2. **Tokenizer** — inline link destinations (`[text](url)`) kept HTML entities literal (`&amp;` stayed as `&amp;`), which is correct per CommonMark, but the renderer never escaped them on output.

### Fix (2 parts)

1. **Tokenizer** (`src/Tokenizer.ts`) — Decode HTML entities in inline link destinations before rendering, per CommonMark spec (entity references are recognized in link destinations but not in autolinks).

2. **Renderer** (`src/Renderer.ts`) — Escape `&` → `&amp;` in `href`/ attributes after `cleanUrl()`, ensuring valid HTML output.

3. **Helper** (`src/helpers.ts`) — Added `decodeHtmlEntities()` function for named/numeric character reference decoding.

### Test results

- ✅ All 1783 CommonMark spec tests pass (no regressions)
- ✅ All 196 unit tests pass (including 5 new regression tests)
- ✅ CommonMark example 595 now produces correct output

### Regression tests added

5 new test cases in `test/unit/marked.test.js` covering:
- Autolink with `&` (CommonMark example 595)
- Inline link with `&amp;` entity
- Plain autolink URL with `&`
- Autolink with `&lt;` entity
- Image with `&` in src

### DIFF
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -163,7 +163,7 @@ export class _Renderer<ParserOutput = string, RendererOutput = string> {
     if (cleanHref === null) {
       return text as RendererOutput;
     }
-    href = cleanHref;
+    href = cleanHref.replace(/&/g, '&amp;');
     let out = '<a href="' + href + '"';
     if (title) {
       out += ' title="' + (escapeHtmlEntities(title)) + '"';
@@ -180,7 +180,7 @@ export class _Renderer<ParserOutput = string, RendererOutput = string> {
     if (cleanHref === null) {
       return escapeHtmlEntities(text) as RendererOutput;
     }
-    href = cleanHref;
+    href = cleanHref.replace(/&/g, '&amp;');
 
     let out = `<img src="${href}" alt="${escapeHtmlEntities(text)}"`;
     if (title) {
--- a/src/Tokenizer.ts
+++ b/src/Tokenizer.ts
@@ -5,6 +5,7 @@ import {
   findClosingBracket,
   expandTabs,
   trimTrailingBlankLines,
+  decodeHtmlEntities,
 } from './helpers.ts';
 import type { Rules } from './rules.ts';
 import type { _Lexer } from './Lexer.ts';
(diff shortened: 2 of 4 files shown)


UziTech:
This is already being fixed in #4053

PR #4068, opened 2026-08-21
fix: apply GFM tagfilter to raw HTML

evgenyponomarev:
**Marked version:** master

## Description

Addresses the remaining tagfilter subgroup of #4050 (GFM 0.29 example 657).

When GFM is enabled, the default HTML renderer now replaces only the leading `<` of the nine tag names defined by the GFM Disallowed Raw HTML extension. Matching is case-insensitive, supports opening and closing tags in inline or block HTML, and requires a complete tag-name boundary. Other HTML remains untouched, and `gfm: false` preserves the previous output.

This is deliberately not a sanitizer; it implements only the narrow GFM tagfilter rule. The existing sanitization guidance remains applicable.

The GFM copy of the CommonMark fixtures now marks examples 170–173, 176, and 178 as intentional divergences because tagfilter changes their raw `script`, `textarea`, and `style` output. The original CommonMark suite remains 44/44 for HTML blocks. This follows the existing convention used for CommonMark examples changed by the GFM autolink extension.

AI disclosure: OpenAI Codex assisted with implementation and validation; I reviewed and own the contribution.

## Contributor

- [x] This implements existing GFM behavior, so no feature documentation is required.

## Testing

- `npm test` — 1,783 spec tests and 193 unit tests passed, plus UMD/CJS, types, package checks, and lint.
- `npm run rules`
- `npm run build:reset`

## Committer

### DIFF
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -46,6 +46,9 @@ export class _Renderer<ParserOutput = string, RendererOutput = string> {
   }
 
   html({ text }: Tokens.HTML | Tokens.Tag): RendererOutput {
+    if (this.options.gfm) {
+      return text.replace(other.gfmTagFilter, '&lt;') as RendererOutput;
+    }
     return text as RendererOutput;
   }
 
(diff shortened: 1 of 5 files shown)


UziTech:
I feel like this may be a breaking change. Honestly I think this might be better as an extension. I don't think we have ever had any complaints that marked does not filter those html tags (I haven't actually checked). I know this follows the GMF spec but I don't think it follows most marked users expectations.

PR #4070, opened 2026-08-24
perf: fast-fail link tokenizing when no ')' remains ahead

koding88:
**Marked version:** master (c430a64, v18.0.10)

**Markdown flavor:** n/a (parser performance)

## Description

The inline link rule backtracks quadratically over unterminated link sequences:

```js
marked.parse('[a](b'.repeat(20000)) // 97KB input
```

| Input size | Before | After |
|---|---|---|
| 48KB (`[a](b`×10k) | ~550ms | ~7ms |
| 97KB (`[a](b`×20k) | ~2500ms | **~24ms** |
| 195KB | ~9.9s | ~28ms |
| 390KB | (minutes) | ~63ms |

Doubling the input quadruples the time — O(n²). At every `[`, the rule's greedy unquoted-href run `[^ \t\n\x00-\x1f]+` swallows the entire remainder and then re-partitions it char-by-char while hunting for a closing `)" that never comes; the lexer advances one character and repeats. This is the same class as #4013 (empty-href + whitespace variant), but with non-empty content between brackets, so that fix does not cover it — and none of the existing redos guards match this shape.

## The change

`Lexer.inlineTokens` anchors one `lastIndexOf(')')` scan per inline run, and `Tokenizer.link` consults it before running the regex: since every src it receives during the run is a suffix of that anchor, "is there a `) ahead?" is answerable in O(1). When no paren remains, the regex cannot match anyway, so skipping is semantics-preserving — the skip only removes the dead backtracking.

Details:

- Anchors are saved/restored around nested inline runs (e.g. link-label lexing), so outer runs keep a valid anchor after recursion returns.
- Hints live in a module-level WeakMap keyed by the tokenizer instance, so nothing appears on the public tokenizer API surface.
- A standalone `tokenizer.link()` call outside a lexer run lazily anchors itself.

An earlier design that re-anchored on run-id mismatch instead of saving/restoring was 100× slower on reflink-heavy input (`quadratic_inline_masking[2]` blew its budget at ~10s); save/restore fixed that while keeping the unterminated-link win.

## Verification

- Differential check: 1,506 generated documents (links with titles, balanced parens, angle destinations, images, autolinks, code spans containing brackets/parens, escaped chars, reflinks + defs, nested brackets, unterminated sequences) × `gfm` on/off → byte-identical output before/after.
- New guard `test/specs/redos/quadratic_unterminated_links.cjs`: exceeds the redos budget on master (~13.7s), passes with this change.
- `npm test` green: 1791 spec tests, 191 unit tests, UMD/CJS, types, ESLint.

AI disclosure: OpenAI Codex assisted with implementation and validation; I reviewed the parser behavior, ran the differential suite, and own the contribution.

## Contributor

## Committer

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -349,9 +349,11 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
     // One paren scan per inline run, shared with Tokenizer.link: without a
     // ')' ahead the link regex cannot match, and skipping it avoids its
     // quadratic backtracking over unterminated-link inputs like
-    // `'[a](b'.repeat(n)`.
+    // `'[a](b'.repeat(n)`. A nested run receives a substring of the outer
+    // run, so if the outer had no ')' neither does this substring — skip
+    // the scan in that case.
     const prevLinkParenPossible = this.state.linkParenPossible;
-    this.state.linkParenPossible = src.includes(')');
+    this.state.linkParenPossible = prevLinkParenPossible && src.includes(')');
     try {
       return this.inlineTokensInner(src, tokens);
     } finally {
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -1,4 +1,4 @@
-import { _Tokenizer } from './Tokenizer.ts';
+import { _Tokenizer, linkParenHints } from './Tokenizer.ts';
 import { _defaults } from './defaults.ts';
 import { other, block, inline } from './rules.ts';
 import type { Token, TokensList, Tokens } from './Tokens.ts';
@@ -338,7 +338,29 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
    * Lexing/Compiling
    */
   inlineTokens(src: string, tokens: Token[] = []): Token[] {
+    const prevHint = linkParenHints.get(this.tokenizer);
+    linkParenHints.set(this.tokenizer, {
+      anchorLen: src.length,
+      lastParenFromStart: src.lastIndexOf(')'),
+    });
+    try {
+      return this.inlineTokensInner(src, tokens);
+    } finally {
+      // Nested runs anchored their own label substring; restoring keeps the
+      // outer run's anchor valid for the remainder of its loop.
+      if (prevHint) {
+        linkParenHints.set(this.tokenizer, prevHint);
+      } else {
+        linkParenHints.delete(this.tokenizer);
+      }
+    }
+  }
+
+  private inlineTokensInner(src: string, tokens: Token[] = []): Token[] {
     this.tokenizer.lexer = this;
+    // Every src seen in the loop below is a suffix of the string anchored by
+    // inlineTokens(), which lets Tokenizer.link answer "is there a ')' ahead?"
+    // in O(1) instead of backtracking quadratically over such input.
     // String with links masked to avoid interference with em and strong
     let maskedSrc = src;
 
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -1,4 +1,4 @@
-import { _Tokenizer, linkParenHints } from './Tokenizer.ts';
+import { _Tokenizer } from './Tokenizer.ts';
 import { _defaults } from './defaults.ts';
 import { other, block, inline } from './rules.ts';
 import type { Token, TokensList, Tokens } from './Tokens.ts';
@@ -15,6 +15,12 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
     inRawBlock: boolean;
     /** a link was produced in the inline run currently being scanned */
     linkEmitted: boolean;
+    /**
+     * false while scanning an inline run whose source contains no ')',
+     * letting Tokenizer.link bail out before its quadratic backtracking.
+     * Anchored once per run by inlineTokens.
+     */
+    linkParenPossible: boolean;
     top: boolean;
   };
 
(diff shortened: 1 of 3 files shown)


src/Tokenizer.ts
```diff
@@ -683,6 +698,24 @@ export class _Tokenizer<ParserOutput = string, RendererOutput = string> {
...
+      }
+    } else {
+      // No anchor yet (direct call outside a lexer run) or the string grew
+      // beyond it: anchor lazily. Every later call in this run is a suffix
+      // of this string, so anchoring here stays sound.
+      linkParenHints.set(this, {
+        anchorLen: src.length,
+        lastParenFromStart: src.lastIndexOf(')'),
+      });
+    }
```
UziTech, line 694: Seems like this will be a source of many bugs in marked and other extensions. Even though it would take a few extra nano seconds I think it would be worth it to simplify the code and just do:

```js
if (!src.includes(')')) {
  return;
}
```

koding88:
Thank you for the review @UziTech — implemented along the lines you suggested, with one measurement worth sharing.

Your exact snippet (`if (!src.includes(')')) return;` at the top of `link()`) does fix the unterminated-link case (97KB: ~2500ms → ~62ms), but it re-scans the remainder on **every** link attempt. On reflink-heavy paragraphs that adds up: `quadratic_inline_masking[2]` (`'[a]: x\n\n' + '[a] '.repeat(100000)`, no parens at all) went from ~140ms on master to **~1030ms — right over this suite's 1s redos budget** — because link() is attempted once per token and each attempt re-scans.

So I kept your idea but hoisted the scan to one per inline run:

- `Lexer.inlineTokens` sets `state.linkParenPossible = src.includes(')')` (single scan), restoring the previous value after nested runs — the same save/restore pattern `linkEmitted` already uses since #4051.
- `Tokenizer.link` bails only when the flag is `false`, which is exactly when its regex cannot match, so behavior is unchanged otherwise.

Current numbers on Node 24 (same machine):

| case | master | this PR |
|---|---|---|
| `'[a](b'.repeat(20000)` (97KB) | ~2500ms | ~22ms |
| quadratic_inline_masking[2] | ~140ms | ~144ms |

Differential: 1,506 generated docs × gfm on/off → byte-identical to master. Full `npm test` green including all redos guards.

Happy to flatten it further to the plain two-liner if you'd rather trade the masking[2] budget for fewer moving parts — say the word and I'll push that instead.

src/Lexer.ts
```diff
@@ -339,6 +346,22 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
    */
   inlineTokens(src: string, tokens: Token[] = []): Token[] {
     this.tokenizer.lexer = this;
+    // One paren scan per inline run, shared with Tokenizer.link: without a
+    // ')' ahead the link regex cannot match, and skipping it avoids its
+    // quadratic backtracking over unterminated-link inputs like
+    // `'[a](b'.repeat(n)`.
+    const prevLinkParenPossible = this.state.linkParenPossible;
+    this.state.linkParenPossible = src.includes(')');
```
UziTech: This also only needs to be checked if `prevLinkParenPossible` is `true`, correct? Since `inlineTokens` will only be called with a substring of the previous call. If the previous call didn't include a `')'` than neither will this one.

koding88:
Good catch on both points — pushed `908ad32`.

**Line 354** (skip scan when outer had no `)`): you're right, a nested run receives a substring of the outer run, so if the outer had no `)`, neither does the substring. Changed to:

```js
this.state.linkParenPossible = prevLinkParenPossible && src.includes(')');
```

so the `includes(')')` scan is skipped entirely for nested runs whose parent already established no `)` is possible. The short-circuit also avoids touching the string at all for the common nested-reflink case.

Verified: 1791 spec + 191 unit tests pass, and the quadratic input (`'[a](b'.repeat(50000)`) still finishes in ~43ms (would be minutes without the guard).

UziTech (approved):
Thanks for helping make marked faster! 💯

src/Lexer.ts
```diff
@@ -339,6 +346,24 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
...
+    try {
+      return this.inlineTokensInner(src, tokens);
+    } finally {
+      // A nested run (e.g. a link label) anchored on its own substring;
+      // restore the outer run's flag for the rest of its loop.
+      this.state.linkParenPossible = prevLinkParenPossible;
+    }
+  }
+
+  private inlineTokensInner(src: string, tokens: Token[]): Token[] {
```
styfle, line 366: I we want this to be truly private, we should use the JS standard since the TS private keyword doesn't actually do anything at runtime.

```suggestion
  #inlineTokensInner(src: string, tokens: Token[]): Token[] {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements#private_methods

spokodev:
A note on the guard's reach — I'll keep the details out of this thread in case you'd rather have them somewhere else.

The fast-fail closes the unterminated-link shape it targets. It does not close the same rule when a `)` appears later in the same inline run: measured on this branch against 18.0.13, the cost there is unchanged. The payload this PR adds as a guard is not affected, so the specs stay green either way — which is why I'm mentioning it rather than assuming you'd see it.

I went through private vulnerability reporting first; that report is closed, so I genuinely don't know which channel you want. I can post the measurements and a `test/specs/redos/` case here, or send them wherever suits you.

UziTech:
@spokodev it sounds like you are saying that marked could be faster in specific situations that only arise when allowing untrusted markdown to be parsed. If you have a fix then you can create a new PR with the fix 😁👍. If not then it is already known and you can read https://marked.js.org/using_advanced#workers on how to prevent it from causing an issue.

PR #4071, opened 2026-08-28
fix: O(1) fast-reject for unclosed link syntax to prevent quadratic-time DoS.

rishaankgupta:
Fixes the algorithmic complexity issue described in GHSA-66g2-49g4-pwrq.

The inline tokenizer's link-matching attempt was repeated once per
unmatched '[' across the loop, each call costing time proportional to
the remaining string length when no closing ')' exists ahead --
aggregate O(n^2) for long runs of unclosed "[x](" sequences.

This adds a one-time O(n) precomputation of the last ')' position
before the loop, then an O(1) check per iteration to skip the link
attempt entirely when it's structurally impossible to match.

Verified against the full spec suite (1,783 tests, including all
existing quadratic/redos regression tests) -- all pass. Added a new
regression test (quadratic_link_unclosed_repeated.cjs) covering this
specific case.

Benchmark: 250KB adversarial payload, 26.66s -> 136.86ms.

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -385,6 +385,12 @@ export class _Lexer<ParserOutput = string, RendererOutput = string> {
     let keepPrevChar = false;
     let prevChar = '';
     let srcLength = Infinity;
+    // One-time precomputation (O(n)) so the link-tokenizer fast-reject
+    // below can be an O(1) check per loop iteration instead of re-scanning
+    // the shrinking remainder of src on every call (which would itself be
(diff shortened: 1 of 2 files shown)


UziTech:
Is this the same fix as #4070?

rishaankgupta:
Yeah, same root cause and approach same fix idea. theirs looks more refined at this
point, they caught the same quadratic_inline_masking[2] edge case I
ran into and did broader differential testing on top of it. happy to
close this one in favor of #4070 if that's the direction you'd
rather go.

PR #4073, opened 2026-08-30
fix: do not add a newline to an empty code block

Kjubikstronk:

## Description

`Renderer.code` appends a newline unconditionally:

```js
const code = text.replace(other.endingNewline, '') + '\n';
```

For a code block with no content that turns `''` into `'\n'`, so an empty fence renders as `<pre><code>\n</code></pre>` instead of `<pre><code></code></pre>`.

The lexer is not involved. `new Lexer({gfm: false}).lex('```\n')` already gives `text: ''`.

## Expectation

```
```
```

renders `<pre><code></code></pre>`

## Result

renders `<pre><code>\n</code></pre>`

## What was attempted

This fixes four CommonMark examples, all empty fenced code blocks: 126 (unclosed fence), 130 (closed fence), 144 (empty fence with an info string) and 237 (empty fence inside a block quote). I diffed the exact-match failure set across the whole spec before and after: those four move to passing and nothing else changes, in both the CommonMark and GFM runs.

Worth flagging why this went unnoticed, since it applies to more than this bug. `htmlIsEqual` builds `@markedjs/html-differ` with only `ignoreSelfClosingSlash` and `ignoreComments`, so `ignoreWhitespaces` stays at its default of `true`. Whitespace inside `pre`/`code` is significant, so all four of these were already reported as passing and carry no `shouldFail` flag. That also means the spec suite cannot protect this fix, which is why the tests below assert exact output instead.

There is a second, separate case the same blind spot hides: example 318, a fenced block in a list item, expects `b\n\n\n` and gets `b\n\n`. That one is a lexer issue rather than a renderer one, since the code token already comes out as `b\n\n`, so I have left it out of this PR.

## Contributor

Four tests in `test/unit/marked.test.js` under `empty code block`. Three of them fail without the change and one non-empty case acts as a control. Full spec suite (1789) and unit suite (195) pass.

## Committer


UziTech (changes requested):
move tests to test/specs/new with renderExact

Kjubikstronk:
Moved to `test/specs/new/empty_code_block.md` with `renderExact` in 474ae055, and `marked.test.js` is untouched again.

One fixture covers all four cases. The unclosed fence has to come last since it runs to EOF. Reverting the renderer change fails it, so it has teeth.

UziTech (approved):
Thanks! 💯

styfle (approved):

PR #4077, opened 2026-08-31
fix: case fold reference link labels

gyanu2507:
## Description

Reference link labels are matched with `toLowerCase()`, but CommonMark asks for a Unicode case fold:

> To normalize a label, strip off the opening and closing brackets, perform the *Unicode case fold*, strip leading and trailing spaces, tabs, and line endings, and collapse consecutive internal spaces, tabs, and line endings to a single space.

`toLowerCase()` doesn't get there, so labels that only agree once folded never match:

```md
[ẞ] and [Straße].

[SS]: /sharp-s
[STRASSE]: /street
```

renders as literal text today, because `'ẞ'.toLowerCase()` is `'ß'` — never `'ss'`. Round-tripping through upper case reaches the fold, which is what commonmark.js does in `normalizeReference()`:

```js
.toLowerCase()
.toUpperCase();
```

I lowercase once more afterwards so `def.tag` and the `tokens.links` keys stay in the lower case form they've always had — an uppercase key would be a visible change for anyone reading `lexer.tokens.links` or a `def` token, and it isn't needed to get the fold.

The two link-map lookups in `Lexer` (`linkInText` and the reflink mask) used the label verbatim rather than normalized, so they go through the same helper. Without that they'd start missing definitions they used to find as soon as the stored keys are folded. As a side effect the mask now also handles labels whose case differs from the definition.

This enables CommonMark example 540, taking Links from 81/90 to 82/90.

## Contributor

- [x] Docs have been added/updated (for bug fixes/features)

Verification notes:

- Removed the `shouldFail` flag from example 540 in both `test/specs/commonmark` and `test/specs/gfm`.
- Added `test/specs/new/reflink_case_folding.{md,html}` covering `ẞ`/`SS`, `Straße`/`STRASSE`, `ǅ`/`Ǆ` and `ﬁ`/`FI`. I generated the expected HTML from commonmark.js 0.32 and it is byte-identical to marked's output with this change; the fixture fails on `master`.
- `npm test` is green: specs 1789 -> 1791 passing, unit 191 passing, 0 failures, lint clean. No existing test needed adjusting.

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -1,6 +1,7 @@
 import { _Tokenizer } from './Tokenizer.ts';
 import { _defaults } from './defaults.ts';
 import { other, block, inline } from './rules.ts';
+import { normalizeLabel } from './helpers.ts';
 import type { Token, TokensList, Tokens } from './Tokens.ts';
 import type { MarkedOptions } from './MarkedOptions.ts';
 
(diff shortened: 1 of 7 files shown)


UziTech (approved):
Awesome! 💯

styfle (approved):

PR #4083, opened 2026-09-03
fix: reject invalid characters in HTML tag names

spokodev:
Per CommonMark (Raw HTML, and HTML blocks type 7) and GFM, an HTML tag name is `[A-Za-z][A-Za-z0-9-]*`. The block and inline tag regexes used `[\w-]` (and `[\w:-]` for the inline closing tag), so `_` — and `:` on closing tags — were accepted. As a result sequences like `<bar_baz>`, `<em_>…</em_>` and `</a:b>` were emitted as raw HTML instead of being escaped as text.

This also removes an inconsistency where the inline opening-tag rule rejected `:` but the closing-tag rule accepted it.

Example:

```
foo <bar_baz> qux
```

before: `<p>foo <bar_baz> qux</p>` — after: `<p>foo &lt;bar_baz&gt; qux</p>` (matching commonmark.js).

Attribute names still accept `_`/`:` — only tag names are affected. Added a `test/specs/new` case covering inline open/close and block tags with `_`/`:`. Full spec and unit suites pass.


UziTech (approved):
Nice catch! 💯

styfle (approved):

PR #4086, opened 2026-09-05
fix: preserve internal tabs in list item content

nrps9909:
**Marked version:** 18.0.11, base `4417582f35d69926f0778ee133c3af75b7f3f966`

**Markdown flavor:** CommonMark / GitHub Flavored Markdown

## Description

List parsing expands tabs throughout item text before that text reaches code-span or fenced-code tokenization. For example, `- \`foo\tbar\`` renders two spaces between `foo` and `bar`, and a fence continuation containing `foo\tbar` renders four spaces. Both should retain the literal tab under [CommonMark's tab rules](https://spec.commonmark.org/0.31.2/#tabs).

Expand only the leading spaces/tabs when preparing list lines. Content after that prefix reaches the child tokenizer unchanged. The first-line calculation retains its existing tab stops, continuation indentation retains its existing behavior, and pedantic mode retains its previous path.

This addresses internal content tabs in the list tokenizer discussed in [#4050](https://github.com/markedjs/marked/issues/4050#issuecomment-5537906424). It does not close that broad issue: tabs within the leading indentation, including a content tab directly after the list indent, still need original-column tracking across nested containers. It is separate from #4080's indented-code de-indentation change.

Validation:

- All 16 Markdown/HTML exact-output fixture pairs fail on the base and pass after the change. The eight scenarios each run with GFM both enabled and disabled and use `renderExact: true`; cases cover first-line/continuation code spans, backtick/tilde fences, ordered/nested lists, repeated internal tabs, and a structural indentation tab before the content.
- `npm test` passes: 1,833 spec-harness tests, 191 unit tests, docs/build, UMD/CJS checks, type/package checks and lint.
- An independent 2,058-input comparison against `commonmark` 0.31.2 checks list-marker widths, whitespace prefixes and nested fences. This patch fixes 452 mismatches and introduces zero new mismatches; 1,410 existing mismatches in that deliberately broad corpus remain. The comparison preserves code whitespace while normalizing insignificant HTML formatting.

Implemented and validated with Codex.

## Contributor

- [x] This fixes existing parsing behavior; no new option is introduced.

## Committer


test/unit/marked.test.js
UziTech, line 1: Move the tests to test/specs/new. You can use renderExact if you need to test white space.

UziTech (changes requested):

test/unit/marked.test.js
nrps9909: Moved the eight scenarios to `test/specs/new/list_content_tabs.json`, with separate GFM-on/off cases and `renderExact: true` on all 16 fixtures. They now compare the complete HTML string, including literal tabs and newlines, through the spec harness. The unit-test block is removed.

All 16 fixtures fail against the original base and pass on the updated branch. Full `npm test` passes with 1,818 spec-harness tests, 191 unit tests, docs/build, UMD/CJS, type/package checks and lint. The source fix is unchanged.

test/specs/new/list_content_tabs.json
UziTech, line 1: Can you move this to .md and .html files? They are easier to view. You can add options as front matter

UziTech (changes requested):

test/specs/new/list_content_tabs.json
nrps9909: Moved all 16 cases to paired `.md` / `.html` fixtures, with `gfm` and `renderExact` in front matter. I verified the Markdown, HTML and options are identical to the JSON cases. All 16 still fail on the original base and pass on this branch; `npm test` passes (1,833 spec-harness tests, 191 unit tests, build/docs, UMD/CJS, types/package checks and lint).

UziTech (approved):
Thanks for fixing this! 💯

PR #4100, opened 2026-09-19
fix: load CLI configs with top-level await

pengboyu-dev:
The CLI cannot load an ESM config containing top-level `await` on Node.js versions that support synchronous `require(esm)`. For example:

```js
// config.mjs
export default await Promise.resolve({ breaks: true });
```

`marked --config config.mjs --string text` exits with `ERR_REQUIRE_ASYNC_MODULE` on Node 24.19.0. The config loader only falls back to `import()` for `ERR_REQUIRE_ESM`, so it never reaches the existing asynchronous import path.

Treat `ERR_REQUIRE_ASYNC_MODULE` as another reason to use that path. Other loading errors continue to propagate, and the existing JSON/CommonJS loading path is unchanged.

Validation:
- Added a CLI regression test using a config with top-level `await`; it fails before the fix and passes after it.
- Full `npm test` passes on Node 24.19.0, including specs, unit tests, UMD/CJS, types, and lint.

## Contributor
- [ ] No tests required for this PR.
- [ ] If submitting a new feature, it has been documented in the appropriate places.


UziTech (approved):
Nice catch! 💯

bin/main.js
```diff
@@ -180,7 +180,7 @@ export async function main(nodeProcess) {
         // try require for json
         markedConfig = require(configFile);
```
styfle, line 181: I think we can remove `require(configFile)` now since we can always run `await import(pathToFileURL(configFile).href)` with any LTS version of Node.js, and it works with ESM and CJS

PR #4101, opened 2026-09-20
fix: keep text after empty nested blockquote

### LINKED ISSUE #4098: Text after an empty nested blockquote is dropped (18.0.9+)
**Marked version:** 18.0.9 through 18.0.13 (18.0.8 is fine)

**Describe the bug**
Text on the line after an empty nested blockquote is dropped from the output, with default options.

**To Reproduce**
1. [Marked Demo](https://marked.js.org/demo/?outputType=html&text=%3E%3E%0Atext)
2. [CommonMark Demo](https://spec.commonmark.org/dingus/?text=%3E%3E%0Atext)

`marked.parse('>>\ntext')` in 18.0.8, matching the dingus:

```html
<blockquote>
<blockquote>
</blockquote>
</blockquote>
<p>text</p>
```

In 18.0.9 to 18.0.13:

```html
<blockquote>
<blockquote>
</blockquote>
</blockquote>
```

Same for `>>>\ntext` and `> quote\n>>\ntext`. With `>>\ntext\n> more`, the following `> more` blockquote disappears as well.

**Expected behavior**
The empty nested blockquote closes and `text` becomes a paragraph after it, as in CommonMark and 18.0.8.

This looks like it comes from #4030: the continuation branch in `Tokenizer.blockquote()` now appends every continuation line to `raw`, also when the re-tokenized nested blockquote did not consume them, so the lexer skips past text no token holds. Reverting that commit brings the paragraph back. I have a small fix with a spec test in `test/specs/new/` and would be glad to open a PR if that is welcome.

Yahiro025:
## Description

Since 18.0.9 (#4030), text after an empty nested blockquote was dropped. In `Tokenizer.blockquote()`, the nested-blockquote continuation path appended every leftover line to `raw` even when the re-tokenized nested quote did not consume them. The lexer then skipped that source, so following paragraphs (and later `>` quotes) vanished.

This only merges original continuation lines that `newToken` actually consumed into outer `raw`/`text` (leftover count from `newText.substring(newToken.raw.length)`, same idea as the adjacent list branch). Unconsumed lines stay out of this token so the outer lexer can still parse them. Fully consumed continuations stay unchanged.

Fixes #4098.

## Checklist

- [x] Spec fixtures for `>>\\ntext`, `>>>\\ntext`, `> quote\\n>>\\ntext`, and `>>\\ntext\\n> more`
- [x] `npm test` green (specs, unit, lint)


UziTech (approved):
Thanks! 💯

styfle (approved):

PR #4102, opened 2026-09-20
fix: keep text after empty nested blockquotes

dyk1454683243-sudo:
**Marked version:** 18.0.9–18.0.13 (18.0.8 is fine)

**Markdown flavor:** CommonMark

## Description

- **Fixes #4098**

Text after an empty nested blockquote must remain a sibling paragraph. The #4030 continuation path in `Tokenizer.blockquote()` re-lexes a nested blockquote after stripping one `>` marker, then appended the entire remaining `continuation` string to outer `raw` even when the nested token did not consume those lines. The lexer then skipped leftover source, so `marked.parse('>>\ntext')` dropped `<p>text</p>`.

This keeps the #4030 strip/re-lex behavior for restated markers, and advances outer `raw` only by the continuation lines the nested token actually consumed.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression (if no tests added, list tests covering this PR); or,
- [ ] no tests required for this PR.
- [ ] If submitting new feature, it has been documented in the appropriate places.

Regression specs under `test/specs/new/`:

- `>>\ntext`
- `>>>\ntext`
- `> quote\n>>\ntext`
- `>>\ntext\n> more` (following `> more` must not vanish)

## Committer

In most cases, this should be a different person than the contributor.

- [ ] CI is green (no forced merge required).
- [ ] Squash and Merge PR following [conventional commit guidelines](https://www.conventionalcommits.org/).

PR #4103, opened 2026-09-20
fix: avoid quadratic scan for unmatched emphasis delimiters

### LINKED ISSUE #4099: emStrong is quadratic in the number of unmatched emphasis delimiters (12 KB input takes ~16 s)
**Marked version:** 18.0.13 (also `main` at 4eb2d97)

**Markdown flavor:** n/a (default / GFM)

## Description

Parsing time grows quadratically with the number of unmatched `*` / `_` delimiters, so a small, crafted input freezes the parser. A 12 KB document takes about 16-19 seconds on my machine (Windows 11, Node 24; the machine was under some load, so treat absolute numbers loosely, but the growth rate is the point):

```js

for (const unit of ['- *', '+ _', '*x *x ']) {
  for (const n of [500, 1000, 2000, 4000]) {
    const src = unit.repeat(n);
    const t = performance.now();
    marked.parse(src);
    console.log(JSON.stringify(unit), src.length + 'B', Math.round(performance.now() - t) + 'ms');
  }
}
```

| input | 1.5-3 KB | 3-6 KB | 6-12 KB | 12 KB |
| --- | --- | --- | --- | --- |
| `'- *'.repeat(n)` (n = 500..4000) | 311 ms | 1205 ms (x3.9) | 5238 ms (x4.3) | 19517 ms (x3.7) |
| `'+ _'.repeat(n)` | 332 ms | 995 ms (x3.0) | 4001 ms (x4.0) | 15726 ms (x3.9) |
| `'*x *x '.repeat(n)` (n = 500..2000) | 893 ms | 4111 ms (x4.6) | 15993 ms (x3.9) | |

Every doubling of the input multiplies the time by about 4. Similar inputs made of `1.*`, `1)_`, `* _`, `- _` and so on behave the same way, so it is the emphasis handling rather than the list rule. The same shape without list markers (`' *_'.repeat(n)`) is also quadratic.

## Where the time goes

`node --cpu-prof` on `'- *'.repeat(2000)`: ~2.4 s of ~3.9 s is self time in `Tokenizer.emStrong`, the rest is GC and the `emStrongRDelim*` regex. Each unmatched opener runs the `while ((match = endReg.exec(maskedSrc)) !== null)` loop over *all* remaining delimiters (adding to `delimTotal` for every later left delimiter, and doing a `[...rDelim].length` spread per iteration) and only gives up at the end of the string. With `m` unmatched delimiters that is `m^2 / 2` iterations, and none of the work is shared between openers.

I looked for a small fix and did not find one that is clearly safe: whether a later opener can succeed depends on the running `delimTotal` (including the `% 3` rule and `midRun`), so failure of an earlier opener does not imply failure of a later one. The usual linear solution is the CommonMark reference algorithm (a delimiter stack with "openers_bottom" per delimiter type/length), which would be a bigger change to `emStrong` / `inlineTokens`. If there is a preferred direction (a cheap early-exit such as "no right-flanking delimiter of this kind left in the string", constant-factor cleanups in the loop, or the delimiter-stack approach) I am happy to try a PR.

## Expectation

Parse time roughly linear in input size, like other inputs of that size (the `'a_'` and plain `'*'` inputs of the same length parse in a few ms).

## Result

Seconds of CPU per 10 KB, growing 4x per doubling.

Related earlier reports were fixed the same way as regex backtracking (#3916, #3917, #4013, #4014, #4090); I did not find an existing issue for emphasis delimiters.

<sub>Found by fuzzing `marked.parse` with generated delimiter patterns and timing each in a worker thread. I drafted this report with AI assistance (Claude Code) and reproduced the numbers above myself.</sub>

dyk1454683243-sudo:
**Marked version:** 18.0.13 (`4eb2d97`)

**Markdown flavor:** n/a (parser performance)

## Description

- **Fixes #4099**

Unmatched left-flanking `*` / `_` (and GFM `~`) openers made `marked.parse` quadratic. After a left delim matched, `Tokenizer.emStrong` ran `while ((match = endReg.exec(maskedSrc)) !== null)` over every later delimiter, updating `delimTotal` / `midDelimTotal`, and only gave up at EOF. `del()` had the same scan. With `m` unmatched openers that is `m² / 2` work and none of it is shared.

This keeps the existing pairing loop (CommonMark emphasis rules 9–10 and mid-run openers such as `**a*b*c` → `**a<em>b</em>c` are unchanged). It adds a small cached scan of `maskedSrc` for the last `*` / `_` / `~` run that can still close (not preceded by whitespace or the start of the string). The right-delim regexes only treat a run as a closer in groups 1, 2, 5 and 6, all of which need a non-whitespace predecessor, so if none remains after the opener the scan cannot succeed and is skipped. A hit only means “keep looking.”

Local times for `marked.parse(('- *').repeat(n))` and the other issue shapes:

| input | n=2000 before | n=2000 after | n=4000 after |
| --- | --- | --- | --- |
| `'- *'` | ~360ms (and ~4× per doubling) | ~10ms | ~12ms |
| `'+ _'` | ~319ms | ~5ms | ~13ms |
| `'*x *x '` | ~1224ms | ~7ms | ~15ms |
| `'~x ~x '` | ~777ms | ~5ms | ~9ms |

Doubling no longer costs ~4×. A full delimiter-stack rewrite is not needed for this path.

Separate from #4102 / #4098 (blockquote leftover text).

## Contributor

- `test/specs/redos/quadratic_unmatched_emphasis.cjs` — `('- *')`, `('+ _')`, `('*x *x ')`, `('~x ~x ')` × 2000 (spec harness fails if a case takes >1s)
- `test/unit/marked.test.js` — `('- *').repeat(2000)` finishes in <1s, plus mid-run / rules 9–10 / `del` checks
- Existing CommonMark / GFM emphasis and `test/specs/new/del_strikethrough` fixtures

## Committer


test/unit/marked.test.js
```diff
@@ -1122,4 +1122,37 @@ br
       assert.strictEqual(html.trim(), '<p><em>text</em></p>');
     });
   });
+
+  describe('unmatched emphasis delimiters', () => {
```
UziTech, line 1126: I don't think we need the performance tests if we have the redos tests. And the mid-run opener tests should be in test/specs/new

UziTech (changes requested):

src/Tokenizer.ts
```diff
@@ -113,6 +113,84 @@ function isLabelEndInsideToken(src: string, label: string, labelStart: number, r
   return false;
 }
 
+type CloserDelim = '*' | '_' | '~';
```
UziTech, line 116: Can we be more descriptive. closer of what?

UziTech (changes requested):

PR #4106, opened 2026-09-21
fix: avoid cubic backtracking on unicode whitespace in a link destination

spokodev:
**Marked version:** 7f7a496 (master)

## Description

Follow-up to your note on #4070 — here's the fix.

The gap before a link destination was `\s*`, which matches unicode whitespace. The destination class next to it, `[^ \t\n\x00-\x1f]+`, excludes only ASCII space, tab, newline and the C0 controls. A run of U+00A0 (or U+2003, U+202F, U+2028, U+FEFF …) matches **both**, so it can be divided between them in n ways, and the anchored rule is retried at every offset.

`marked.parse('[](' + ' '.repeat(n))`, default options, Node 26. Each point is the minimum of several runs, taken with a fixed reference workload in the same process and accepted only within 8% of its idle baseline; a second sweep reproduced every row within 10%.

| input | master | this PR |
| --- | --- | --- |
| 250 B | 6 ms | 0.02 ms |
| 500 B | 27 ms | 0.04 ms |
| 1 KB | 180 ms | 0.02 ms |
| 2 KB | 1.3 s | 0.03 ms |
| 4 KB | 10.5 s | 0.05 ms |
| 8 KB | **80 s** | 0.10 ms |

×8 per doubling on master; flat here. 8 KB of ordinary prose parses in 0.10 ms on both builds.

`test/specs/redos/quadratic_link_empty_href.cjs` is `'[](' + ' '.repeat(50000)`. The new guard is that input with the space replaced by U+00A0: master does not finish it inside 120 s, this branch renders it in 9 ms.

## The change

`\s*` → `[ \t\n]*`, both occurrences in the `link` rule. Changing only one of them takes the growth from ×7.9 to ×3.3 (leading gap) or ×3.5 (trailing gap) — better, but still super-linear. Both together make it flat.

Narrowing the other side instead — `[^\s\x00-\x1f]+` for the destination, which is what the neighbouring `autolink` rule already uses — also flattens it (×1.5), but it makes `[a](a b)` stop being a link, and the reference implementation accepts that destination. So I left the destination alone.

## What it changes

Nothing unless unicode whitespace sits in one of the two gaps:

- 14,651 generated link forms with no unicode whitespace × 5 option sets (`{}`, `gfm`, `no-gfm`, `pedantic`, `breaks`) — **0 rendered differences**
- unicode whitespace *inside* the destination, 48 forms — **0 differences**
- 765 README/CHANGELOG files from npm, including the 23 that contain unicode whitespace — **0 differences**

Where it does change, `commonmark@0.31.2` never sides with master. Over the gap grid, 464 forms render differently; the reference agrees with this branch in 208 of them and with master in 0. The rest are cases where marked and the reference already disagreed for unrelated reasons, such as `'` escaping.

```
[a]( /p "t" )        master: <a href="/p" title="t">
                          this PR and reference: not a link

[a]( /p "t" )   master: <a href="/p" title="t">
                          this PR and reference: not a link

[a](  "t" )          master:    <a href="%22t%22">      (the title is taken as the destination)
                          this PR:   <a href="" title="t">
                          reference: <a href="%C2%A0" title="t">
```

The third is an honest "neither" case: all three differ, and this branch is the closer of the two to the reference.

## Test suite

`test:specs` 1861/1861, `test:unit` 191/192, `test:types`, `test:umd`, `test:cjs`, `eslint` clean.

The one unit failure, `Lexer > inline > inlineTokens > url > url email protocols`, is present on master as well — I reproduced it at 7f7a496 before making any change. The token carries an `autolink: true` field that the expectation does not list. Today's CI runs on master are all `cancelled`, so it hasn't surfaced there yet. Not touched here.

## Contributor

- [x] Test(s) exist to ensure functionality and minimize regression — `test/specs/redos/cubic_link_unicode_space.cjs`


UziTech (changes requested):
Rebase on master to get tests to pass.

edit: Sorry didn't see you already did.

UziTech (approved):
Thanks! 💯

PR #4107, opened 2026-09-21
fix: reject unbalanced parentheses in link destinations

giaBaoJS:
**Marked version:** 18.0.13

## Description

`marked.parse('[a]: /url(x\n')` returns an empty string. The line disappears from the output. `commonmark@0.31.2`, the devDependency already in the repo, returns `<p>[a]: /url(x</p>`.

A bare link destination "includes parentheses only if (a) they are backslash-escaped or (b) they are part of a balanced pair of unescaped parentheses" ([spec 0.31.2](https://spec.commonmark.org/0.31.2/#link-destination)), and a [link reference definition](https://spec.commonmark.org/0.31.2/#link-reference-definition) is defined in terms of that destination. So `[a]: /url(x` is not a definition and the line stays a paragraph. `Tokenizer.def` accepted it anyway and produced a `def` token, which renders nothing, so the source text was dropped.

`Tokenizer.link` has run this check since #3664. `def` never did.

Second half: `findClosingBracket` returned early when the string held no `)`, so an unmatched `(` counted as balanced. That reaches inline links too, where `[link](foo(bar)` linked instead of staying literal. Both cases are in the fixtures, and every expected output there matches the reference parser byte for byte.

Conformance does not move: 649/652 with gfm off, 645/652 with gfm on, before and after, and the completion table is unchanged. The 652 hold no example of this shape, which is why it was never measured.

`test:unit` already fails on `url email protocols` at master 7f7a4967, before this branch.

## Contributor

## Committer


UziTech (changes requested):
Rebase on master to get tests to pass

